PR #14, opened 2019-05-08
Made edges have weak references to Nodes.

Thenerdstation:
When running in eager mode, tensorflow will deallocate tensors only when python garbage collects their objects. Since we have cyclic dependencies (Node -> Edge -> Node), our objects don't get garbage collected immediately after contraction.

In this commit, we use a weak reference to break the cycle between nodes and edges. When a Node is removed from the node_set, it's tensor will also be deallocated (assuming there are no other references by the user.)

### DIFF
--- a/tensornetwork/tensornetwork.py
+++ b/tensornetwork/tensornetwork.py
@@ -21,6 +21,7 @@ import collections
 from typing import List, Optional, Union, Text, Tuple
 import numpy as np
 import tensorflow as tf
+import weakref
 from tensornetwork import decompositions
 
 
@@ -284,9 +285,9 @@ class Edge:
       raise ValueError(
           "node2 and axis2 must either be both None or both not be None")
     self.name = name
-    self.node1 = node1
+    self.set_node1(node1)
     self.axis1 = axis1
-    self.node2 = node2
+    self.set_node2(node2)
     self.axis2 = axis2
 
   def get_nodes(self) -> List[Optional[Node]]:
@@ -308,16 +309,37 @@ class Edge:
     """
     if self.axis1 == old_axis and self.node1 is old_node:
       self.axis1 = new_axis
-      self.node1 = new_node
+      self.set_node1(new_node)
     elif self.axis2 == old_axis and self.node2 is old_node:
       self.axis2 = new_axis
-      self.node2 = new_node
+      self.set_node2(new_node)
     else:
       raise ValueError("Edge '{}' did not contain node '{}' on axis {}. "
                        "node1: '{}', axis1: {}, node2: '{}', axis2: {}".format(
                            self, old_node, old_axis, self.node1, self.axis1,
                            self.node2, self.axis2))
 
+  def set_node1(self, node):
+    try:
+      self._node1 = weakref.ref(node)
+    except TypeError:
+      self._node1 = node
+
+  def set_node2(self, node):
+    try:
+      self._node2 = weakref.ref(node) if node else None
+    except TypeError:
+      self._node2 = node
+
+  @property
+  def node1(self):
+    return self._node1() if self._node1 else None
+
+  @property
+  def node2(self):
+    return self._node2() if self._node2 else None
+  
+  
   def is_dangling(self):
     """Whether this edge is a dangling edge."""
     return self.node2 is None
@@ -509,13 +531,13 @@ class TensorNetwork:
         to_reduce += 1 if tmp_edge.axis1 > axes[0] else 0
         to_reduce += 1 if tmp_edge.axis1 > axes[1] else 0
         tmp_edge.axis1 -= to_reduce
-        tmp_edge.node1 = new_node
+        tmp_edge.set_node1(new_node)
       if tmp_edge.node2 is edge.node1:
         to_reduce = 0
         to_reduce += 1 if tmp_edge.axis2 > axes[0] else 0
         to_reduce += 1 if tmp_edge.axis2 > axes[1] else 0
         tmp_edge.axis2 -= to_reduce
-        tmp_edge.node2 = new_node
+        tmp_edge.set_node2(new_node)
     # Update edges for the new node.
     for i, e in enumerate(node_edges):
       new_node.add_edge(e, i)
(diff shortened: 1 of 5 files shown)


tensornetwork/tensornetwork.py
```diff
@@ -308,19 +309,43 @@ def update_axis(self, old_axis: int, old_node: Node, new_axis: int,
...
-      self.node2 = new_node
+      self.set_node2(new_node)
     else:
       raise ValueError("Edge '{}' did not contain node '{}' on axis {}. "
                        "node1: '{}', axis1: {}, node2: '{}', axis2: {}".format(
                            self, old_node, old_axis, self.node1, self.axis1,
                            self.node2, self.axis2))
 
-  def is_dangling(self):
+  def set_node1(self, node: Node) -> None:
```
viathor: You should be able to avoid breaking change to the API by using the `@*.setter` [decorator](https://docs.python.org/3/library/functions.html#property).

tensornetwork/tensornetwork.py
```diff
@@ -308,19 +309,43 @@ def update_axis(self, old_axis: int, old_node: Node, new_axis: int,
...
+    try:
+      self._node1 = weakref.ref(node)
+    except TypeError:
+      self._node1 = node
```
viathor: So when this happen we end up with strong references, but the user isn't notified. AFAICT, they'll only be able to observe this indirectly via surprising and detrimental effects on performance...

When is this exception raised? Can we instead let it propagate?

tensornetwork/tensornetwork.py
Thenerdstation: Oh sweet. I didn't know that existed.

tensornetwork/tensornetwork.py
Thenerdstation: Actually, this is left over when I tried using weakref.proxy. We don't need it anymore. I'll remove it.

tensornetwork/tensornetwork.py
```diff
@@ -317,8 +318,26 @@ def update_axis(self, old_axis: int, old_node: Node, new_axis: int,
...
+  def node2(self) -> Optional[Node]:
+    return self._node2() if self._node2 else None  
+  
+  @node1.setter
+  def node1(self, node: Node) -> None:
+    self._node1 = weakref.ref(node)
+
+  @node2.setter
+  def node2(self, node: Node) -> None:
+    self._node2 = weakref.ref(node) if node else None
```
viathor: We probably don't need to check `node` here, so the conditional can be removed.

tensornetwork/tensornetwork.py
```diff
@@ -317,8 +318,26 @@ def update_axis(self, old_axis: int, old_node: Node, new_axis: int,
...
+  @property
+  def node1(self) -> Node:
+    val = self._node1()
+    if val is None:
+      raise ValueError("node1 for edge '{}' no longer exists.".format(self))
+    return val
+
+  @property
+  def node2(self) -> Optional[Node]:
+    return self._node2() if self._node2 else None
```
viathor: It looks like this makes it impossible to distinguish between a dangling edge and an edge with a garbage-collected neighbor. WDYT about introducing a boolean flag to indicate whether the edge is dangling so that we don't have to rely on special value `None` for this purpose?

tensornetwork/tensornetwork_test.py
```diff
@@ -791,6 +791,33 @@ def build_tensornetwork(tensors):
...
+    b = net.add_node(np.eye(2))
+    c = net.add_node(np.eye(2))
+    e1 = net.connect(a[0], b[0])
+    e2 = net.connect(b[1], c[0])
+    net.contract(e1)
+    net.contract(e2)
+    # This won't raise an exception since we still have a referance to 'a'.
+    e1.node1
+    with self.assertRaises(ValueError):
+      e2.node1
```
viathor: Could use a comment about why this raises an exception.

tensornetwork/tensornetwork.py
Thenerdstation: That seems fine. I'll add it.

tensornetwork/tensornetwork.py
Thenerdstation: There can't exist weak references to NoneType. Keeping the conditional for now.

tensornetwork/tensornetwork_test.py
Thenerdstation: Done.

tensornetwork/tensornetwork.py
viathor: Yes, but `node` is of type `Node`, so cannot be `None`. Unless the type annotation is incorrect...

Thenerdstation: It was incorrect. surprised pytype didn't catch it...

tensornetwork/tensornetwork.py
```diff
@@ -317,10 +319,32 @@ def update_axis(self, old_axis: int, old_node: Node, new_axis: int,
...
+      raise ValueError("node2 for edge '{}' no longer exists.".format(self))
+    return self._node2()
+  
+  @node1.setter
+  def node1(self, node: Node) -> None:
+    self._node1 = weakref.ref(node)
+
+  @node2.setter
+  def node2(self, node: Optional[Node]) -> None:
+    self._node2 = weakref.ref(node) if node else None
```
viathor: If `node` is `None` we should probably set `_is_dangling`, too.

tensornetwork/tensornetwork.py
```diff
@@ -288,6 +289,7 @@ def __init__(self,
     self.axis1 = axis1
     self.node2 = node2
     self.axis2 = axis2
+    self._is_dangling = node2 is None
```
viathor: I think there are some places that need to be updated now to check this flag instead of doing `node is None` or `node is not None` checks. Examples: l.808, l.835.

tensornetwork/tensornetwork.py
Thenerdstation: Good point.

tensornetwork/tensornetwork.py
Thenerdstation: Both of those are because of edge.get_edges() returning [edge.node1, edge.node2]. This is intended behavior.

PR #21, opened 2019-05-10
Added 3SAT TensorNetwork example.

Thenerdstation:


experiments/sat/sat_tensornetwork.py
```diff
@@ -0,0 +1,166 @@
...
+  """
+  net1, var_edges1 = sat_tn(clauses)
+  net2, var_edges2 = sat_tn(clauses)
+  sat_net = tensornetwork.TensorNetwork.merge_networks([net1, net2])
+  for edge1, edge2 in zip(var_edges1, var_edges2):
+    sat_net.connect(edge1, edge2)
+  return sat_net
+
+
+def contract_badly(net: tensornetwork.TensorNetwork,
```
viathor: nit: I'd change this name to better reflect what problem the user should expect. Current name suggests that there may be an issue with correctness, which is not the case. The problem relates to performance. Perhaps something like `contract_slow()`?

experiments/sat/sat_tensornetwork.py
```diff
@@ -0,0 +1,166 @@
...
+          ) -> Tuple[tensornetwork.TensorNetwork, List[tensornetwork.Edge]]:
+  """Create a 3SAT TensorNetwork of the given 3SAT clauses.
+
+    After full contraction, this network will be a tensor of size (2, 2, ..., 2)
+    with the rank being the same as the number of variables. Each element of the
+    final tensor represents whether the given assignment satisfies all of the
+    clauses. For example, if final_node.get_tensor()[0][1][1] == 1, then the
+    assiment (False, True, True) satisfies all clauses.
+
+
```
viathor: nit: remove one newline

experiments/sat/sat_tensornetwork.py
Thenerdstation: How about I delete it entirely and replace it with naive?

viathor: Sure, that works, too.

tensornetwork/__init__.py
```diff
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 from tensornetwork.tensornetwork import TensorNetwork, Node, Edge
 from tensornetwork.ncon_interface import ncon, ncon_network
-
+from tensornetwork.contractors import *
```
viathor: I think it'd be good to avoid wildcard imports since they can cause name collisions. Safer to list them all explicitly here. We only have a few :-)

experiments/sat/sat_tensornetwork.py
viathor: flagging unresolved

Thenerdstation: Done

tensornetwork/__init__.py
Thenerdstation: Done

PR #23, opened 2019-05-10
Plotting tree-like TNs.

stavros11:
A networkx wrapper for plotting tree-like tensor networks. Nodes should be given in lists grouped by the level in the tree. This has been approved internally.

There are several ideas I would like / plan to try, including: 
* Better API, ideally giving only a TensorNetwork object and getting the plot. I have a simple `graphviz` implementation but it is less customizable and does not use `pyplot` which we probably want.
* Something better looking or interactive (open to suggestions from people with more experience).
* "Smart" ways to plot very large networks, eg. plot a unit cell or a smaller part only.

Will update or raise issue but this is probably of lower priority right now.


experiments/nxwrapper.py
```diff
@@ -0,0 +1,220 @@
...
+    if edge.is_dangling():
+      n += 1
+  return n
+
+
+class PlotWrapper(object):
+  """Wraps tree-like TensorNetwork to NetworkX for plotting."""
+  # TODO: Consider case with multiple edges between nodes.
+  # Problem with the way networkx handles multiple edges between nodes.
+  # Should also change nx.Graph() to nx.MultiGraph().
```
Thenerdstation: Any reason you can't do this now?

experiments/nxwrapper.py
```diff
@@ -0,0 +1,220 @@
...
+  """
+  # TODO: Simplify this using get_all_nondangling() from #22
+  n = 0
+  for edge in node.edges:
+    if edge.is_dangling():
+      n += 1
+  return n
+
+
+class PlotWrapper(object):
```
Thenerdstation: Rename to TreePlotWrapper.

experiments/nxwrapper.py
```diff
@@ -0,0 +1,220 @@
...
+  """Wraps tree-like TensorNetwork to NetworkX for plotting."""
+  # TODO: Consider case with multiple edges between nodes.
+  # Problem with the way networkx handles multiple edges between nodes.
+  # Should also change nx.Graph() to nx.MultiGraph().
+
+  def __init__(self,
+               node_lists: List[List[tensornetwork.Node]],
```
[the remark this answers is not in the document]
stavros11: I added this and changed the API to a single function `draw_tree(root, [optional settings])`. There is a small issue with this when the root level has multiple nodes (eg. MPS) so I allow the user to give a list of roots.

I think the ideal plotter takes the TensorNetwork object and decides automatically how to plot it according to connectivity by checking the neighbors of every node (regardless of tree structure). This is a bit hard to do with my code (hard-coding all position coordinates), but I am sure there are libraries that do that automatically (perhaps networkx too), so I will try to look at this.

experiments/nxwrapper.py
stavros11: MultiGraph supports multiple edges between nodes, however they overlap in the plot and [I have not found a way to solve this](https://stackoverflow.com/questions/22785849/drawing-multiple-edges-between-two-nodes-with-networkx) (also [here](https://stackoverflow.com/questions/15053686/networkx-overlapping-edges-when-visualizing-multigraph), athough all these are a bit old).

Thenerdstation:
Looks good! 

Can you add unit tests that can be run with pytest? That helps make sure that the code doesn't die. 

You might want to use mock patching for the drawing calls.
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

Thenerdstation:
Also, add empty __init__.py files to make this a module.

stavros11:
Currently this works only for trees. I think the tests fail only because of a small matplotlib/pytest issue. But since we expect to have a more general plotter on graphviz, would it be a good idea to close this?

Thenerdstation:
Yeah I think it would be best to close it in favor of our graphviz implementation

PR #27, opened 2019-05-13
Ncon tests

amilsted:
Improve `ncon()` error handling and add some tests.

Not yet tested: `ncon_network()`. Although it is implicitly tested via `ncon()`, the interface should probably be subject to tests.

### DIFF
--- a/tensornetwork/ncon_interface.py
+++ b/tensornetwork/ncon_interface.py
@@ -117,13 +117,42 @@ def ncon_network(tensors, network, con_order=None, out_order=None):
     tn, edges = _build_network(tensors, network)
 
     if con_order is None:
-        con_order = sorted((k for k in edges.keys() if k >= 0))
+        try:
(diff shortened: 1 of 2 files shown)


tensornetwork/ncon_interface.py
```diff
@@ -147,6 +176,12 @@ def _build_network(tensors, network):
     nodes = []
     edges = {}
     for (i, (tensor, edge_lbls)) in enumerate(zip(tensors, network)):
+        # Allow single indices to be passed as scalars.
+        try:
+            edge_lbls = list(edge_lbls)
```
Thenerdstation: This is gonna cause some issues when edge_lbls are strings.

amilsted: Ah, good point.

amilsted:
Typing is in.

Thenerdstation:
Thanks for changing the spacing to 2 as well!

PR #29, opened 2019-05-13
Squeeze edge method

stavros11:
[Squeeze](https://www.tensorflow.org/api_docs/python/tf/squeeze) edge method (code is exact copy from the internal). I was going to remove this because I do not think it is very useful after all and will save us ~40 lines, but up to you to decide.

### DIFF
--- a/tensornetwork/tensornetwork.py
+++ b/tensornetwork/tensornetwork.py
@@ -967,11 +967,11 @@ class TensorNetwork:
     return self.contract_between(edge.node1, edge.node2)
 
   def squeeze(self, edge: Edge) -> Union[Node, Tuple[Node]]:
-    """Squeezes an edge with unit size.
+    """Removes an edge with unit size.
 
     If the edge is dangling it removes it from its node and updates the
-    corresponding tensor shape. If the edge is non-dangling it disconnects
-    and removes it.
+    corresponding tensor. If the edge is non-dangling it disconnects
+    and removes it from both nodes. Works for trace edge too.
 
     Args:
       edge: Edge to be squeezed (removed).
@@ -984,11 +984,19 @@ class TensorNetwork:
     """
     node = edge.node1
     axis = edge.axis1
-    if node.get_dimension(axis) > 1:
+    axis1_size = node.get_tensor().shape.as_list()[axis]
+    if axis1_size is None:
+      raise ValueError(
+          "Attempted to squeeze edge '{}' with None size.".format(edge))
+    elif axis1_size > 1:
       raise ValueError(
           "Attempted to squeeze edge '{}' with non-unit size.".format(edge))
 
     if not edge.is_dangling():
+      axis2_size = edge.node2.get_tensor().shape.as_list()[edge.axis2]
+      if axis1_size != axis2_size:
+        raise ValueError(
+            "Edge '{}' connects node axes with different sizes.".format(edge))
       dangling1, dangling2 = self.disconnect(edge)
       node1 = self.squeeze(dangling1)
       node2 = self.squeeze(dangling2)
--- a/tensornetwork/tensornetwork.py
+++ b/tensornetwork/tensornetwork.py
@@ -965,6 +965,44 @@ class TensorNetwork:
     if edge.is_dangling():
       raise ValueError("Attempted to contract dangling edge: '{}'".format(edge))
     return self.contract_between(edge.node1, edge.node2)
+  
+  def squeeze(self, edge: Edge) -> Node:
+    """Squeezes a dangling edge with unit size.
+
+    Args:
+      edge: Edge
+
+    Returns:
+      node: The node that had the squeezed edge.
+
+    Raises:
+      ValueError: If the given edge is not dangling.
+      ValueError: If the given edge has non-unit size.
+    """
+    if not edge.is_dangling():
+      raise ValueError(
+          "Attempted to squeeze non-dangling edge '{}'.".format(edge))
+
+    node = edge.node1
+    axis = edge.axis1
+    if node.get_dimension(axis) > 1:
+      raise ValueError(
+          "Attempted to squeeze edge '{}' with non-unit size.".format(edge))
+    node.tensor = tf.squeeze(node.tensor, axis=axis)
+    node.edges.pop(axis)
+    trace_edges = set()
+    for e in node.edges[axis:]:
+      if e.node1 is node:
+        if e.node2 is node:
+          if e not in trace_edges:
+            e.axis1 -= int(e.axis1 > axis)
+            e.axis2 -= int(e.axis2 > axis)
+            trace_edges.add(e)
+        else:
+          e.axis1 -= 1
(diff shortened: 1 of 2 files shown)


tensornetwork/tensornetwork.py
```diff
@@ -966,6 +966,44 @@ def contract_parallel(self, edge: Edge) -> Node:
       raise ValueError("Attempted to contract dangling edge: '{}'".format(edge))
     return self.contract_between(edge.node1, edge.node2)
 
+  def squeeze(self, edge: Edge) -> Node:
+    """Squeezes a dangling edge with unit size.
+
+    Args:
+      edge: Edge
```
Thenerdstation: Please improve the documentation

tensornetwork/tensornetwork.py
```diff
@@ -966,6 +966,44 @@ def contract_parallel(self, edge: Edge) -> Node:
       raise ValueError("Attempted to contract dangling edge: '{}'".format(edge))
     return self.contract_between(edge.node1, edge.node2)
 
+  def squeeze(self, edge: Edge) -> Node:
+    """Squeezes a dangling edge with unit size.
```
Thenerdstation: Also say that this removes the edge from the node.

Thenerdstation:
Yeah. I'm ok with not including it either (not sure where we would use it). Feel free to close the PR.

tensornetwork/tensornetwork.py
amilsted: Why restrict this to dangling edges?

amilsted:
What currently happens if we do a truncated SVD to split a node and the resulting singular-value dimension is of size 1? I assume we currently still assign an edge to it? I suppose one could use something like squeeze() to remove that (trivial) edge. 

It doesn't seem *necessary* to have this built in, but it's not an operation one can easily do manually, since it requires manipulating the network in many places.

tensornetwork/tensornetwork.py
Thenerdstation: This basically just removes unit sized dangling edges. Non-dangling unit sized edges are either outer products or no-op contractions.

Thenerdstation:
I don't see how the SVD is a special case. If there was a unit dangling edge after the SVD, it had to exist before the SVD as well. Unit sized non-dangling edges can just be contracted normally.

stavros11:
Ash probably meant the middle (non-dangling) edge that results from SVD split and did not exist before. This can be unit if there is only one singular value greater than `max_truncation_err`. We can easily extend `squeeze` to contract unit non-dangling edges but I guess the user can call just `contract` on them directly anyway.

tensornetwork/tensornetwork.py
amilsted: Yes, I know. I think it makes just as much sense to squeeze them away as it does to squeeze dangling edges.

amilsted:
Ah, but `contract` would not do the same as `squeeze` for a non-dangling edge. At least not in my head :)
It would change the shape of each tensor and remove an edge, not do an outer product.

stavros11:
So I guess the equivalent use would be to call `disconnect` in the non-dangling edge and then squeeze the two new dangling edges. I can add this to squeeze if it is useful.

tensornetwork/tensornetwork.py
```diff
@@ -966,28 +966,36 @@ def contract_parallel(self, edge: Edge) -> Node:
       raise ValueError("Attempted to contract dangling edge: '{}'".format(edge))
     return self.contract_between(edge.node1, edge.node2)
 
-  def squeeze(self, edge: Edge) -> Node:
-    """Squeezes a dangling edge with unit size.
+  def squeeze(self, edge: Edge) -> Union[Node, Tuple[Node]]:
+    """Squeezes an edge with unit size.
```
amilsted: I think we can come up with a better one-line description, preferably one that doesn't refer to "squeezing". How about "Removes an edge with unit size."?

tensornetwork/tensornetwork.py
```diff
@@ -966,28 +966,36 @@ def contract_parallel(self, edge: Edge) -> Node:
...
     node = edge.node1
     axis = edge.axis1
     if node.get_dimension(axis) > 1:
       raise ValueError(
           "Attempted to squeeze edge '{}' with non-unit size.".format(edge))
+
+    if not edge.is_dangling():
+      dangling1, dangling2 = self.disconnect(edge)
+      node1 = self.squeeze(dangling1)
+      node2 = self.squeeze(dangling2)
```
amilsted: Perhaps we should check the dimension of both axes before squeezing anything? Currently, if the axis of node2 (erroneously) has non-unit size, this may modify the network *before* throwing an exception.

Also, note that the dimension may be `None`.

stavros11:
Thanks for the comments. I changed the size check because comparisons with `get_dimension` might break in graph mode as it returns a `tf.Tensor`. Also added a check of whether the two axes have the same size, however I suspect that this is not necessary as this error should come up during the construction of the network (eg. when one tries to connect the two incompatible axes).

Regarding the `None` case, `tf.squeeze` works well and decides whether to squeeze during run time, when shapes are defined. I am not sure how this can be transferred in our case as it is not a good idea to modify anything in the network before assigning the shapes. So, I raise error for now.

tensornetwork/tensornetwork.py
amilsted: The size being `None` should not be an error - this occurs if the size of a dimension is dynamically determined during the execution of a graph and should be allowed. If the axis has dimension `None`, we should instead just let it go through and allow `tf.squeeze` to fail if the dynamic dimension was not 1.

Actually, I don't know if we should try to check the size at all. In `connect()` we do *not* check for compatible sizes, presumably because of the `None` case. Probably we should either check dimensions (where not `None`) in both places, or in neither.

Thenerdstation:
> as this error should come up during the construction of the network (eg. when one tries to connect the two incompatible axes).

You would hope, but sadly we don't :disappointed:.

Adding the `tf.assert_equal` call made creating edges incredibly slow. This issue was raised in stefan's repository.

tensornetwork/tensornetwork.py
Thenerdstation: We should be checking sizes in both cases.

I think the best we can do is to do static checks outside of `tf.assert_equal` and just let `None`s slip by. Super not ideal, but having slow `Edge` creation is even worse.

amilsted: That would be my preference too. Check using `tensor.shape`. If `tensor.shape[i] is None` we let it pass.

We *could* add a `tf.assert_equal()` op only in case the shapes are undetermined, but I don't think it adds much... Also, one is supposed to explicitly make dependent graph ops depend on the assert op succeeding, which would be horrible to do within `TensorNetwork`.

Thenerdstation:
I vote we remove this. I'm still not sold that it's ever needed.

stavros11:
From my side you are free to close.

PR #30, opened 2019-05-13
Simple stochastic contractor

stavros11:
Idea from algorithm 2 in [qTorch paper](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0208510)). I might need to rethink the cost and threshold definition as it differs from the paper (and perhaps include more test cases).


tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -0,0 +1,102 @@
...
+
+import random
+import numpy as np
+from tensornetwork import tensornetwork
+
+
+def get_random_edge(network: tensornetwork.TensorNetwork) -> tensornetwork.Edge:
+  """Returns a random connected edge.
+  If it finds a trace edge it contracts it.
+  Args:
```
Thenerdstation: Add new line before Args. (Same below).

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -0,0 +1,102 @@
...
+    network: Connected TensorNetwork we are trying to contract.
+  Returns:
+    edge: A random connected edge from the network.
+  """
+  node = random.choice(tuple(network.nodes_set))
+  edge = random.choice(node.edges)
```
Thenerdstation: I believe we have a node.get_nondanglinge_edges() method now. You can use that instead of this while loop (which will break if you have a node with only dangling edges.

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -0,0 +1,102 @@
...
+  # TODO: Verify whether this loss makes sense.
+  node1, node2 = edge.node1, edge.node2
+  # Calculate dimension of all shared edges
+  nodes = {node1, node2}
+  shared_dimension = 1
+  for edge in node1.edges:
+    if set(edge.get_nodes()) == nodes:
+      shared_dimension *= int(edge.node1.get_tensor().shape[edge.axis1])
+
+  dimension1 = int(np.prod(node1.get_tensor().shape))
```
Thenerdstation: You should try to gracefully handle tensors with None shape. Either raise an error of change the calculation, your choice.

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -0,0 +1,102 @@
...
+    max_rejections: Maximum number of rejections before you increase threshold.
+    threshold: Initial value for the threshold.
+  Returns:
+    network: TensorNetwork with a single node after fully contracting.
+  """
+  # TODO: Think about a proper threshold value for our loss.
+  network.check_connected()
+  rejections = 0
+
+  while len(network.nodes_set) > 1:
```
Thenerdstation: This might cause some issues when the network is disconnected, and this also won't contract a single node with trace edges.

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -0,0 +1,102 @@
...
+    edge = get_random_edge(network)
+    cost = edge_cost(edge)
+    if cost <= threshold:
+      network.contract_between(edge.node1, edge.node2)
```
Thenerdstation: use `contract_parallel`

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -0,0 +1,102 @@
...
+      rejections = 0
+    else:
+      rejections += 1
+      if rejections > max_rejections:
+        threshold += 1
```
Thenerdstation: Perhaps consider `threshold *= 2` instead? Your threshold cost only grows linearly, whereas tensor contractions usually have exponential cost growth.

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -0,0 +1,102 @@
...
+  dimension1 = int(np.prod(node1.get_tensor().shape))
+  dimension2 = int(np.prod(node2.get_tensor().shape))
+  prod_dimension = ((dimension1 // shared_dimension) *
+                    (dimension2 // shared_dimension))
+  cost = prod_dimension - max(dimension1, dimension2)
+  return cost
+
+
+def stochastic(network: tensornetwork.TensorNetwork,
+               max_rejections: int, threshold: int = -1
```
Thenerdstation: Our cost is determined by the memory usage of a contraction. So with that, let's make the default threshold something larger like 128MB (~2**26)?

tensornetwork/contractors/stochastic_contractor.py
stavros11: This was designed to work only with connected networks (`network.check_connected` would raise an error for disconnected). In this case also the while loop in the first function would not break because it would never get a node with dangling edges only. Trace edges would still be a problem though.
Anyway, I changed it and now it should work with disconnected networks too (hopefully). The only drawback is the `random.choice(tuple(set))` which is probably suboptimal and has to run many times.

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -71,32 +55,39 @@ def edge_cost(edge: tensornetwork.Edge) -> int:
...
+      nondangling_edges.remove(edge)
       rejections = 0
     else:
-      rejections += 1
-      if rejections > max_rejections:
-        threshold += 1
+      cost = edge_cost(edge)
+      if cost <= threshold:
+        network.contract_parallel(edge)
+        nondangling_edges.remove(edge)
```
Thenerdstation: You'll have to remove all of the edges contracted from `contract_parallel`. Sadly I don't think there is an easy way to do it. :(

tensornetwork/contractors/stochastic_contractor.py
stavros11: The exponential *2 (previous comment) argument sounds good and I think it is in agreement with the paper. However I think that 2**26 might be huge for most cases. For example there are cases where cost < 0 (when the result of the contraction is smaller (in terms of memory) than the larger of the two tensors - these are the ones you would like to contract first I think). There is nothing wrong with using large threshold but then the order of contractions will be totally random (the cost calculation won't matter). I think it is hard to come up with a good universal default value because it depends a lot on the tensor sizes.

tensornetwork/contractors/stochastic_contractor.py
stavros11: Thanks for the quick review. I think an easy way is in the loop that I do in `edge_cost` (just pushed this - it was a bit quick though so can be wrong!)

tensornetwork/contractors/stochastic_contractor.py
stavros11: In the last update I just ignore None sizes in the cost calculation. I am not sure if that's a good choice but probably better than non supporting this at all. The contractor is mostly random so it won't be perfect anyway.

tensornetwork/contractors/stochastic_contractor.py
Thenerdstation: We could set it as `None` as default to be the size of the largest tensor in the initial network. I think that is a good compromise.

stavros11: Calculating a default value according to tensor sizes is a good idea. I am not sure if the largest is the best choice (vs an average for example) because most random moves will still be accepted this way, but we can check that. The only issue is that we have to calculate all sizes prior starting so it would be more efficient to keep them in a map for future cost calculations. I started implementing this but currently it gets a bit messy because I have to update the map after every contraction. That's just a heads up that the code might change a lot to do this.

Thenerdstation: No problem! Keep doing the good work and let me know when you are ready.

stavros11:
Made a couple of changes and now threshold is initialized as the maximum size of the tensors. Since I have to loop over all nodes to find this, I thought it was a good idea to contract all trace edges before starting any other contractions, as it always suboptimal to carry the trace edges. This is what `contract_trace_edges` does. In fact it might be useful to have a contract all trace edges functionality in the core class.

Thenerdstation:
> Made a couple of changes and now threshold is initialized as the maximum size of the tensors. Since I have to loop over all nodes to find this, I thought it was a good idea to contract all trace edges before starting any other contractions, as it always suboptimal to carry the trace edges. This is what `contract_trace_edges` does. In fact it might be useful to have a contract all trace edges functionality in the core class.

Yeah that's a good idea. We might want a specialized method that does that and `flatten_all_edges`.

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -0,0 +1,144 @@
...
+  if edge.is_dangling():
+    raise ValueError(
+             "Cannot find parallel edges for dangling edge {}".format(edge))
+  nodes = {edge.node1, edge.node2}
+  parallel_dim = 1
+  parallel_edges = set()
+  for edge in edge.node1.edges:
+    if set(edge.get_nodes()) == nodes:
+      parallel_edges.add(edge)
+      edge_size = edge.node1.get_tensor().shape.as_list()[edge.axis1]
```
Thenerdstation: This reminds me, we should add a `node.shape` property to support multi backends.

No changes here for now though.

Thenerdstation:
Everything is good on my side once the pylint checks are fixed.

PR #35, opened 2019-05-14
Fix some issues with the wavefunction example and TTN experiment

amilsted:
This removes the import hacks and adds a little more documentation.
There is a lot more doc work to do on the TTN side.

### DIFF
--- a/examples/wavefunctions/wavefunctions.py
+++ b/examples/wavefunctions/wavefunctions.py
@@ -21,8 +21,6 @@ from functools import reduce
 import sys
 import tensorflow as tf
 
-# FIXME: Hack to pull in tensornetwork from the parent directory.
-sys.path.insert(0, "../../")
 import tensornetwork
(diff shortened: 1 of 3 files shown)


examples/wavefunctions/wavefunctions.py
```diff
@@ -104,6 +102,13 @@ def evolve_trotter_defun(
     euclidean=False,
     callback=None,
     batch_size=1):
```
Thenerdstation: We made this new `tf.vectorized_map` that allows you to add batching to any function and run it in parallel. It should be included in the next tensorflow release. 

No need to change anything, just a heads up.

amilsted: That's cool, although I don't think it applies to this case. Here, "batch" refers to the amount of the computation placed in a single computational graph. The computation itself is still sequential.

PR #41, opened 2019-05-14
Added craig's FFT code.

Thenerdstation:


Thenerdstation:
Our library really makes things easy to build. Doing this in numpy would've been way hard. Cool!

PR #43, opened 2019-05-15
Added Dockerfile for prod and dev deployments

zoltanegyed:
Added Dockerfile for prod and dev deployments, and
updated documentation with these new installation options.

### DIFF
--- a/dev_tools/Dockerfile
+++ b/dev_tools/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu
+
+# Install basic tools.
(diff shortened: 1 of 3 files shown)


dev_tools/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu
```
Thenerdstation: Any reason to have a second copy of this?

Thenerdstation: Oh i see release vs nightly. got it.

Issue #15, opened 2019-05-10
Rename tensornetwork.py?


amilsted:
Currently, we have a module `tensornetwork.tensornetwork`, which I find unnecessarily ambiguous. Both `tensornetwork` and `tensornetwork.tensornetwork` provide `TensorNetwork`, but only one of them provides `ncon`, etc. 

Can we rename `tensornetwork.py` to `core.py` or somesuch?

Thenerdstation:
I think what I'll likely do now is split up `tensornetwork.py` into `node.py`, `edge.py` and `network.py`.

kshithijiyer:
@Thenerdstation are you working on this or is it up for grab?

Thenerdstation:
We need to move some code from internal first before we break it up. That should be done soon so afterwards go for it.

kshithijiyer:
Okay. Will wait for your go single on this one.

Thenerdstation:
Once https://github.com/google/TensorNetwork/pull/22 is in you should be good to go.

Thenerdstation:
Ok go for it.

Be sure that `Node`, `Edge` and `TensorNetwork` classes are publicly accessible from `import TensorNetwork`

amilsted:
Hey, thanks for working on this @kshithijiyer. I noticed you still have a `tensornetwork.py`. Can we make it `network.py`?

kshithijiyer:
Splitting the files is hitting a circular import error. Hence what I feel is that splitting the file into 3 files won't be possible but a split into 2 files would be possible.

Thenerdstation:
Yeah I think you are correct. Basically, what happened is it would start to parse `node.py` and then hit `from tensornetwork import edge`, continue to parse `edge.py` until it hit `node.Node` type declaration and raise an undefined error. This only happens because we use pytype. 

Dang that sucks.

Thenerdstation:
I think we can get around it by using strings for the typing instead of using the module directly.

kshithijiyer:
I am not sure if that is a proper solution to this. This is what is happening when we split it into 3 files https://stackabuse.com/python-circular-imports/.
I feel a way out would be to have two files:
1. `support.py` -> Having classes `Node` and `Edge.`
2. `network.py` -> `TensorNetwork` class only

Thenerdstation:
So I started working on this and what is frustrating is that the tests pass by pytype doesn't. I'll be sure to raise a bug on that internally.

[...]

PR #45, opened 2019-05-15
Split tensornetwork.py into network and network_components.

Thenerdstation:
@kshithijiyer and I both tried to split `network_components` into `node` and `edge` but ran into cyclic dependency issues with pytype.

### DIFF
--- a/tensornetwork/ncon_interface.py
+++ b/tensornetwork/ncon_interface.py
@@ -19,11 +19,12 @@ from __future__ import print_function
 from typing import Any, Sequence, List, Optional, Union, Text, Tuple, Dict
 import numpy as np
 import tensorflow as tf
-from tensornetwork import tensornetwork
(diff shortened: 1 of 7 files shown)


tensornetwork/ncon_interface.py
```diff
@@ -112,20 +115,20 @@ def ncon_network(tensors: Sequence[Union[np.ndarray, tf.Tensor]],
     This is used internally by `ncon()`.
 
     Args:
-      tensors: List of `Tensor`s.
-      network: List of lists specifying the tensor network.
+      tensors: List of `tf.Tensor`s.
```
viathor: nits:
- s/List/Sequence/
- "...or numpy array.

Thenerdstation:
See https://github.com/google/TensorNetwork/issues/15

Issue #51, opened 2019-05-15
Multi backend support


Thenerdstation:
Currently, we have hard coded dependency for tensorflow, but this need not be the case. Many external researchers use JAX, Pytorch, and even raw numpy and might not feel comfortable switching over to tensorflow just to experiment with tensor networks.

Our goal is to get more people using TensorNetworks for ML and physics research and to have these computations run on accelerated hardware. Tensorflow is one option to achieve this, but maybe not the best for some researchers. 

The only things we use from tensorflow are `tensordot`, `svd`, `reshape`, `transpose` and array slicing, all of which exist in JAX/Pytorch/numpy.

I purpose we do something similar to what Keras does and allow users to hotswap these backends without affecting the public API. (We've already designed the API very well to abstract as much tensorflow away as possible).

This can be done with a `Backend` class that can be inherited to make `TensorflowBackend`, `JaxBackend` etc. We would only need to support the functions we actually use, which is quiet small. 

A user could set the backend one of two ways. Either by doing `tensornetwork.set_backend("tensorflow")` for global backends or with `tensornetwork.TensorNetwork(backend="tensorflow")` for network specific backends (which would be very useful for benchmarking!)

In side of `TensorNetwork` we would just do `self.backend.tensordot` instead of `tf.tensordot`.

Overall, not a lot of work for possibly increasing our user base substantially.

Thenerdstation:
This is a big design change so I would appreciate your inputs.

sleichen:
I think this is a fine idea --- we're not necessarily wedded to TF.

How do we handle requirements/installation? If possible, I'd prefer having a default installation that's the same as what currently exists, but with options for more sophisticated users who want to play around with different backends. Total beginners shouldn't have to make any choices.

Thenerdstation:
Yeah that's a good question. 

I agree, let's keep the default install as easy as possible for new users. It should just work out of the box. I think `pip install tensornetwork` should include tensorflow and numpy dependencies and have a separate `pip install tensornetwork-api` that doesn't have any direct dependencies.

PR #52, opened 2019-05-16
Refactored code and created tensorflow backend

Thenerdstation:
This is a big change, but I think it is for the better. (See https://github.com/google/TensorNetwork/issues/51 for details). 

The next backend we will add will be a pure numpy implementation, then JAX, and then leave torch to the community.


Thenerdstation:
This is a big design change so I would appreciate your inputs.

tensornetwork/backends/base_backend.py
```diff
@@ -0,0 +1,178 @@
...
+    Args: 
+      tensor: A tensor.
+    Returns:
+      The reshaped tensor.
+    """
+    raise NotImplementedError(
+        "Backend '{}' has not implemented reshape.".format(self.name))
+
+  def transpose(self, tensor: Tensor, perm: Sequence[int]) -> Tensor:
+    """Transpose a tensor to a given permutation.
```
sleichen: I would say "Transpose a tensor according to a given permutation"

tensornetwork/backends/base_backend.py
```diff
@@ -0,0 +1,178 @@
...
+      "Backend '{}' has not implemented concat.".format(
+          self.name))
+
+  def shape(self, tensor: Tensor) -> Tensor:
+    """Get the shape of a tensor.
+
+    Args:
+      tensor: A tensor.
+    Returns:
+      The shape of the tensor as a different tensor.
```
sleichen: How about "The shape of the input tensor returned as another tensor."

tensornetwork/backends/base_backend.py
```diff
@@ -0,0 +1,178 @@
...
+  def sqrt(self, tensor: Tensor) -> Tensor:
+    """Take the square root (element wise) of a given tensor."""
+    raise NotImplementedError(
+      "Backend '{}' has not implemented sqrt.".format(
+          self.name))
+
+  def diag(self, tensor: Tensor) -> Tensor:
+    """Create a diagonal matrix from the given vector tensor."""
+    raise NotImplementedError(
+      "Backend '{}' has not implemented sqrt.".format(
```
sleichen: sqrt --> diag

tensornetwork/backends/base_backend.py
```diff
@@ -0,0 +1,178 @@
...
+  def trace(self, tensor: Tensor) -> Tensor:
+    """Calculate the trace over the last two axes of the given tensor."""
+    raise NotImplementedError(
+      "Backend '{}' has not implemented trace.".format(
+          self.name))
+
+  def outer_product(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
+    """Calculate the outer product of the two given tensors."""
+    raise NotImplementedError(
+      "Backend '{}' has not implemented trace.".format(
```
sleichen: trace --> outer_product

Thenerdstation:
> Also, I'm not sure how switching backends will be implemented.

Two ways (I'm waiting on stavros building the numpy backend before trying it). 
1. Doing `tensornetwork.TensorNetwork(backend='numpy')`
2. Defaulting to the global backend, and changing the global backend with `tensornetwork.set_backend('numpy')`

> but it seems conceivable that someone could try to merge two networks created using two different backends. How do we deal with that?

Easiest thing to do is throw an error. We can check that the backends are the same before merging, which is pretty easy. We *could* have a converter by changing each node's tensor by doing `torch.Tensor -> np.array -> tf.Tensor`, but I expect that would get messy and I'm not sure people would use it often enought to justify the complexity. 

All 4 backends we plan on supporting right now (JAX/PyTorch/TF/numpy) allow tensors to be created by converting a `np.array` to their own `Tensor` type. All higher level network creators we build would just have to build nodes using numpy and they would get converted internally.

tensornetwork/backends/base_backend.py
Thenerdstation: Can you tell I copy and pasted? lol

Thenerdstation:
> network_components and ncon_interface need to be updated as well, right?

Yes they do,

Issue #55, opened 2019-05-16
Add pure Numpy backend


Thenerdstation:
Add a pure numpy backend once https://github.com/google/TensorNetwork/pull/52 is merged.

stavros11:
I have some progress on this, it seems that it should not be much different from the tensorflow backend. What is the current plan? Are we waiting for feedback on #52 or do we need the numpy to change the tests? Currently I can test my numpy locally by modifying `tensornetwork_test.py` (changing some `tf.ones` to `np.ones` and removing the eager related tests) but we will probably have general backend-independent tests at some point.

Thenerdstation:
Awesome!

Just wait for https://github.com/google/TensorNetwork/pull/52 to be merged. Then we can merge your numpy backend and change the tests.

Thenerdstation:
If you want to do some stuff today, you can replace the `tf` with `np` when possible and move the tensorflow specific tests into the graphmode test file.

PR #60, opened 2019-05-17
Pure Numpy backend

stavros11:
Addressing #55


tensornetwork/backend/numpy/decompositions.py
```diff
@@ -0,0 +1,70 @@
...
+"""Tensor Decomposition Numpy Implementation."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import numpy as np
+from typing import Optional, Tuple
+
+
+def svd_decomposition(tensor: np.ndarray,
```
Thenerdstation: Can you remove the numpy import, add a argument `np` and pass `self.np` into this function? 

Normally I wouldn't make you do this, but by doing it that way we can get JAX support for free.

stavros11: Oops, yes, I forgot to do this in decompositions. Just make sure typing is okay.

Thenerdstation:
Add `__init__.py`, and go ahead and do the `Tensor = Any` hack we've been doing to replace the `"np.array"` type issues.

stavros11:
Okay, changed everything to `Tensor = Any`. Can you fix the typing of the `np` argument in `svd_decomposition` because I am not sure what that should be. Also, will we continue using `tf.test` in the tests for all backends?

Thenerdstation:
> Okay, changed everything to Tensor = Any. Can you fix the typing of the np argument in svd_decomposition because I am not sure what that should be.

Yeah I can do that.

> Also, will we continue using tf.test in the tests for all backends?

Probably not. The only `tf.test.TestCase` thing we use is `self.AssertAllClose` and I tested replacing it with `np.testsing.assert_allclose` and it worked for TF and PyTorch. We'll just replace it with that. (pytest doesn't even require using classes).

It might cause some issues internally but we'll deal with that when we get to it.

stavros11:
I think lint doesn't like my `__init__.py` but I think the rest is okay. I can change that to `np.testing` then (I have used this internally too).

Thenerdstation:
Yeah, go ahead and do that in a different PR.

Issue #63, opened 2019-05-19
Inner and outer shape


viathor:
The number of legs and the dimensions of vector spaces corresponding to each leg are determined by the shape property of the tensor encapsulated in a node. Certain types of tensors, e.g. diagonal ones resulting from SVD, can be manipulated more efficiently if the distinction is made between their inner and outer shapes. The former determines the shape of the array that holds the data (e.g. 1-D array for a diagonal matrix) and the latter determines the legs of the tensor (e.g. two legs for a matrix, whether diagonal or not).

Note that a tensor D whose inner and outer shapes differ can be thought of as a tensor network consisting of D and a few copy tensors that "adjust" its outer shape.

Thenerdstation:
How do you want to deal with this exactly? What would the API changes look like?

amilsted:
It might be worth considering whether we want ultimately to also support symmetric tensors, with guaranteed invariance under the action of some symmetry group on their indices. Such tensors have a "block-diagonal" structure: They have many zero elements by construction, so they can be handled more efficiently (in many cases) by storing only the non-zero blocks. Each block is an "inner" tensor with the same rank as the "outer" tensor, but with different axis dimensions.

Such tensors are not only stored differently; they also want to be treated differently when contracted with other tensors. If the other tensor is also symmetric, the contraction can be done block-wise (sparsely) for more efficiency.

Are you also imagining such cases @viathor?

mganahl:
I think block-sparsity would be interesting to have, but I am not sure if it should be on the level of the API.  It should probably be implemented directly into the tensor.

amilsted:
@mganahl Wouldn't that mean implementing sparse tensors in the backends? I'm not sure it's possible to extend TensorFlow's tensors that way - how would you tell the ops to handle sparse tensors differently? I think you would have to modify ops like tensordot() to make it work.

Thenerdstation:
Actually, I think this is totally doable without a large code change.

Basically, the idea is to make `node.tensor` an `@property` method. Then, for special case tensors like `CopyTensor`, you could just generate the tensor on the fly instead of storing the dense representation. We could add a default way to do this using just preexisting backend components and/or we could build specialized ways of doing it just for JAX/TF. There would be no need to modify `net.contract`, and we could add special contraction methods like `net.contract_copy_tensor` that will utilize the sparsity of the tensor for more efficient contraction. Best of both worlds!

[...]

Issue #62, opened 2019-05-19
More flexible add_node()


viathor:
Currently, `add_node()` requires one to specify a tensor. Certain types of nodes are more efficient to manipulate based on their abstract properties than based on their component representation (e.g. copy tensors, Levi-Civita tensors).

We should generalize the API to allow one to add nodes into the network without explicitly computing those tensors. For example, we could re-define `add_node()` to take a Node instead of a Tensor and leave Node construction to the caller. This would allow us to do things such as:

```python
tensor = np.array(...)
net.add_node(tn.Node(tensor))

net.add_node(tn.CopyNode(rank=2, dimension=3))

net.add_node(tn.LeviCivitaNode(rank=4))
```

Of course in practice most nodes are created from multidimensional arrays, so for convenience we could also offer a shortcut e.g. `add_tensor()` such that the following two code snippets are equivalent:

shortcut:

```python
net.add_tensor(tensor)
```

long form:

```python
net.add_node(tn.Node(tensor))
```

Thenerdstation:
I weary of changing the `add_node` API, but I agree adding higher level APIs would be good for common tensors.

Can we do something like `net.add_copy_tensor(rank=2, dimension=3)`?

viathor:
We're already doing it, see #64. Existing `add_node()` introduces strong coupling between TensorNetwork and Node which makes it hard to add new node types.

Also, I'm concerned with cluttering the API with `add_*_node()` methods in TensorNetwork for various node types that we may introduce and with code duplication in those methods.

sleichen:
I think `net.add_node(tn.Node(tensor))` is too verbose. Can we make `net.add_node()' parse its input so that `net.add_node(tensor)' still works and is equivalent to `net.add_node(tn.Node(tensor))`?

Thenerdstation:
Yeah I want gonna suggest that. It would be pretty easy to do as long as all nodes inherit `Node`

Thenerdstation:
If you want to make the tensors of the node lazily evaluated, go ahead and make `node.tensor` a `@property`. I was actually planning on doing this already and deprecate `get_tensor()`

PR #64, opened 2019-05-19
Initial version of CopyNode

viathor:
Copy tensor is a tensor of shape `(d, d,..., d)` which is equal to 1 when its indices are equal and 0 otherwise. These tensors admit very efficient contraction algorithms by allowing one to contract all their edges simultaneously. This requires ranging over only d values instead of d^k where k is the rank of the copy tensor.

This PR introduces a new type of node named `CopyNode` to represent copy tensors and implements the efficient contraction using einsum.

Some follow-ups:
 * generalization of `add_node()` and removal of `add_copy_tensor()`, see #62,
 * integration of `contract_all_edges()` into `TensorNetwork`,
 * code to merge multiple copy tensors connected by edges,
 * performance benchmarks to demonstrate the gains from the contraction algorithm in this PR,
 * new contractor using Bucket Elimination (contraction inside a bucket is performed using the algorithm in this PR),
 * inner and outer shapes to enable efficient representation of diagonal nodes, see #63.

### DIFF
--- a/tensornetwork/network.py
+++ b/tensornetwork/network.py
@@ -17,7 +17,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 import collections
-from typing import List, Optional, Union, Text, Tuple, Any
+from typing import Any, List, Optional, Union, Text, Tuple, Type
 import numpy as np
 import weakref
 from tensornetwork.backends.tensorflow import decompositions
@@ -114,7 +114,8 @@ class TensorNetwork:
       rank: int,
       dimension: int,
       name: Optional[Text] = None,
-      axis_names: Optional[List[Text]] = None) -> network_components.CopyNode:
+      axis_names: Optional[List[Text]] = None,
+      dtype: Type[np.number] = np.float64) -> network_components.CopyNode:
     """Create a new copy node in the network.
 
     Copy node represents the copy tensor, i.e. tensor C such that
@@ -138,7 +139,7 @@ class TensorNetwork:
     if axis_names is None:
       axis_names = [self._new_edge_name(None) for _ in range(rank)]
     new_node = network_components.CopyNode(
-            rank, dimension, name, axis_names, self.backend)
+            rank, dimension, name, axis_names, self.backend, dtype)
     self.nodes_set.add(new_node)
     return new_node
 
--- a/tensornetwork/network.py
+++ b/tensornetwork/network.py
@@ -114,7 +114,7 @@ class TensorNetwork:
       rank: int,
       dimension: int,
       name: Optional[Text] = None,
-      axis_names: Optional[List[Text]] = None) -> network_components.Node:
+      axis_names: Optional[List[Text]] = None) -> network_components.CopyNode:
     """Create a new copy node in the network.
 
     Copy node represents the copy tensor, i.e. tensor C such that
--- a/tensornetwork/network.py
+++ b/tensornetwork/network.py
@@ -109,6 +109,39 @@ class TensorNetwork:
     self.nodes_set.add(new_node)
     return new_node
 
+  def add_copy_node(
+      self,
+      rank: int,
+      dimension: int,
+      name: Optional[Text] = None,
+      axis_names: Optional[List[Text]] = None) -> network_components.Node:
+    """Create a new copy node in the network.
+
+    Copy node represents the copy tensor, i.e. tensor C such that
+
+        Cij...k = 1    if i = j = ... = k
+        Cij...k = 0    otherwise
+
+    Args:
+      rank: Number of edges of the copy tensor.
+      dimension: Dimension of each edge.
+      name: The name of the new node. If None, a name will be generated
+        automatically.
+      axis_names: Optional list of strings to name each of the axes.
+
+    Returns:
+      new_node: The new node object.
+    Raises:
+      ValueError: If `name` already exists in the network.
+    """
+    name = self._new_node_name(name)
+    if axis_names is None:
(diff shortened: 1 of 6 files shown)


tensornetwork/network_components.py
```diff
@@ -232,6 +237,82 @@ def __str__(self) -> Text:
...
+    einsum_input_terms = []
+    for partner_node, shared_axis in partners:
+      einsum_input_term, next_index = self._make_einsum_input_term(
+              partner_node, shared_axis, next_index)
+      einsum_input_terms.append(einsum_input_term)
+    einsum_output_term = self._make_einsum_output_term(next_index)
+    einsum_expression = ",".join(einsum_input_terms) + "->" + einsum_output_term
+    return einsum_expression
+
+  def contract_all_edges(self) -> Tensor:
```
Thenerdstation: This really shouldn't be part of `CopyNode`. We've kept the interface to be `net.contract_*` and we did that to allow all of the nodes and edges to be updated automatically. Since this only calculates the contracted tensor, maybe name it something like `calculated_full_contraction` and add `contract_copy_tensor` to TensorNetwork.

tensornetwork/tensornetwork_test.py
```diff
@@ -884,6 +884,29 @@ def test_get_all_nondangling(self):
...
 
+  def test_copy_tensor(self):
+    net = tensornetwork.TensorNetwork()
+    a = net.add_node(np.array([1, 2, 3], dtype=np.float64))
+    b = net.add_node(np.array([10, 20, 30], dtype=np.float64))
+    c = net.add_node(np.array([5, 6, 7], dtype=np.float64))
+    d = net.add_node(np.array([1, -1, 1], dtype=np.float64))
+    cn = net.add_copy_node(rank=4, dimension=3)
+    edge1 = net.connect(a[0], cn[0])
+    edge2 = net.connect(b[0], cn[1])
```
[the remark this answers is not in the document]
Thenerdstation: Ignore that, I that the values were the shapes.

tensornetwork/network_components.py
```diff
@@ -232,6 +237,82 @@ def __str__(self) -> Text:
...
+    # TODO: Make this computation lazy, once Node doesn't require tensor
+    # at instatiation.
+    copy_tensor = self.make_copy_tensor(rank, dimension)
+    copy_tensor = backend.convert_to_tensor(copy_tensor)
+    super().__init__(copy_tensor, name, axis_names, backend)
+
+  @staticmethod
+  def make_copy_tensor(rank: int, dimension: int) -> Tensor:
+    shape = (dimension,) * rank
+    copy_tensor = np.zeros(shape)
```
Thenerdstation: We'll have to deal with the `dtype` as well. Not sure whether to do that at the `TensorNetwork` or `make_copy_tensor` level.

viathor: Done.

tensornetwork/network_components.py
```diff
@@ -232,6 +237,82 @@ def __str__(self) -> Text:
...
+  def _get_partners(self) -> List[Tuple[Node, int]]:
+    partners = []
+    for edge in self.edges:
+      if edge.is_dangling() or self._is_my_trace(edge):
+        continue
+      partner_node, shared_axis = self._get_partner(edge)
+      partners.append((partner_node, shared_axis))
+    return partners
+
+  _VALID_LETTERS = list('abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ')
```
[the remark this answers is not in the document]
viathor: Done.

tensornetwork/network_components.py
viathor: I agree. This method doesn't make any of the re-arrangements of the network that contraction entails. Instead, it computes the tensor that we'll need in the implementation of `contract_copy_tensor()` in `TensorNetwork`. I plan this as a follow-up PR.

Renamed.

tensornetwork/tensornetwork_test.py
```diff
@@ -884,6 +884,29 @@ def test_get_all_nondangling(self):
...
+    edge3 = net.connect(c[0], cn[2])
+    edge4 = net.connect(d[0], cn[3])
+
+    result = cn.contract_all_edges()
+    self.assertEqual(result.shape, [])
+    self.assertAllClose(result, 50 - 240 + 630)
+
+    for edge in [edge1, edge2, edge3, edge4]:
+      net.contract(edge)
```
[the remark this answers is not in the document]
viathor: They do since we're not actually contracting anything on l.899 above. Renamed the function to make this clearer.

tensornetwork/network_components.py
Thenerdstation: Aww I see. Ok cool sgtm. I would still recommend a name change but the structure seems ok.

viathor: I did rename it :-)

New name: `compute_contracted_tensor()`.

Thenerdstation: Yes you did. My bad.

tensornetwork/network.py
```diff
@@ -108,6 +108,40 @@ def add_node(
     self.nodes_set.add(new_node)
     return new_node
 
+  def add_copy_node(
+      self,
+      rank: int,
+      dimension: int,
+      name: Optional[Text] = None,
+      axis_names: Optional[List[Text]] = None,
+      dtype: Type[np.number] = np.float64) -> network_components.CopyNode:
```
Thenerdstation: This reminds me, we should add a `dtype` to TensorNetwork to catch errors early. No changes for now though.

PR #67, opened 2019-05-20
Multi backend

Thenerdstation:
Have `tensornetwork_test` run against all backends and moved tensorflow specific backend tests to `graphmode_tensornetwork_test.py`


tensornetwork/graphmode_tensornetwork_test.py
```diff
@@ -38,6 +39,65 @@ def test_gradient_decent(self):
       sess.run(train_op)
       self.assertLess(sess.run(final_tensor), 10.0)
 
+  def test_dynamic_network_sizes(self):
```
stavros11: Would it be a good idea to have the eager tests in a seperate class or change the name of this file since not everything is graph mode anymore?

tensornetwork/network.py
```diff
@@ -20,19 +20,21 @@
...
-
 Tensor = Any
 
 
 class TensorNetwork:
   """Implementation of a TensorNetwork."""
 
-  def __init__(self) -> None:
+  def __init__(self, backend: Optional[Text] = None) -> None:
     # TODO(chaseriley): Allow variable backend and default to global
```
stavros11: TODO is to be deleted since it is fixed.

tensornetwork/tensornetwork_test.py
```diff
@@ -16,874 +16,837 @@
 from __future__ import division
 from __future__ import print_function
 import tensornetwork
+import pytest
+import unittest
```
stavros11: Is `import unittest` needed? (haven't tried to run this yet)

tensornetwork/graphmode_tensornetwork_test.py
Thenerdstation: Yeah that's a good point. 

Since these are all tensorflow specific tests, I think it makes sense to move this to backends/tensorflow. Maybe call it `tensorflow_tensornetwork_test.py`?

tensornetwork/tensornetwork_test.py
Thenerdstation: No it is not. Good catch.

tensornetwork/network.py
Thenerdstation: Done

stavros11:
Not directly related to this PR but found it by trying to run the tests:

I think there is an issue with [`merge_networks` method](https://github.com/google/TensorNetwork/blob/e7e911294a8f9098b11605ed0e7c95a4f74887a4/tensornetwork/network.py#L70) because the class method gets initialized in the default backend regardless of the given networks backend. Not sure what's the best way to fix this but perhaps the easiest is to add a line like `new_network.backend = next(iter(networks)).backend`. Perhaps we should also test whether the given networks have the same backend or think how to treat the case of different backends (error?).

A smaller thing: In order to pass `tensornetwork_test.py` locally I have to add `tf.enable_v2_behavior`. I guess this might depend on the local TF version so perhaps we don't need it since the checks pass.

Thenerdstation:
That's really strange. The tests shouldn't pass if v2 isn't enabled.

It might be the case that when running all of the tests, v2 gets enabled else where and it's ok. But we should def add it.

And that's also true, we should check the backend types for `merge_networks`. They should all be the same, and if not raise an error. The merged network backend should be the same as the rest of the backends.

PR #68, opened 2019-05-21
Jax backend

Thenerdstation:


Thenerdstation:
Yeah that's a good observation. I think we should change those tests to be deterministic. Will do that in the next CL.

PR #69, opened 2019-05-21
Kill transposes 2

amilsted:
This adds a couple more tweaks to tensordot2 and also, perhaps more controversially, adds an `output_edge_ordering` argument to `contract_between()`. 

The latter allows `contract_between()` to flip the argument ordering in the call to `tensordot()` in case this would minimize transposition of the output tensor. For this to be possible, it must know the desired output ordering. This is used in `ncon()`.

Probably I should add some explicit tests for this is `tensornetwork_test`. Am happy to do that.

### DIFF
--- a/tensornetwork/backends/tensorflow/tensordot2.py
+++ b/tensornetwork/backends/tensorflow/tensordot2.py
@@ -106,7 +106,7 @@ def tensordot(a, b, axes, name=None):
       free_dims = [shape_a[i] for i in free]
       prod_free = int(np.prod([shape_a[i] for i in free]))
       prod_axes = int(np.prod([shape_a[i] for i in axes]))
-      perm = list(axes) + free if flipped else free + list(axes)
+      perm = axes + free if flipped else free + axes
       new_shape = [prod_axes, prod_free] if flipped else [prod_free, prod_axes]
 
       # Skip the transpose op if possible. Although the graph optimizer should
@@ -123,6 +123,7 @@ def tensordot(a, b, axes, name=None):
       axes = [i if i >= 0 else i + len(shape_a) for i in axes]
       free = [i for i in range(len(shape_a)) if i not in axes]
       flipped = _tensordot_should_flip(axes, free)
+      perm = axes + free if flipped else free + axes
 
       axes_dims = [shape_a[i] for i in axes]
       free_dims = [shape_a[i] for i in free]
@@ -130,6 +131,12 @@ def tensordot(a, b, axes, name=None):
       axes = tf.convert_to_tensor(axes, dtype=tf.dtypes.int32, name="axes")
       free = tf.convert_to_tensor(free, dtype=tf.dtypes.int32, name="free")
       shape_a = tf.shape(a)
+
+      # TODO(amilsted): Handle tensor-valued `axes`!
+      if perm == list(range(len(perm))):
+        transposed_a = a
+      else:
+        transposed_a = tf.transpose(a, perm)
     else:
       free_dims_static = None
       shape_a = tf.shape(a)
@@ -140,6 +147,8 @@ def tensordot(a, b, axes, name=None):
       # Matmul does not accept tensors for its transpose arguments, so fall
       # back to the previous, fixed behavior.
       flipped = is_right_term
+      perm = tf.concat([axes, free], 0) if flipped else tf.concat([free, axes], 0)
+      transposed_a = tf.transpose(a, perm)
 
     free_dims = tf.gather(shape_a, free)
     axes_dims = tf.gather(shape_a, axes)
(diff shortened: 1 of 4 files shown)


Thenerdstation:
> perhaps more controversially, adds an output_edge_ordering argument to contract_between().

Doesn't seem controversial at all to me. 
>Probably I should add some explicit tests for this is tensornetwork_test. Am happy to do that.

Yes, always add tests. :)

amilsted:
> Doesn't seem controversial at all to me.

Good :) It irked me a little to add a second way of achieving an output ordering, but I guess it's not like we're adhering to "TOOWTDI"* elsewhere.

Tests are in.

*just found out there's an acronym for this...

Thenerdstation:
Well, the reason it doesn't break TOOWTDI is because you only do it if you have to. It's a necessary optimization.

tensornetwork/backends/tensorflow/tensordot2.py
```diff
@@ -72,10 +72,30 @@ def _tensordot_should_flip(contraction_axes, free_axes):
       should_flip: `True` if `contraction_axes` should be moved to the left,
         `False` if they should be moved to the right.
     """
+    # NOTE: This will fail if the arguments contain any Tensors.
     if contraction_axes and free_axes:
       return bool(np.mean(contraction_axes) < np.mean(free_axes))
     return False
 
+  def _tranpose_if_necessary(tensor, perm):
```
Thenerdstation: Please add typing.

tensornetwork/backends/tensorflow/tensordot2.py
```diff
@@ -72,10 +72,30 @@ def _tensordot_should_flip(contraction_axes, free_axes):
...
+    """Like transpose(), but avoids creating a new tensor if possible.
+    Although the graph optimizer should kill trivial transposes, it is best not
+    to add them in the first place!
+    """
+    # If perm is tensor-valued, test will fail and we fall back to tranposing.
+    if perm == list(range(len(perm))):
+      return tensor
+    return tf.transpose(tensor, perm)
+
+  def _reshape_if_necessary(tensor, new_shape):
```
Thenerdstation: Typing.

tensornetwork/backends/tensorflow/tensordot2.py
```diff
@@ -99,37 +120,34 @@ def _tensordot_reshape(a, axes, is_right_term=False):
     """
     if a.get_shape().is_fully_defined() and isinstance(axes, (list, tuple)):
       shape_a = a.get_shape().as_list()
+      # NOTE: This will fail if axes contains any tensors
```
Thenerdstation: I don't think we should ever have axes be tensors, but if you add typing we can make sure that never happens ;)

tensornetwork/backends/tensorflow/tensordot2.py
```diff
@@ -180,13 +203,15 @@ def _tensordot_axes(a, axes):
           isinstance(b_axes, tf.compat.integral_types):
         a_axes = [a_axes]
         b_axes = [b_axes]
+      # NOTE: This fails if either a_axes and b_axes are Tensors.
```
Thenerdstation: Axes will likely never be Tensors, so go ahead and remove this comment.

tensornetwork/network.py
```diff
@@ -713,13 +717,40 @@ def contract_between(
...
 
+    if output_edge_order:
+      # Determine heuristically if output transposition can be minimized by 
+      # flipping the arguments to tensordot.
+      node1_output_axes = []
+      node2_output_axes = []
+      for (i, edge) in enumerate(output_edge_order):
+        if edge in shared_edges:
+          raise ValueError(
+            "Edge '{}' in output_edge_order is shared by the nodes to be "
```
Thenerdstation: Can you add a test for this case?

tensornetwork/network.py
```diff
@@ -713,13 +717,40 @@ def contract_between(
...
+            "contracted: '{}' and '{}'.".format(edge, node1, node2)
+          )
+        edge_nodes = set(edge.get_nodes())
+        if node1 in edge_nodes:
+          node1_output_axes.append(i)
+        elif node2 in edge_nodes:
+          node2_output_axes.append(i)
+        else:
+          raise ValueError(
+            "Edge '{}' in output_edge_order is not connected to node '{}' or "
```
Thenerdstation: Add test for this case too. Generally, strive to have 100% test converge when ever possible. It makes refactoring and fixing bugs waaaaayyyyy easier in the long run.

amilsted:
So, I've been trying to keep the tensordot2 code close to the TF tensordot code so that it can be pushed upstream at some point. I could add typing, but the TF code does not use it. This also affects the assumptions I can make about input arguments. For example, tensordot supports `axes` being a single Tensor, but TensorNetwork never uses this.

I can certainly still add typing to tensordot2 if you deem it necessary.

tensornetwork/network.py
amilsted: This is covered by the test I added.

tensornetwork/network.py
amilsted: This is also covered by the test I added.

tensornetwork/backends/tensorflow/tensordot2.py
amilsted: See main thread.

Thenerdstation:
Go ahead and add the typing. Removing it when you make an upstream PR should be pretty easy. 

If you want to include `tf.Tensor` as one of the types to `axes`, I would be ok with that.

Thenerdstation:
Oh no a flaky test.

Why did I use np.random. let me fix that

PR #84, opened 2019-05-30
Binary mera

mganahl:
This PR adds an optimization of a scale-invariant binary MERA network to approximate
the ground-state of the transverse field Ising model at criticality.

it includes:
1. benchmarks for ascending and descending operations
2. benchmark for optimizations
3. code for computing scaling dimensions

Additionally, almost all contractions are carried out by directly using the TensorNetwork API
(instead of the ncon wrapper).


Thenerdstation:
Let me know when you are ready to merge this.

mganahl:
Hey @Thenerdstation, I think this can be merged

Thenerdstation:
looks good afaict. Can you add a small unit test? That will ensure this code doesn't die when we do inevitable changes.

mganahl:
sure, what's the preferred way of testing (i.e. unittest, pytest, tensorflow test suite)?

Thenerdstation:
We use pytest, but pytest can also run all 3. do what you think is easiest.

experiments/MERA/test_binary_mera.py
```diff
@@ -0,0 +1,131 @@
+# Copyright 2019 The TensorNetwork Authors
```
Thenerdstation: change to binary_mera_test.py for consistency.

mganahl: renamed and added a few more tests.

PR #89, opened 2019-06-03
Implement copy node contraction

viathor:
There are two parts to contracting a copy node (i.e. a node representing a copy tensor). First, you need to compute the tensor resulting from simultaneously contracting all edges of a copy node. Second, you need to re-arrange the network to incorporate the new node and remove the old nodes. The former was done in #64. The latter is done in this PR.

Follow-up: This will help implement a new contractor based on Bucket Elimination (aka Variable Elimination). In that contractor we exploit the fact that variables (indices) to be eliminated in general correspond to copy nodes, not just single-edges.

### DIFF
--- a/tensornetwork/tensornetwork_test.py
+++ b/tensornetwork/tensornetwork_test.py
@@ -959,7 +959,6 @@ def test_contract_copy_node(backend):
 def test_contract_copy_node_connected_neighbors(backend):
   net = tensornetwork.TensorNetwork(backend=backend)
   a = net.add_node(np.array([[1, 2, 3], [10, 20, 30]], dtype=np.float64))
-  assert a.tensor.shape == (2, 3), a.tensor.shape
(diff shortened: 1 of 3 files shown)


tensornetwork/tensornetwork_test.py
```diff
@@ -938,6 +938,49 @@ def test_copy_tensor_parallel_edges(backend):
   assert list(result.shape) == []
   np.testing.assert_allclose(result, 10 + 40 + 90)
 
+def test_contract_copy_node(backend):
```
Thenerdstation: Why doesn't this break when using tensorflow?

viathor: Tensorflow is not entirely unusable with copy tensors! It only errors out when you happen to have a node A and a copy node C with *more than one* edge between A and C. Otherwise it works just fine!

Issue #50, opened 2019-05-15
Create Bucket elimination algorithm


Thenerdstation:
Bucket elimination is a fast algorithm that works well for networks with small "tree width". 

Positive examples include: TTN, quantum circuits, MPS, and MERA.

Negative examples include: SAT TensorNetwork.

Seeing how most of our work involves the former, we should implement this algorithm asap. @viathor has been working on a first version.

viathor:
Note that BE contractor exploits sparsity in the tensors by reducing the number of indices to range over in cases where terms corresponding to most index combinations are zero. This arises very often in quantum circuits where a lot of tensors are diagonal.

Treewidth sets the lower bound for contraction complexity.

This should work fairly well on quantum circuits and SAT tensor network. For the other types, a lot depends on the tensors, but note that performing SVD on an edge does introduce a diagonal tensor that BE contractor can exploit.

PR #92, opened 2019-06-03
Implement Bucket-Elimination-based contractor

viathor:
The core of the logic for BE contractor is already implemented by #64 and #89. This just wraps it up in a contractor interface.

Note that like many other contraction algorithms, this one is very sensitive to contraction order. For this reason, we let the caller decide it. There are some heuristics in the literature which we should implement. Once we have them one of the can be made default to simplify the API by not requiring the second argument.

Context: #50.


tensornetwork/contractors/bucket_contractor_test.py
```diff
@@ -0,0 +1,113 @@
...
+  net.connect(q0, control[0])
+  net.connect(q1, target[0])
+  net.connect(control[1], target[1])
+  return control, control[2], target[2]
+
+
+class BucketTest(tf.test.TestCase):
+
+  def test_cnot_gate(self):
+    net = network.TensorNetwork(backend="tensorflow")
```
Thenerdstation: Convert this to just use pytest (Remove tf.test.TestCase) and pass in the backends. None of your tests require tensorflow directly, so this will make sure it's always compatible with the other backends.

tensornetwork/contractors/bucket_contractor_test.py
```diff
@@ -0,0 +1,113 @@
...
+    # Build quantum circuit
+    copy_node, q0_t1, q1_t1 = add_cnot(net, q0_in[0], q1_in[0])
+    net.connect(q0_t1, q0_out[0])
+    net.connect(q1_t1, q1_out[0])
+    # Contract the network, first using Bucket Elimination, then once
+    # no more copy tensors are left to exploit, fall back to the naive
+    # contractor.
+    contraction_order = (copy_node,)
+    net = bucket(net, contraction_order)
+    net = naive(net)
```
Thenerdstation: This feels weird. Naive shouldn't work if any of the nodes were contracted. Why doesn't this hit the ValueError in naive?

tensornetwork/contractors/bucket_contractor_test.py
```diff
@@ -0,0 +1,113 @@
...
+    result = net.get_final_node()
+    # Verify that CNOT has turned |11> into |10>.
+    self.assertAllClose(result.get_tensor(), 1.0)
```
Thenerdstation: np.testing.assert_allclose (same below)

viathor: Done.

tensornetwork/contractors/bucket_contractor_test.py
viathor: `TensorNetwork.contract_copy_node()` removes edges it contracts from `self.edge_order` in order to enable other contractors like naive to be used to finish contracting tensors that BE doesn't take care of.

tensornetwork/contractors/bucket_contractor_test.py
viathor: Done.

Thenerdstation:
Squash and merge if you are ready.

viathor:
Accidentally merged all commits to master. Sorry.

I disabled "Allow merge commits" in project settings to prevent this in future. You can now choose between squash or rebase (copying all commits is disabled).

Thenerdstation:
That's ok, I did that before too. Thanks for flipping the disable switch!

Issue #82, opened 2019-05-30
Derivatives and environments


amilsted:
If we define a tensor network for e.g. a scalar value (no dangling edges) and we want to compute the derivative of that number with respect to a tensor `T` in the network, one could use autodiff (depending on the backend). This is also known as computing the "environment" of `T`.

However, the first derivative is also given by the contraction of the same network with the `T` tensor removed. For doing this, it would be nice to have a `remove_node()` method that deletes a node from the network. Any dangling edges attached to that node are also removed. Any connected edges become dangling edges (modified in place, so that the edge objects persist).

This would be useful for a number of algorithms: One can define the network for, say, the energy of a quantum state once, then compute all required environments/derivatives using `remove_node()`. These are then used to minimize the energy.

Going further, if multiple environments of the same network are desired, it is possible to take the optimal contraction order for one environment and derive optimal contraction orders for all the others: https://arxiv.org/abs/1310.8023
Intermediate results can also often be reused for multiple environments.

One might imagine having a method `environments([n1, n2, n3])` (where `n1,n2,n3` are nodes) that efficiently computes environments of multiple tensors this way.

Thoughts?

amilsted:
By the way, I suspect (although I have not checked) that using the proposed `environments()` has the same asymptotic computational time scaling as using `tf.gradients()`. Hence it might only be advantageous for backends without autodiff functionality. On the other hand, `environments()` might be easier to fit into any upcoming automatic multi-device support?

mganahl:
I think `remove_node()` is a great idea. It would reduce the length of MERA code (and also others) a lot. It would be good to also have the automated optimal contractor then. @Thenerdstation what is the status on that? I saw the stochastic contractor. Is someone working on a deterministic optimal cotractors?

Thenerdstation:
Adam is working on it, but he is a 20%er. I've asked him to add a branch of his work so far, so hopefully that'll be added soon.

Thenerdstation:
Remove node should be easy enough to add. Though I think we'll use `disconnect` to break the edges rather than modify the edges in place. This is consistent with the rest of the code base.

amilsted:
@Thenerdstation The only trouble I see with replacing the connected edges is that it makes it tricky to keep track of them across the `remove_node()`. In particular, I would want to be able to know which of the new dangling edges was connected to which axis of the removed node. Ideally, I would not have to remember the axis ordering for this to work - the axis names would be enough.

[...]

PR #101, opened 2019-06-08
Added remove_node method

Thenerdstation:
This was requested in https://github.com/google/TensorNetwork/issues/82 and can be used for calculating derivatives and environments for backends without auto differentiation (i.e. pure numpy).

### DIFF
--- a/tensornetwork/network.py
+++ b/tensornetwork/network.py
@@ -17,7 +17,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 import collections
-from typing import Any, Sequence, List, Set, Optional, Union, Text, Tuple, Type
+from typing import Any, Sequence, List, Set, Optional, Union, Text, Tuple, Type, Dict
 import numpy as np
 import weakref
 from tensornetwork import config
@@ -912,6 +912,31 @@ class TensorNetwork:
     self.connect(singular_values_node[1], right_node[0])
     self.nodes_set.remove(node)
     return left_node, singular_values_node, right_node, trun_vals
+
+  def remove_node(self, node: network_components.Node
+    ) -> Dict[Union[int, Text], network_components.Edge]:
+    """Remove a node from the network.
+
+    Args:
+      node: The node to be removed.
+
+    Returns:
+      A dictonary mapping name/index to the new broken edge.
+
+    Raises:
+      ValueError: If the node isn't in the network.
+    """
+    if node not in self:
+      raise ValueError("Node '{}' is not in the network.".format(node))
+    broken_edges = {}
+    for i, name in enumerate(node.axis_names):
+      if not node[i].is_dangling() and not node[i].is_trace():
+        edge1, edge2 = self.disconnect(node[i])
+        new_broken_edge = edge1 if edge1.node1 is not node else edge2
+        broken_edges[i] = new_broken_edge
+        broken_edges[name] = new_broken_edge
+    return broken_edges
+
 
   def check_correct(self, check_connected: bool = True) -> None:
     """Check that the network is structured correctly.
--- a/tensornetwork/network.py
+++ b/tensornetwork/network.py
@@ -915,7 +915,9 @@ class TensorNetwork:
     return left_node, singular_values_node, right_node, trun_vals
 
   def remove_node(self, node: network_components.Node
-    ) -> Dict[Union[int, Text], network_components.Edge]:
+    ) -> Tuple[
+      Dict[Text, network_components.Edge],
+      Dict[int, network_components.Edge]]:
     """Remove a node from the network.
 
     Args:
--- a/tensornetwork/network.py
+++ b/tensornetwork/network.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 import collections
+# pylint: disable=line-too-long
 from typing import Any, Sequence, List, Set, Optional, Union, Text, Tuple, Type, Dict
 import numpy as np
 import weakref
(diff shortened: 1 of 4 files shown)


tensornetwork/network.py
```diff
@@ -913,6 +914,32 @@ def split_node_full_svd(
...
+      raise ValueError("Node '{}' is not in the network.".format(node))
+    broken_edges = {}
+    for i, name in enumerate(node.axis_names):
+      if not node[i].is_dangling() and not node[i].is_trace():
+        edge1, edge2 = self.disconnect(node[i])
+        new_broken_edge = edge1 if edge1.node1 is not node else edge2
+        broken_edges[i] = new_broken_edge
+        broken_edges[name] = new_broken_edge
+    self.nodes_set.remove(node)
+    return broken_edges
```
viathor: An alternative would be to return two dictionaries: one with broken edged by axis number and one with broken edges by axis name. I think the latter API is more convenient since in the current one the caller iterating over the dictionary must check key type if they're only interested in numbers/names. Also, they may accidentally double-process an edge.

tensornetwork/network.py
```diff
@@ -913,6 +914,32 @@ def split_node_full_svd(
...
+      A dictonary mapping name/index to the new broken edge.
+
+    Raises:
+      ValueError: If the node isn't in the network.
+    """
+    if node not in self:
+      raise ValueError("Node '{}' is not in the network.".format(node))
+    broken_edges = {}
+    for i, name in enumerate(node.axis_names):
+      if not node[i].is_dangling() and not node[i].is_trace():
```
viathor: After `remove_node()` the dangling and trace edges of the node are ripe for garbage collection. However, by leaving them in the state where they refer to each other we create reference cycles that aren't handled by the reference counting mechanism.

I think the garbage collector may still handle those, but it's easy and probably safer to just disconnect the dangling and trace edges from the node here. WDYT?

Thenerdstation: So, since edges have weak reference to their nodes, the only strong references are Node -> Edge, so once the Node is garbage collected, the edges will be too.

I don't want to fully disconnect these edges, as it is conceivable that someone might want to reconnect that node to another part of the network.

tensornetwork/network.py
Thenerdstation: That's a really good point. However, that means that one of the dictionaries will be ignored most of the time, so we'll be seeing a lot of code that looks like `broke_edges, _ = net.remove_node(node)`

This isn't really that bad, and would certainly be preferable to checking the types when looping over the dictionary. Give me a day to think about it some more. If I can't figure out a better solution we'll do the two dictionaries.

Thenerdstation: @amilsted @mganahl I would love to have your opinions on this.

amilsted: Tricky! Although it seems kind of ugly to return two dictionaries, but I agree that having two instances of each edge in the dict is asking for trouble. I also agree that it's not a big deal if one of the dictionaries is ignored, so I guess I think it would indeed be better to return two.

amilsted: I suppose an alternative would be to have an argument determine whether you get names or indices as keys, e.g. `return_axes_by_name=True`.

viathor: I like the solution with an extra argument, but it hides a minor issue.

Making the argument a boolean spells some trouble with the type-checker down the road, since mypy won't be able to tell the return type of a given invocation.

We could solve this problem by defining two special values for the extra argument and using [overloads](https://mypy.readthedocs.io/en/latest/more_types.html#function-overloading):

```python
BY_NUMBER=-1
BY_NAME=""

@overload
def remove_node(self, node: Node, return_axes: int) -> Dict[int, Edge]: ...

@overload
def remove_node(self, node: Node, return_axes: str) -> Dict[str, Edge]: ...

def remove_node(self, node: Node, return_axes: Union[int, str]=BY_NAME) -> Union[Dict[int, Edge], Dict[str, Edge]]:
    if node...
    ...
    if return_axes is BY_NAME:
        return broken_edges_by_name
    return broken_edges_by_number

```

This way mypy will know that the key in the returned dictionary has the same type as the value passed in as `return_axes` argument (which is supposed to be either `BY_NUMBER` or BY_NAME which are an `int` and a `str` respectively).

Not sure whether default values play nicely with overloads, though.

Thenerdstation: Yikes, I don't like that at all. It's pretty confusing and requires even more characters than `, _`. Alos, where would we store `BY_NAME`? Would it become a constant accessible via `tensor_network`? I don't wanna add a constant for this single method. 

Let's just return two dictionaries. Adding a `, _` really isnt that bad.

tensornetwork/network.py
```diff
@@ -913,6 +914,38 @@ def split_node_full_svd(
...
+      Dict[Text, network_components.Edge],
+      Dict[int, network_components.Edge]]:
+    """Remove a node from the network.
+
+    Args:
+      node: The node to be removed.
+
+    Returns:
+      broken_edges_by_name: A Dictionary mapping `node`'s axis names 
+        the newly broken edges.
```
viathor: nit: *to* the newly broken edges (missing "to")

tensornetwork/network.py
```diff
@@ -913,6 +914,38 @@ def split_node_full_svd(
...
+      broken_edges_by_axis: A Dictionary mapping `node`'s integer axis 
+        values to the newly broken edges.
```
viathor: nit: replace "integer axis values" with "axis numbers" - it reads better

Thenerdstation:
Aww shoot I just saw your comments.

I'll do another PR with your fixes. Github is slightly different from piper in that if I change anything after you approve, you have to approve again.

viathor:
Interesting. It must be something peculiar to this repo since the I-trust-you-to-address-comments-so-merge-without-me-checking workflow works fine in other repos. I'll see if I can find the cause later.

Issue #72, opened 2019-05-22
Create Shell Backend


Thenerdstation:
Create a shell backend for "ShellTensors" that only store their shape.

stavros11:
I would be interested in looking into this and/or #75. I guess the plan is to have the current `BaseBackend` as a reference and replace all its methods by "shape methods". Perhaps we might need to modify that later depending on the needs of the ContractionNetwork. Let me know if you have any other ideas.

Thenerdstation:
Great!

I would recommend doing this first as #75 will require this.

PR #112, opened 2019-06-15
Shell Backend

stavros11:
Not sure if these issues are still relevant but that's a first try for #72 .
It treats `Tensor` as a tuple that has the corresponding tensor shape but depending on how the conversion from tensor to shape will work, we may need to change that.
SVD is not implemented because I am not sure how we should treat the truncation size when we do not have the actual tensor values to find singular values and compare to cut-off.

### DIFF
--- a/tensornetwork/backends/shell/shell_backend.py
+++ b/tensornetwork/backends/shell/shell_backend.py
@@ -15,8 +15,11 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
-from typing import Optional, Sequence, Tuple, List
+
+import functools
+import operator
 from tensornetwork.backends import base_backend
+from typing import Optional, Sequence, Tuple, List
 
 # Treat tensors as tuples that carry the real tensor's shape.
 # Conversion from tensor to tuple will happen somewhere else?
@@ -56,16 +59,14 @@ class ShellBackend(base_backend.BaseBackend):
     raise NotImplementedError("SVD shape cannot be calculated without"
                               "explicit tensor values.")
 
-  def concat(self, values: Sequence[Tensor], axis: int) -> Tensor:
-    # Does not work when axis < 0
-    concat_dim = len(values) * values[0][axis]
-    return values[0][:axis] + (concat_dim,) + values[0][axis + 1:]
+  def shape_concat(self, values: Sequence[Tensor]) -> Tensor:
+    return functools.reduce(operator.concat, values)
 
   def shape(self, tensor: Tensor) -> Tensor:
     return tensor
 
-  def prod(self, values: Tensor) -> Tensor:
-    return values
+  def shape_prod(self, values: Tensor) -> Tensor:
+    return functools.reduce(operator.mul, values)
 
   def sqrt(self, tensor: Tensor) -> Tensor:
     return tensor
--- a/tensornetwork/backends/shell/shell_backend.py
+++ b/tensornetwork/backends/shell/shell_backend.py
@@ -27,6 +27,7 @@ class ShellBackend(base_backend.BaseBackend):
   """See base_backend.BaseBackend for documentation."""
 
   def __init__(self):
+    super(ShellBackend, self).__init__()
     self.name = "shell"
 
   def tensordot(self, a: Tensor, b: Tensor, axes: Sequence[Sequence[int]]):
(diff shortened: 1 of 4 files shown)


tensornetwork/backends/shell/shell_backend.py
```diff
@@ -0,0 +1,107 @@
...
+
+  def concat(self, values: Sequence[Tensor], axis: int) -> Tensor:
+    # Does not work when axis < 0
+    concat_dim = len(values) * values[0][axis]
+    return values[0][:axis] + (concat_dim,) + values[0][axis + 1:]
+
+  def shape(self, tensor: Tensor) -> Tensor:
+    return tensor
+
+  def prod(self, values: Tensor) -> Tensor:
```
Thenerdstation: So `prod` is actually only used for shape calculations. Same for concat. (We really should just rename these methods to shape_prod and shape_concat to avoid future confusion). 

We likely will need to modify these methods so they work well with our current implementations.

stavros11: Actually `prod` was wrongly implemented anyway. I guess since these methods are only used for shape calculations in `network` they should do the same in the shell backend, that's how I did it now.

stavros11:
Do you think it is a good idea to take the shape tuple as the "Shell tensor"? Because it seems quite difficult to integrate this as it is (for example use `tensornetwork_test.py`).

Thenerdstation:
Yeah, the tuple might be too restrictive of a data structure, especially since there is no real way to extend it.

Can you try building a `ShellTensor` class? That way `tensor.shape` could still work.

tensornetwork/backends/shell/shell_backend.py
```diff
@@ -0,0 +1,129 @@
...
+    tensor = tensor.reshape(shape)
+    return tensor
+
+  def svd_decomposition(self,
+                        tensor: Tensor,
+                        split_axis: int,
+                        max_singular_values: Optional[int] = None,
+                        max_truncation_error: Optional[float] = None
+                       ) -> Tuple[Tensor, Tensor, Tensor, Tensor]:
+    raise NotImplementedError("SVD shape cannot be calculated without"
```
Thenerdstation: I think you should be able to implement this (You can leave it for a later PR), but everything except for max_truncation_error doesn't depend on the concrete tensor.

tensornetwork/backends/shell/shell_backend.py
```diff
@@ -0,0 +1,129 @@
...
+class ShellTensor:
+  def __init__(self, shape: Tuple):
+    self.shape = shape
+
+  def reshape(self, new_shape: Tuple):
+    self.shape = new_shape
+    return self
+
+
+Tensor = "ShellTensor"
```
Thenerdstation: You should just be able to use ShellTensor directly.

tensornetwork/backends/shell/shell_backend_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+def assertShapesEqual(a, b, rtol=1e-8):
+  assert get_shape(a) == get_shape(b)
+
+@pytest.fixture(
+  name="backend", params=["shell"])
+def backend_fixure(request):
+    return request.param
+
+# Override np.testing to check only shapes and not values
```
Thenerdstation: I'd rather you'd explictly mark the tests as a skip for the shell backend rather than doing this weird `set(dir(...))` hack that will become difficult to maintain.

stavros11: Sorry, just saw this comment. The issue is that I don't want to skip the tests. I want to run all the `tensornetwork_test.py` tests but instead of doing `np.testing.assert_allclose` only test that the shapes agree. The hack seems to screw up other tests so it is probably a bad idea...

stavros11:
I am not sure how to `tensornetwork_test` can be used for the shell backend, because of the `assert_allclose` issue, so I decided to write from scratch tests that compare shapes between the shell and numpy backends. Also implemented SVD when `max_truncation_error` is not given.

Thenerdstation:
Thanks! This looks good.

Thenerdstation:
You'll need to add the `shape_tuple` method for the greedy contractor (this was just merged today). Should be easy, just return the shape.

PR #117, opened 2019-06-19
Upgrades to MPS and MERA

mganahl:
This PR implements:

1. fix import issues which caused code to break
2. add unit tests for MPS and DMRG 
3. for binary MERA optimization, replace calls to `flatten_edges_between` with `contract_between` to avoid transposes
4. update docstrings for MPS modules


mganahl:
@Thenerdstation I  added some unit tests and fixed imports. The MPS stuff should now be working again. I also added docstrings to the most important routines.

Thenerdstation:
I was gonna let it slide, but yikes the tests take 10 minutes now. Is there any way you can reduce the required time? (smaller bond dimensions, shorter MPS, etc?) If not, let's mark these as large and only run them on the Travis build.

mganahl:
10min! is that mostly because of the MPS tests? On my laptop running pytest finished in ~20s. 
I'm already using really small D and small systems (i.e. D=10 for infinite DMRG and D=32 and N=10 for finite DMRG). I will put the DMRG tests into a seperate dmrg_test.py. The remaining tests are much faster to finish. We can then mark dmrg_test.py as a big test

mganahl:
@Thenerdstation please wait for a bit longer before merging. I'll fix some more docstrings. I'll let you know once I'm ready

mganahl:
OK, this is ready from my side.

mganahl:
Changed indentation.

Issue #90, opened 2019-06-03
Should ncon allow "0" as a label?


amilsted:
I skimmed the [NCON paper](https://arxiv.org/abs/1402.0939) the other day and realized why other implementations tend to disallow `0` as a label for contraction edges: It is because the reference implementation of NCON treats `0` specially - it is used as part of a separate contraction order argument to specify when to carry out outer products (sometimes it is more efficient to do outer products as intermediate steps). This raised two issues for me:

1. Should we stop accepting `0` as a valid contraction edge label to ensure our ncon calls are compatible with the reference implementation? (Probably yes...)
2. Should we implement the (rather obscure) interpretation of `0` in the contraction order? (Probably no.)

Elaborating on 2: It's a lot of work for possibly very little gain. Instead just support using a specific smart contractor, or a saved contraction order, in `ncon()`: e.g. `ncon(..., contractor='optimal')`.

Thoughts?

mganahl:
1: Either way is fine with me. 2: probably no (I never used it so far)

Thenerdstation:
I guess my preference is to keep ncon consistent with the original interface.

For now, throw a `NotImplementedError` and maybe set this issue as a "Help wanted".

Thenerdstation:
Close this? I don't think we will be supporting 0 in the future.

amilsted:
Yeah.

PR #129, opened 2019-06-24
Kill zeros as ncon edge labels

amilsted:
Disallows `0` and an edge label (when contraction order is not specified manually) in `ncon()`. This forces `ncon()` calls to be compatible with other NCON implementations (such as the original: https://arxiv.org/abs/1402.0939), so that calls written for our API are compatible with other implementations. It also means we can support the special interpretation of "0" in the reference implementation in the future without breaking existing code. The value of supporting this is, however, probably not very high, and I don't expect we will ever do it.

Also worth noting: There may be other implementations of NCON that allow zeros, so this might make porting some code *to* our API more difficult. Alternatives would be to emit a warning, or add a note to the docstring warning that other implementations need not allow "0". 

**Thoughts?**

This also updates example and experiment code to eliminate such usage.


Thenerdstation:
fixes https://github.com/google/TensorNetwork/issues/90

Thenerdstation:
I think just raising an error is acceptable. Porting is always painful, but at least this way the user will know what is wrong right away.

Issue #110, opened 2019-06-14
Improve Documentation


Thenerdstation:
Currently, our readthedocs documentation is very lack luster. The formatting for some of the function docs looks pretty bad, and the layout for the sidebar needs improvement. We also need example code and some basic tutorials to help people get started (these can possibly be taken from the existing README).

mganahl:
Also, there are some errors thats should be fixed. I‘go through the docs and fix them.

shubham15gupta09:
I would like to contribute here . 
How can i edit it 
please help me as i am new to these things

PR #131, opened 2019-06-26
Added tutorial to documentation

Thenerdstation:
Added basic tutorial to help new users.
https://github.com/google/TensorNetwork/issues/110

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -93,7 +93,7 @@ This example code calculates the dot product of two vectors.
 
 Optimized Contractions
 ----------------------
-During computation, it’s very common for two nodes to have multiple edges connecting each other. If only one of the edges are contracted at a time, then all of the remaining edges become trace edges. This is usually very bad for computation, as the new node will allocate significantly more memory than required. Also, since trace edges only sum the diagonal of the underlying matrix, all of the other values calculated during the first contraction are useless. During contraction, it always more efficent to contract all of these edges at the same time.
+At intermediate states of a computation, it’s very common for two nodes to have multiple edges connecting them. If only one of those edges is contracted, then all of the remaining edges become trace edges. This is usually very inefficient, as the new node will allocate significantly more memory than is ultimately required. Since trace edges only sum the diagonal of the underlying matrix, all of the other values calculated during the first contraction are useless. It is always more efficient to contract all of these edges simultaneously.
 
 The methods `contract_between` or `contract_parallel` will do this for you automatically. You should see huge speedups when comparing these methods against contracting one edge at a time.
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -34,7 +34,7 @@ Trace edges connect a node to itself. To contract this type of edge, you take a 
 
 **Dangling Edge**
 
-Dangling edges are edges that only have one side point to a node, where as the other side is left “dangling”. These edges represent output axes or intermediate axes that have yet to be connected to other dangling edges. These edges are automatically created when adding a node to the network.
+Dangling edges are edges that only have one side point to a node, with the other side left “dangling”. These edges can represent output axes, or they can represent intermediate axes that have yet to be connected to other dangling edges. These edges are automatically created when adding a node to the network.
 
 .. code-block:: python
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1,0 +1,144 @@
+Tutorial
+========
+To get started, let's first install the TensorNetwork library.
+
+.. code-block:: bash
+
+  pip3 install tensornetwork
+
+
+Nodes
+-----
+Nodes are one of the basic building blocks of a tensor network. They represent a tensor in the computation. Each axis will have a corresponding edge that can possibly connect different nodes (or even the same node) together. The number of edges represents the order of the underlying tensor. For example, a node without any edges is a scalar, a node with one edge is a vector, etc.
+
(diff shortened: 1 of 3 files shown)


docs/tutorial.rst
```diff
@@ -0,0 +1,144 @@
...
+  import tensornetwork
+  import tensorflow as tf
+
+  net = tensornetwork.TensorNetwork()
+  node = net.add_node(tf.eye(2)) # NumPy arrays can also be used.
+  print(node.tensor)  # This is how you access the underlying tensor.
+
+Edges
+-----
+Edges describe different computations of the underlying tensors in the tensor network. Each edge points to which axes of the tensors to do the computation. There are 3 basic kinds of edges in a tensor network:
```
sleichen:
```suggestion
Edges describe different contractions of the underlying tensors in the tensor network. The edges are associated to the tensor axes involved in the contraction. There are 3 basic kinds of edges in a tensor network:
```

docs/tutorial.rst
```diff
@@ -0,0 +1,144 @@
...
+
+**Standard Edges**
+
+Standard edges are like any other edge you would find in an undirected graph. They connect 2 different nodes and define a dot product among the given vector spaces. In numpy terms, this edge defines a tensordot operation over the given axes.
```
sleichen:
```suggestion
Standard edges are like any other edge you would find in an undirected graph. They connect 2 different nodes and represent a dot product between the associated vector spaces. In numpy terms, this edge defines a tensordot operation over the given axes.
```

docs/tutorial.rst
```diff
@@ -0,0 +1,144 @@
...
+
+**Trace Edges**
+
+Trace edges connect a node to itself. To contract this type of edge, you take a trace of the matrix created by the two given axes.
```
sleichen:
```suggestion
Trace edges connect a node to itself. Contracting this type of edge means taking the trace of the matrix formed by the two associated axes.
```

docs/tutorial.rst
```diff
@@ -0,0 +1,144 @@
...
+
+**Dangling Edge**
+
+Dangling edges are edges that only have one side point to a node, where as the other side is left “dangling”. These edges represent output axes or intermediate axes that have yet to be connected to other dangling edges. These edges are automatically created when adding a node to the network.
```
sleichen:
```suggestion
Dangling edges are edges that only have one side point to a node, with the other side left “dangling”. These edges can represent output axes, or they can represent intermediate axes that have yet to be connected to other dangling edges. These edges are automatically created when adding a node to the network.
```

docs/tutorial.rst
```diff
@@ -0,0 +1,144 @@
...
+  dangling_edge = a[1]
+  # Create a standard edge by connecting any two separate nodes together.
+  # We create a new edge by "connecting" two dangling edges.
+  standard_edge = net.connect(a[0], b[0]) 
+  # Create a trace edge by connecting a node to itself.
+  trace_edge = net.connect(c[0], c[1])
+
+Connecting Dangling Edges 
+-------------------------
+One common paradigm in building quantum circuits is to add computation based on edges rather than nodes. That is, it’s very common to want to connect two dangling edges rather than trying to keep track of node axes.
```
sleichen:
```suggestion
One common paradigm in building quantum circuits, for example, is to add computation based on edges rather than nodes. That is, it’s very common to want to connect two dangling edges rather than trying to keep track of node axes.
```

docs/tutorial.rst
```diff
@@ -0,0 +1,144 @@
...
+  a = net.add_node(tf.ones(2))
+  b = net.add_node(tf.ones(2))
+  edge = net.connect(a[0], b[0])
+  c = net.contract(edge)
+  print(c.tensor.numpy()) # Should print 2.0
+
+
+Optimized Contractions
+----------------------
+During computation, it’s very common for two nodes to have multiple edges connecting each other. If only one of the edges are contracted at a time, then all of the remaining edges become trace edges. This is usually very bad for computation, as the new node will allocate significantly more memory than required. Also, since trace edges only sum the diagonal of the underlying matrix, all of the other values calculated during the first contraction are useless. During contraction, it always more efficent to contract all of these edges at the same time.
```
sleichen:
```suggestion
At intermediate states of a computation, it’s very common for two nodes to have multiple edges connecting them. If only one of those edges is contracted, then all of the remaining edges become trace edges. This is usually very inefficient, as the new node will allocate significantly more memory than is ultimately required. Since trace edges only sum the diagonal of the underlying matrix, all of the other values calculated during the first contraction are useless. It is always more efficient to contract all of these edges simultaneously.
```

docs/tutorial.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Doing either `contract_between` or `contract_parallel` will do this for you automatically. You should see huge speedups when comparing these methods against contracting one edge at a time.
```
sleichen:
```suggestion
The methods `contract_between` or `contract_parallel` will do this for you automatically. You should see huge speedups when comparing these methods against contracting one edge at a time.
```

Thenerdstation:
Not much we can do on the testing side other than "try to build it and fail on errors". We can do that, but I'm not sure it would catch much, especially any formatting errors.

sleichen:
> A Googler has manually verified that the CLAs look good.
> 
> (Googler, please make sure the reason for overriding the CLA status is clearly documented in these comments.)
> 

googlebot didn't recognize me so I overrode it

PR #133, opened 2019-06-27
Cleaned up documentation

Thenerdstation:
This adds some math equations and removes "return type" issues in some of the function documentation.

### DIFF
--- a/tensornetwork/network.py
+++ b/tensornetwork/network.py
@@ -458,7 +458,7 @@ class TensorNetwork:
       name: Optional name to give the new node created.
 
     Returns:
-      A new node. It's shape will be node1.shape + node2.shape
+      A new node. Its shape will be node1.shape + node2.shape
     """
     new_tensor = self.backend.outer_product(node1.tensor, node2.tensor)
     new_node = self.add_node(new_tensor, name)
@@ -683,7 +683,7 @@ class TensorNetwork:
   def get_parallel_edges(
       self, 
       edge: network_components.Edge) -> Set[network_components.Edge]:
-    """Get all of the edge parallel to the given `edge`.
+    """Get all of the edges parallel to the given `edge`.
 
     Args:
       edge: The given edge.
@@ -849,8 +849,8 @@ class TensorNetwork:
     Let M be the matrix created by flattening left_edges and right_edges into
     2 axes. Let :math:`U S V^* = M` be the Singular Value Decomposition of 
     :math:`M`. This will split the network into 2 nodes. The left node's 
-    tensor will be :math:`U * \\sqrt{S}` and the right node's tensor will be 
-    :math:`\\sqrt{S} * V^*` where :math:`V^*` is
+    tensor will be :math:`U \\sqrt{S}` and the right node's tensor will be 
+    :math:`\\sqrt{S} V^*` where :math:`V^*` is
     the adjoint of :math:`V`.
 
     The singular value decomposition is truncated if `max_singular_values` or
@@ -880,11 +880,12 @@ class TensorNetwork:
       A tuple containing:
         left_node: 
           A new node created that connects to all of the `left_edges`.
+          Its underlying tensor is :math:`U \\sqrt{S}`
         right_node: 
           A new node created that connects to all of the `right_edges`.
+          Its underlying tensor is :math:`\\sqrt{S} V^*`
         truncated_singular_values: 
-          A vector of the dropped smallest singular
-          values.
+          The vector of truncated singular values.
     """
     node.reorder_edges(left_edges + right_edges)
     u, s, vh, trun_vals = self.backend.svd_decomposition(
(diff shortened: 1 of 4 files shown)


tensornetwork/network.py
```diff
@@ -455,7 +458,7 @@ def outer_product(self,
       name: Optional name to give the new node created.
 
     Returns:
-      new_node: A new node. It's shape will be node1.shape + node2.shape
+      A new node. It's shape will be node1.shape + node2.shape
```
sleichen: Its

tensornetwork/network.py
```diff
@@ -678,13 +683,14 @@ def get_shared_edges(self, node1: network_components.Node,
   def get_parallel_edges(
       self, 
       edge: network_components.Edge) -> Set[network_components.Edge]:
-    """Get all of the edge parallel to the given edge.
+    """Get all of the edge parallel to the given `edge`.
```
sleichen: all of the edges

tensornetwork/network.py
```diff
@@ -840,10 +847,11 @@ def split_node(
     """Split a network_components.Node using Singular Value Decomposition.
 
     Let M be the matrix created by flattening left_edges and right_edges into
-    2 axes. Let U S V* = M be the Singular Value Decomposition of M.
-    This will split the network into 2 nodes. The left node's tensor will be
-    U * sqrt(S) and the right node's tensor will be sqrt(S) * (V*) where V* is
-    the adjoint of V.
+    2 axes. Let :math:`U S V^* = M` be the Singular Value Decomposition of 
+    :math:`M`. This will split the network into 2 nodes. The left node's 
+    tensor will be :math:`U * \\sqrt{S}` and the right node's tensor will be
```
sleichen: I don't think we need the * between U and sqrt{S} (and likewise between sqrt{S} and V^*). Similarly to how we don't have them between U, S, and V^* in the specification of the SVD above. Matrix multiplication is understood.

tensornetwork/network.py
```diff
@@ -869,10 +877,14 @@ def split_node(
...
-      right_node: A new node created that connects to all of the `right_edges`.
-      truncated_singular_values: A vector of the dropped smallest singular
-        values.
+      A tuple containing:
+        left_node: 
+          A new node created that connects to all of the `left_edges`.
+        right_node: 
+          A new node created that connects to all of the `right_edges`.
+        truncated_singular_values: 
+          A vector of the dropped smallest singular
```
sleichen: "smallest" not needed here.

tensornetwork/network.py
```diff
@@ -939,13 +953,18 @@ def split_node_full_svd(self,
...
+      singular_values_node: 
+        The new node representing the diagonal matrix of
         singular values.
-      right_node: The new right node created. Its underlying tensor is the same
+      right_node: 
+        The new right node created. Its underlying tensor is the same
         as the V* tensor from the SVD.
-      truncated_singular_values: The vector of truncated singular values.
+      truncated_singular_values: 
+        The vector of truncated singular values.
```
sleichen: The language in the descriptions of the nodes and `truncated_singular_values` should be very similar between `split_node` and `split_node_full_svd`. I'm partial to something like "A new node created that connects to all of the `left_edges`. Its underlying tensor is the same as the U tensor from the SVD." and so on, with appropriate changes for the `split_node` method that mentions sqrt(S).

tensornetwork/network_components.py
```diff
@@ -34,11 +34,11 @@ class Node:
...
   For example:
-  O : No edges means this node represents a scalar value.
-  -O : A single edge means this node is a vector.
-  -O-: Two edges represents a matrix.
-  -O- : Three edges are tensor of rank 3, etc.
-   |
+
+  * No edges means this node represents a scalar value.
+  * A single edge means this node is a vector.
+  * Two edges represents a matrix.
```
sleichen: A node with two edges represents a matrix.

tensornetwork/network_components.py
```diff
@@ -34,11 +34,11 @@ class Node:
...
+  * Three edges are tensor of rank 3, etc.
```
sleichen: A node with three edges is a tensor of rank 3, etc.

(Note: There is a preference now to start using "order" instead of "rank" for the number of legs on a tensor. I don't know how much that sort of change would affect our documentation and code, but we can talk about it.)

tensornetwork/network.py
Thenerdstation: Done.

tensornetwork/network.py
Thenerdstation: Done

tensornetwork/network.py
Thenerdstation: Done

tensornetwork/network.py
Thenerdstation: Done

tensornetwork/network.py
Thenerdstation: Added.

tensornetwork/network_components.py
Thenerdstation: Yes let's discuss this later. I know we've pretty strictly been using "rank" to mean the number of axes in a tensor and "dimension" to mean the size of one of the vector spaces.

tensornetwork/network_components.py
Thenerdstation: Done

tensornetwork/network.py
```diff
@@ -939,13 +954,18 @@ def split_node_full_svd(self,
...
-      right_node: The new right node created. Its underlying tensor is the same
-        as the V* tensor from the SVD.
-      truncated_singular_values: The vector of truncated singular values.
+      A tuple containing:
+        left_node: 
+          A new node created that connects to all of the `left_edges`.
+          Its underlying tensor is :math:`U`
+        singular_values_node: 
+          A new node that has 2 edges connecting `left_node` and `right_node`.
+          It's underlying tensor is :math:`S`
```
sleichen: Its

Issue #140, opened 2019-07-15
SVD of a matrix with numpy backend returns wrong results 


sidek:
If I do SVD of a matrix, M = USV^{dag}, using the TensorNetwork package and numpy backend, I get wrong results. 

The source of this error seems to be that the underlying tensor of V returned by split_node_full_svd depends on the backend chosen.  Specifically, it's the adjoint when using tensorflow but it's just V when using numpy. A code example follows. 

Using numpy, if I enter
```
tensornetwork.set_default_backend("numpy")
net = tensornetwork.TensorNetwork()
tensor = [[5.0,2.0,1.0],[2.0,5.0,1.0],[0.4,0.2,0.3]]
node = net.add_node(tensor)
(u,s,v,terr) = net.split_node_full_svd(node,[node[0]],[node[1]])
print(v.tensor)
print( (u @ s @ v).tensor)
```
I get 
```
[[-0.69395487  0.70553197 -0.1437055 ]
 [-0.6917008  -0.70867212 -0.13904615]
 [-0.19994158  0.00290945  0.9798035 ]]
[[ 2.04827835 -5.06276861  0.41584831]
 [ 4.97540918 -2.05133595  1.01849117]
 [ 0.19076583 -0.4353305   0.25317139]]
```
which is wrong. 

Using tensorflow, if I enter

```
tensornetwork.set_default_backend("tensorflow")
net = tensornetwork.TensorNetwork()
tensor = [[5.0,2.0,1.0],[2.0,5.0,1.0],[0.4,0.2,0.3]]
node = net.add_node(tensor)
(u,s,v,terr) = net.split_node_full_svd(node,[node[0]],[node[1]])
sess= tf.Session()
print(v.tensor.eval(session=sess))
print( (u @ s @ v).tensor.eval(session=sess))
```
I get 
```
[[ 0.6939549   0.6917008   0.19994158]
 [-0.705532    0.7086722  -0.00290945]
 [-0.14370549 -0.13904615  0.97980356]]
[[5.000002   2.0000005  1.0000004 ]
 [2.000001   5.000002   1.0000004 ]
 [0.40000013 0.20000002 0.30000013]]
```
which is correct.

sidek:
I think I've identified the wrong code in the numpy decompositions file and made a pull request, which should solve the issue.

Thenerdstation:
Thanks for catching this!

[...]

PR #141, opened 2019-07-15
Fixed error where v was returned instead of v adjoint by numpy backend

sidek:
The numpy backend was previously mistakenly returning v instead of v adjoint, because np.linalg.svd already returns vh. Hopefully this change solves that problem.


sidek:
I signed it!

Thenerdstation:
https://github.com/google/TensorNetwork/issues/140

PR #147, opened 2019-07-17
MPS classifier

mganahl:
An implementation of DMRG style optimization of a linear MPS classifier


Thenerdstation:
looks like one of your tests is failing because it doesn't have the training data. Can you mock the data in the test instead?

mganahl:
Oh I missed that one! I'll remove it now

mganahl:
oh damn, yes forgot the data ...

mganahl:
should be fixed

PR #149, opened 2019-07-17
Trigger ncon ordering exceptions again.

### LINKED ISSUE #148: ncon throws an obscure error for a valid contraction order
Hey, I just noticed that the following code throws an obscure error:
```python 
d = 2
D1 = 4
D2= 5
a = tf.random_uniform(shape = [D1,D2,d,d])
b = tf.random_uniform(shape = [D1,d,D1])
c = tf.random_uniform(shape = [D2,d,D2])
f = tf.random_uniform(shape = [d,d,d,d])

tn.ncon([a, b, c, f], [[1,3,5,4], [1,2,-1], [3,4,-2], [5,-4,2,-3]])
```
The order above is not chosen smartly, but `ncon` should either throw an interpretable error, or do the contraction

amilsted:
This broke due to the weak references stuff. Fix it and simplify the ncon contraction loop in the process.
Fixes #148.

### DIFF
--- a/tensornetwork/ncon_interface.py
+++ b/tensornetwork/ncon_interface.py
@@ -74,26 +74,31 @@ def ncon(tensors: Sequence[Tensor],
   tn, con_edges, out_edges = ncon_network(
       tensors, network_structure, con_order=con_order, out_order=out_order)
 
+  # Reverse the list so we can pop from the end: O(1).
+  con_edges = con_edges[::-1]
   while con_edges:
-    nodes_to_contract = con_edges.pop(0).get_nodes()
+    e = con_edges.pop()
+    nodes_to_contract = e.get_nodes()
     edges_to_contract = tn.get_shared_edges(*nodes_to_contract)
 
     # Eat up all parallel edges that are adjacent in the ordering.
+    adjacent_parallel_edges = {e}
     while con_edges:
-      if con_edges[0] not in edges_to_contract:
+      if con_edges[-1] in edges_to_contract:
+        adjacent_parallel_edges.add(con_edges.pop())
+      else:
         break
-      con_edges.pop(0)
 
     # In an optimal ordering, all edges connecting a given pair of nodes are
     # adjacent in con_order. If this is not the case, warn the user.
-    leftovers = set(con_edges) & edges_to_contract
+    leftovers = edges_to_contract - adjacent_parallel_edges
     if leftovers:
       warnings.warn(
         "Suboptimal ordering detected. Edges {} are not adjacent in the "
         "contraction order to edges {}, connecting nodes {}. Deviating from "
         "the specified ordering!".format(
           list(map(str, leftovers)),
-          list(map(str, edges_to_contract - leftovers)),
+          list(map(str, adjacent_parallel_edges)),
           list(map(str, nodes_to_contract)))
         )
       con_edges = [e for e in con_edges if e not in edges_to_contract]
--- a/tensornetwork/ncon_interface.py
+++ b/tensornetwork/ncon_interface.py
@@ -16,6 +16,7 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+import warnings
 from typing import Any, Sequence, List, Optional, Union, Text, Tuple, Dict, Any
 from tensornetwork import network
 from tensornetwork import network_components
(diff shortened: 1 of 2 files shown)


tensornetwork/ncon_interface.py
```diff
@@ -76,28 +76,34 @@ def ncon(tensors: Sequence[Tensor],
...
+      # If this already produces the final output, order the edges
+      # here to avoid transposes in some cases.
+      tn.contract_between(
+          *nodes,
+          name="con({},{})".format(*nodes),
+          output_edge_order=out_edges)
+    else:
+      edges_to_contract = tn.get_shared_edges(*nodes)
+      leftovers = set(con_edges) & edges_to_contract
+      if len(leftovers) > 0:
```
Thenerdstation: do `if leftovers` to make pylint happy.

tensornetwork/ncon_interface.py
```diff
@@ -76,28 +76,34 @@ def ncon(tensors: Sequence[Tensor],
...
-      prev_nodes = nodes
+
+    # Eat up all edges sharing these nodes that are adjacent in con_edges.
+    while con_edges:
+      if set(con_edges[0].get_nodes()) != nodes_set:
+        break
+      con_edges.pop(0)
+
+    if nodes_set == tn.nodes_set:
+      assert not con_edges
```
Thenerdstation: raise an error if this is hit instead of doing an assert.

tensornetwork/ncon_interface.py
```diff
@@ -76,28 +76,34 @@ def ncon(tensors: Sequence[Tensor],
...
+      # here to avoid transposes in some cases.
+      tn.contract_between(
+          *nodes,
+          name="con({},{})".format(*nodes),
+          output_edge_order=out_edges)
+    else:
+      edges_to_contract = tn.get_shared_edges(*nodes)
+      leftovers = set(con_edges) & edges_to_contract
+      if len(leftovers) > 0:
+        raise ValueError("Suboptimal ordering: Edges {} are not adjacent "
```
Thenerdstation: I think rather than raising an error, we should just inform the user that we aren't going to follow their ordering exactly.

tensornetwork/ncon_interface.py
Thenerdstation: actually, better yet, just remove this line. This should never happen based on the while loop just before this.

amilsted: Yeah, this should never happen, hence assert rather than raise :). Will remove.

tensornetwork/ncon_interface.py
amilsted: Yeah, I thought about doing that. It's pretty easy to do now that we're looking ahead in con_edges anyway (I was avoiding this in the previous version), so I guess I'll just do it :)

tensornetwork/ncon_interface.py
amilsted: Eugh. Not sure I agree with pylint here, but I submit! ;)

tensornetwork/ncon_interface.py
```diff
@@ -73,31 +74,41 @@ def ncon(tensors: Sequence[Tensor],
...
-            name="con({},{})".format(*nodes),
-            output_edge_order=out_edges)
-      else:
-        tn.contract_between(*nodes, name="con({},{})".format(*nodes))
-      prev_nodes = nodes
+    nodes_to_contract = con_edges.pop(0).get_nodes()
+    edges_to_contract = tn.get_shared_edges(*nodes_to_contract)
+
+    # Eat up all parallel edges that are adjacent in the ordering.
+    while con_edges:
```
Thenerdstation: I don't like having this while loop be nested with the same condition as the outer while loop. This seems like an easy way for things to break.

tensornetwork/ncon_interface.py
```diff
@@ -73,31 +74,41 @@ def ncon(tensors: Sequence[Tensor],
...
+      if con_edges[0] not in edges_to_contract:
+        break
+      con_edges.pop(0)
```
Thenerdstation: This `pop(0)` will cause this code to run in O(n^2) time. Can we modify this to continue to be linear?

amilsted: Sure. How do you feel about just reversing the list beforehand?

amilsted: Alternative: We just increment an integer index.

tensornetwork/ncon_interface.py
amilsted: Hmm. How do you imagine things breaking? The original version had only one loop, but I felt this obscured what was going on (gathering parallel edges before contracting). I found this way more obvious.

Thenerdstation: Can we do this instead?

```python
for edge in reversed(con_edges):
  if edge in edges_to_contract:
    adjacent_parallel_edges.add(edge)
  else:
    break
con_edges = con_edges[:-len(adjacent_parallel_edges)]
```

PR #159, opened 2019-07-20
Tree overhaul, now with tests and switchable backends

amilsted:
A bunch of new functionality. Test coverage could be better, but is way better than nothing!

Currently pylint hates the backend code, which is understandable... it is a little ugly. Perhaps I need to redo it using backend objects, like the TensorNetwork backend code, but that seems like lots of work! ;) Suggestions and patches appreciated.


Thenerdstation:
Ok, for future sake, can we try to make PRs smaller? It's tough to go through 1,000 line PRs.

amilsted:
Sorry! Since this PR adds docstrings to everything, the line count really blows up... The switchable backend also required changes more or less everywhere.

I assumed you had not looked closely at tree tensor network code before (which was already about the same length, just without docstrings!), so I didn't think bunching everything together would matter too much. It's a lot to look at either way, and the documented version is easier to deal with!

experiments/tree_tensor_network/ttn_1d_uniform.py
```diff
@@ -932,11 +1966,180 @@ def get_ham_ising_tube(dtype, Ly, lam=-3.044):
...
+  """Sets the backend to use for tree tensor network computations.
+
+  A backend must be set after importing the module.
+
+  Args:
+    backend: Possible values are "tensorflow", "jax", and "numpy".
+  """
+  tensornetwork.set_default_backend(backend)
+
+  global np
```
Thenerdstation: Wow. This file terrifies me. Why are you globaling everything?

Thenerdstation: This is a really really bad way to swap backends, and it definitely going to break in the future. It's way too easy to miss an assignment that will cause the code to crash, or even worse fail silently. Why didn't you go with the object oriented approach we used eariler?

experiments/tree_tensor_network/ttn_1d_uniform.py
```diff
@@ -903,23 +1785,175 @@ def weylops(q):
...
+
+  Args:
+    dtype: The data type.
+
+  Returns:
+    The Hamiltonian term, separated into a 1-site contribution and a 2-site
+    MPO.
+  """
+  import scipy.io as sio
+  su3_20 = sio.loadmat("experiments/tree_tensor_network/su3_20.mat")
```
Thenerdstation: Don't hard code paths. If eventually want to use this outside of experimental this will break.

experiments/tree_tensor_network/groundstate_example.py
```diff
@@ -16,15 +16,29 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
-import tensorflow as tf
-tf.enable_v2_behavior()
 
 from experiments.tree_tensor_network import ttn_1d_uniform
 
 if __name__ == "__main__":
+  backend = "tensorflow"  # "numpy" and "jax" are also supported!
```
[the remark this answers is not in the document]
amilsted: Well, quite a few of them aren't, hence the need for a backend in the tree code.

experiments/tree_tensor_network/ttn_1d_uniform.py
amilsted: I feel like you are unnecessarily terrified by globals! 😄
Anyway, yes, I warned you the backend stuff was ugly. I'm not sure I see why it's more likely to crash or fail silently (why would it fail silently?) than the OO approach, but the OO approach definitely produces prettier backend code. The main reason I didn't do it was basically an inclination to avoid plastering the rest of the file with "backend" arguments (the tree code is not really OO at all, which makes it easier to use tf.defun/tf.function with it).

experiments/tree_tensor_network/ttn_1d_uniform.py
amilsted: Yeah. Actually, this file is not even included in the PR. Probably I should just take this function out for now...

experiments/tree_tensor_network/ttn_1d_uniform.py
amilsted: Would you be happy with a global backend object?

Thenerdstation: You could do what we do now with the `set_default_backend` and make a module object.

And the reason I'm worried is that we both have a `switch_backend` method and a new guy who is trying to bring in support for Pytorch right now. Like I said, it's too easy to miss something with the current implementation (vs always throwing a `NotImplementedError` in the `BaseBackend` object).

amilsted:
> You could do what we do now with the set_default_backend and make a
module object.

Ok, good. This is what I meant by a "global backend object".

> And the reason I'm worried is that we both have a switch_backend method
and a new guy who is trying to bring in support for Pytorch right now.

We will still both have a `switch_backend` unless we want to add support
for everything the tree code uses to the core backends. If we want to make
it easy for code the uses TensorNetwork to be backend agnostic, we might
want to do that (and more), but it's obviously more code to maintain in the
core. What are your thoughts on this? Should the TensorNetwork core backend
be extended to something that can be used generally by projects that use
TensorNetwork?

> Like I said, it's too easy to miss something with the current
implementation (vs always throwing a NotImplementedError in the BaseBackend
object).

Yes, I agree that throwing a `NotImplementedError` is more informative than
what would happen now, which would be something like "function <blah> is
not defined in module <bleh>".

I will switch to an OO backend for the tree code.

Thenerdstation:
Thank you!

experiments/tree_tensor_network/ttn_1d_uniform.py
amilsted: I realized a reason you might be more concerned... a missing definition might mean one is left with a function from the wrong backend after switching more than once. That *would* be nasty...

experiments/tree_tensor_network/ttn_1d_uniform.py
```diff
@@ -11,67 +11,121 @@
...
+from __future__ import print_function
+import sys
+import copy
+import math
+import time
+import contextlib
+import tensornetwork
+
+
+def set_backend(backend_name):
```
Thenerdstation: This is much better. Thank you!

experiments/tree_tensor_network/ttn_1d_uniform.py
```diff
@@ -11,67 +11,121 @@
...
+  """Sets the backend to use for tree tensor network computations.
+
+  A backend must be set after importing the module.
+
+  Args:
+    backend: Possible values are "tensorflow", "jax", and "numpy".
+  """
+  global backend
```
Thenerdstation: You can avoid this last global by making it a module object. 

See https://github.com/google/TensorNetwork/blob/master/tensornetwork/config.py
and https://github.com/google/TensorNetwork/blob/b6ac4a08b52b6c78d1169231d6c0ec06517dd25d/tensornetwork/__init__.py#L10 for an example on how to do it.

experiments/tree_tensor_network/ttn_1d_uniform.py
```diff
@@ -879,64 +1739,562 @@ def descend_full_state_pure(isos_012):
...
+  Zcol = [backend.convert_to_tensor(Zc, dtype=dtype) for Zc in Zcol]
 
   h1 = lam * sum(Zcol) - sum(Xcol[i] @ Xcol[(i + 1) % Ly] for i in range(Ly))
 
   h_mpo_2site = ([-Xc for Xc in Xcol], Xcol)
 
   return h1, h_mpo_2site
+
+
+class TTNBackend():
```
Thenerdstation: Would it make sense to add these operations to the BaseBacked code in tensornetwork proper? (To be done in a later PR.)

experiments/tree_tensor_network/ttn_1d_uniform.py
Thenerdstation: Once this is fixed we can merge this code.

experiments/tree_tensor_network/ttn_1d_uniform.py
amilsted: Yeah, I wondered about that in a comment above. Do we want to support a more comprehensive API for projects that use TensorNetwork? This would be very useful, but could be tricky due to differing behaviors of different frameworks. I have done the minimal necessary implementation for the tree code here, but it would need upgrading to make it more general.

experiments/tree_tensor_network/ttn_1d_uniform.py
amilsted: I don't really understand the advantage of this; an extra file to avoid the use of a single "global" statement? I think the "global" statement is easier to understand. Also, this would seem to require prepending all `backend.something()` calls with a module name e.g. `config.backend.something()`. Did I miss something?

amilsted: Ok, I figured out a way of avoiding that issue. Still not convinced of the evil of "global" here :)

Thenerdstation:
Thanks for the fixes!

PR #161, opened 2019-07-21
Feature/backend tests

MichaelMarien:
Added basic tests for backends to ensure consistency of methods, especially more functionality or backends are added.

Checks the input to np.diag in the numpy backend as its behaviour depends on input size.

### DIFF
--- a/tensornetwork/backends/numpy/numpy_backend_test.py
+++ b/tensornetwork/backends/numpy/numpy_backend_test.py
@@ -84,7 +84,7 @@ def test_convert_to_tensor():
   array = np.ones((2, 3, 4))
   a = net.add_node(array)
   actual = net.backend.convert_to_tensor(array)
-  expected =np.ones((2, 3, 4))
+  expected = np.ones((2, 3, 4))
   assert isinstance(actual, type(a.tensor))
   np.testing.assert_allclose(expected, actual)
 
--- a/tensornetwork/backends/numpy/numpy_backend_test.py
+++ b/tensornetwork/backends/numpy/numpy_backend_test.py
@@ -41,7 +41,7 @@ def test_concat():
 def test_shape():
   net = tensornetwork.TensorNetwork('numpy')
   a = net.add_node(np.ones([2, 3, 4]))
-  assert type(net.backend.shape(a.tensor)) == tuple
+  assert isinstance(net.backend.shape(a.tensor), tuple)
   actual = net.backend.shape(a.tensor)
   expected = np.array([2, 3, 4])
   np.testing.assert_allclose(expected, actual)
@@ -85,7 +85,7 @@ def test_convert_to_tensor():
   a = net.add_node(array)
   actual = net.backend.convert_to_tensor(array)
   expected =np.ones((2, 3, 4))
-  assert type(actual) == type(a.tensor)
+  assert isinstance(actual, type(a.tensor))
   np.testing.assert_allclose(expected, actual)
 
 def test_trace():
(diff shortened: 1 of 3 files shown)


tensornetwork/backends/numpy/numpy_backend_test.py
```diff
@@ -0,0 +1,106 @@
...
+  @classmethod
+  def setUpClass(cls):
+    set_default_backend('numpy')
+
+  def test_tensordot(self):
+    net = tensornetwork.TensorNetwork()
+    a = net.add_node(2*np.ones((2, 3, 4)))
+    b = net.add_node(np.ones((2, 3, 4)))
+    tensordotted_tensor = net.backend.tensordot(a.tensor, b.tensor, ((1, 2), (1, 2)))
+    self.assertListEqual(np.array(tensordotted_tensor).tolist(), [[24.0, 24.0], [24.0, 24.0]])
```
Thenerdstation: We prefer to use `pytest` for our testing. Can you remove your dependency on unittest and use pytest instead?

MichaelMarien: Sure, was not using correct line-length anyway, so need to revisit all the code : )

Thenerdstation: Yes we should. That was likely left over from the initial open source release

tensornetwork/backends/numpy/numpy_backend_test.py
```diff
@@ -0,0 +1,113 @@
...
+  actual = net.backend.diag(b.tensor)
+  expected = np.array([[1, 0, 0], [0, 2, 0], [0, 0, 3]])
+  np.testing.assert_allclose(expected, actual)
+
+def test_convert_to_tensor():
+  net = tensornetwork.TensorNetwork('numpy')
+  array = np.ones((2, 3, 4))
+  a = net.add_node(array)
+  actual = net.backend.convert_to_tensor(array)
+  expected =np.ones((2, 3, 4))
```
Thenerdstation: You missed a space

tensornetwork/backends/numpy/numpy_backend_test.py
```diff
@@ -0,0 +1,113 @@
...
+  assert isinstance(actual, type(a.tensor))
+  np.testing.assert_allclose(expected, actual)
+
+def test_trace():
+  net = tensornetwork.TensorNetwork('numpy')
+  a = net.add_node(np.array([[1, 2, 3], [4, 5, 6]]))
+  actual = net.backend.trace(a.tensor)
+  expected = 6
+  np.testing.assert_allclose(expected, actual)
```
Thenerdstation: You can just do `np.testing.assert_allclose(actual, 6)`

tensornetwork/backends/numpy/numpy_backend_test.py
```diff
@@ -0,0 +1,113 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import numpy as np
+import tensornetwork
+import pytest
+
+
+def test_tensordot():
+  net = tensornetwork.TensorNetwork('numpy')
```
Thenerdstation: Instead of making a `TensorNetwork`, you can just construct the backend object directly and use that. That is preferred since it decouples the backend tests from the implementation of `TensorNetwork`.

You can do this by doing

```
from tensornetwork.backends.numpy import numpy_backend
backend = numpy_backend.NumPyBackend()
```

MichaelMarien: Totally right

Issue #53, opened 2019-05-16
Add PyTorch backend.


Thenerdstation:
Once https://github.com/google/TensorNetwork/pull/52 is submitted, it would be nice to add a pytorch backend. 

TF eager and pytorch are very similar, so someone could use the `TensorFlowBackend` as an example.

MichaelMarien:
Hi,

would be happy to take it up, as you said, should be quite similar to the tf-backend. Quite a busy July though, so might only finish it in Aug. Is this very urgent?

Thenerdstation:
Not urgent at all! And we would love to have you as a contributor.

It should be fairly straightforward with the exception of some weird shape only functions, specifically `prod` and `concat`. I'll be sure document it better before you get started in August.

If you get stuck or need any help, feel free to email me! chaseriley@google.com

MichaelMarien:
Just remembered PyTorch still does not have full support for complex numbers. Do we expect this to be blocking in the future?

Thenerdstation:
There are still use cases for exclusively real valued tensors, especially in ML. I'd say it's still worth it.

Thenerdstation:
We'd have to explicitly exclude some tests, but I don't think we have that many.

MichaelMarien:
I added a first try / progress in PR #165 , if you have time, feel free to give comments.

Thenerdstation:
We'll push 0.0.6 tonight with your Pytorch backend!

Thenerdstation:
0.0.6 is live! Just do `pip upgrade tensornetwork` and you should be good to go

PR #165, opened 2019-07-28
Feature/pytorch backend

MichaelMarien:
First attempt to add PyTorch backend, expect a lot of comments & input on this PR and there is no hurry as I might not have time to address those quickly. Consider this a first check of progress.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -160,7 +160,7 @@ additional-builtins=
 ignore-mixin-members=yes
 # List of classes names for which member attributes should not be checked
 # (useful for classes with attributes dynamically set).
-ignored-classes=SQLObject,twisted.internet.reactor,hashlib,google.appengine.api.memcache
+ignored-classes=SQLObject,twisted.internet.reactor,hashlib,google.appengine.api.memcache,torch
 # List of members which are set dynamically and missed by pylint inference
 # system, and so shouldn't trigger E0201 when accessed. Python regular
 # expressions are accepted.
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 numpy>=1.16
 tensorflow>=1.13
-torch>=1.1
 graphviz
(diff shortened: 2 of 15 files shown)


tensornetwork/tensornetwork_test.py
```diff
@@ -1165,6 +1165,8 @@ def test_switch_backend():
 
 
 def test_svd_consistency(backend):
+  if backend == "pytorch":
+    pytest.skip()
```
Thenerdstation: Be sure to add "PyTorch doesn't support complex numbers" as the skip reason.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 numpy>=1.16
 tensorflow>=1.13
+torch>=1.1
```
Thenerdstation: We prefer to not force the user to install Pytorch if they don't need it. Can you put this line in the `requirements_travis.txt` file instead?

pylintrc
```diff
@@ -160,7 +160,7 @@ additional-builtins=
 ignore-mixin-members=yes
 # List of classes names for which member attributes should not be checked
 # (useful for classes with attributes dynamically set).
-ignored-classes=SQLObject,twisted.internet.reactor,hashlib,google.appengine.api.memcache
+ignored-classes=SQLObject,twisted.internet.reactor,hashlib,google.appengine.api.memcache,torch
```
Thenerdstation: Any reason you added this?

tensornetwork/tensornetwork_test.py
MichaelMarien: Done!

requirements.txt
MichaelMarien: makes a lot of sense

pylintrc
MichaelMarien: It is to fix this issue: https://github.com/pytorch/pytorch/issues/701
I tried adding the separate functions that are used only, but for some reason that didn't work, so I reverted to the proposed solution of adding the full toch library. Tbh, I'm not 100% happy with this either.

Thenerdstation:
https://github.com/google/TensorNetwork/issues/53

Issue #171, opened 2019-08-05
Better Contraction Algorithms


Thenerdstation:
This has already been brought up in https://github.com/google/TensorNetwork/issues/120, but I'm bringing it up again because our contraction algorithms suck.

We need much better algorithms, as hand coding contraction orders, or worse, just using naive is a really bad practice.

As the first step, we need to implement a fast version of fully optimal contraction. It's ok if it's just in python for now, but we need it to exist. We could also possibly adapt opteinsum's code to our own, but I don't know how easily that will be.

stavros11:
I think it is fairly easy to adapt opt_einsum's code, at least for their optimal DFS contraction. There is a function `optimal` that takes sets with input/output edges and a dictionary with the dimensions and returns a list with the optimal contraction order. We can just write a wrapper that creates the sets and dictionary and use this function as it is. I have only checked how the optimal works but it might be possible to "transfer" their other opt_einsum algorithms similarly (if the opt_einsum people are okay with using their code of course).

At least this is the quick solution. If we would like something more elegant perhaps we should implement the algorithm from scratch using our cost calculators, because opt_einsum has its own way of calculating costs.

Thenerdstation:
>  if the opt_einsum people are okay with using their code of course.

It's MIT open sourced, so they literally have no say in the matter. :)

Making a shim to go from our `TensorNetwork` to their `optimal` might be a good first step. Would love a PR on this! Feel free to include `opt_einsum` as a dependency if need be.

Thenerdstation:
Actually, using `opt_einsum` library seems like exactly what we should be doing. We could use the signatures/ids of the edges instead of letters. This allows us to make a shim pretty easily. 

Let me try writing a PR tonight. If I can't get it done by midnight then feel free to take charge on it.

Thenerdstation:
Stavors, if you want to add the other contraction algorithms from `opt_einsum` after this PR is in, go for it.

stavros11:
Sure, I can look at this. It should be easy now that the first one is in.

stavros11:
By the way, it might be useful to take advantage of `opt_einsum`'s [path reuse functionality](https://optimized-einsum.readthedocs.io/en/latest/reusing_paths.html) so that we avoid recalculating paths when we want to contract the same network many times (for example during optimization). I am not sure how this could be done easily, for now I just mention it to keep it in mind.

Thenerdstation:
Good idea. Add it as a Github issue so we don't forget about it.

PR #173, opened 2019-08-07
Add opt_einsum contractors

stavros11:
Adding the rest contraction methods from `opt_einsum`. Note that this works only for connected network. I also think there was a bug with trace edges in `optimal` (you can check that with a test before merging this) which is fixed here by contracting trace edges first always.

I will try to add the `random-greedy` in some time because currently I have trouble with importing `opt_einsum.random_path`.


Thenerdstation:
This is awesome! Now that was have these auto contractors, I think we can officially bump to a 0.1.0 release ^-^

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -139,12 +139,44 @@ def auto(net: network.TensorNetwork,
...
     memory_limit: Maximum number of elements in an array during contractions.
 
   Returns:
     The network after full contraction.
   """
   alg = functools.partial(optimizer, memory_limit=memory_limit)
+  return base(net, alg)
+
+
+def random_greedy(net: network.TensorNetwork,
```
Thenerdstation: Be sure to include this in the contractors __init__.py

stavros11:
It would be great if someone could test if the `random-greedy` contractor can be accessed from the `pip` version of `opt_einsum`. Locally I could only fix this by cloning the `opt_einsum` repo, but it seems that this won't work for the tests here.

Thenerdstation:
> It would be great if someone could test if the `random-greedy` contractor can be accessed from the `pip` version of `opt_einsum`. Locally I could only fix this by cloning the `opt_einsum` repo, but it seems that this won't work for the tests here.

Yeah, it's missing from my pip install as well. I've raised an issue on the opt_einsum github.  https://github.com/dgasmith/opt_einsum/issues/93

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -19,6 +19,7 @@ def base(net: network.TensorNetwork,
   Returns:
     The network after full contraction.
   """
+  net.check_connected()
```
Thenerdstation: Add a disconnected network test.

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -0,0 +1,158 @@
...
+  if n < 7:
+    return branch(net, memory_limit)
+  if n < 9:
+    return branch(net, memory_limit, nbranch=2)
+  if n < 15:
+    return branch(net, nbranch=1)
+  return greedy(net, memory_limit)
+
+
+def custom(net: network.TensorNetwork, optimizer,
```
Thenerdstation: add typing to optimizer. 

(Last comment. I promise :))

stavros11: No worries! :D I just hope tests pass because I had it and removed it for this reason (I think I had typo)

Thenerdstation:
Strange, I don't understand why so much is missing from the pip opt_einsum install.

Well whatever, just make it an `Any`.

Thenerdstation:
fixes https://github.com/google/TensorNetwork/issues/171

Issue #180, opened 2019-08-09
Flattening and then removing a node causes a crash.


Thenerdstation:
When flattening edges and them removing a node with flattened edges, we hit a crash. 

This is because we are setting `node.axis_names` to `None` after flattening.

PR #181, opened 2019-08-09
Fixed remove/flatten bug

Thenerdstation:
We should figure out a clean way of dealing with flattened edge names.


Thenerdstation:
https://github.com/google/TensorNetwork/issues/180

PR #185, opened 2019-08-11
A very simple MERA.

amilsted:
A MERA in ~100 lines (excluding docstrings), using autodiff to derive all needed functionality from a single tensor network. Also happens to use JAX.

### DIFF
--- a/examples/simple_mera/simple_mera.py
+++ b/examples/simple_mera/simple_mera.py
@@ -14,7 +14,12 @@
 """Very simple scale-invariant MERA.
 
 Uses automatic differentiation to construct ascending and descending
-superoperators, as well as environment tensors.
+superoperators, as well as environment tensors so that only a single tensor
+network needs to be defined: The network for computing the energy using a
+single layer of MERA together with a reduced state and a hamiltonian.
+
+Run as the main module, this module executes a script that optimizes a MERA
+for the critical Ising model.
 """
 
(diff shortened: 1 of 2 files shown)


examples/simple_mera/simple_mera.py
```diff
@@ -0,0 +1,318 @@
...
+from __future__ import print_function
+import jax
+import jax.config
+jax.config.update("jax_enable_x64", True)
+import jax.numpy as np
+import tensornetwork
+from tensornetwork import contractors
+
+
+@jax.jit
```
Thenerdstation: Ooo! You're using Jax now. ^-^

examples/simple_mera/simple_mera.py
```diff
@@ -0,0 +1,318 @@
...
+
+    net.connect(iso_l[2], rho[0])
+    net.connect(iso_c[2], rho[1])
+    net.connect(iso_r[2], rho[2])
+
+    net.connect(iso_l[0], iso_l_con[0])
+    net.connect(iso_l[1], un_l[2])
+    net.connect(iso_c[0], un_l[3])
+    net.connect(iso_c[1], un_r[2])
+    net.connect(iso_r[0], un_r[3])
```
Thenerdstation: `edge1 ^ edge2` is supported now. Feel free to use it. (We're still working on getting pylint to not complain about it though...)

Issue #158, opened 2019-07-19
Splice notation for edges? 


sidek:
We can access edges of a node by list notation, 
```
edge = node[0]
```
so it would be very nice (and useful, to me at least) if one could also access several edges of a node via splice notation 
```
edges = node[:4]
```
If this seems like a good idea, I'm happy to try implementing it.

Thenerdstation:
You could do this right now with `node.edges[:4]`, but I agree having it built into the indexing would be move convenient. 

And go for it! We love getting PRs. :)

PR #189, opened 2019-08-13
Adding slice notation for edges 

sidek:
This adds the functionality of accessing edges like node[1:2], as discussed in issue #158.

Only integer slices work -- I thought about making node['a':'b'] work, for named axes, but this seems like functionality more likely to create bugs than to be useful. I can update the PR to include this functionality if anyone thinks otherwise, however.

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -260,8 +260,13 @@ class Node:
   @overload
   def __getitem__(self, key: slice) -> List["Edge"]:
     return self.edges[key]
-
+  
+  @overload
   def __getitem__(self, key: Union[int, Text]) -> "Edge":
+    return self.get_edge(key)
+
+  def __getitem__(self, key: Union[int, Text, slice]
+                 ) -> Union["Edge", List["Edge"]]:
     if isinstance(key, slice):
       return self.edges[key]
     return self.get_edge(key)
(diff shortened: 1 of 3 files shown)


sidek:
Oops! Will fix typing errors and report back. 

For type consistency elsewhere, it probably it isn't good to have get_edge possibly return a list, so this calls for some refactoring.

tensornetwork/network_components.py
```diff
@@ -256,9 +257,20 @@ def has_nondangling_edge(self):
...
+  def __getitem__(self, key: slice) -> List["Edge"]:
+    return self.edges[key]
+  
+  @overload
   def __getitem__(self, key: Union[int, Text]) -> "Edge":
     return self.get_edge(key)
 
+  def __getitem__(self, key: Union[int, Text, slice]
+                 ) -> Union["Edge", List["Edge"]]:
+    if isinstance(key, slice):
```
Thenerdstation: Isn't this redundant with the `@overload`?

sidek: `@overload`-decorated function signatures are only read by the type checker, and must be followed by a non-overload decorated function signature, at least according to the typing documentation.

Really I wanted to write 
```
  @overload
  def __getitem__(self, key: slice) -> List["Edge"]:
    pass 

  @overload
  def __getitem__(self, key: Union[int, Text]) -> "Edge":
    pass
```
but pytype didn't like this, initially. 

It seems like pytype is fine with the above now that I've demanded travis use a more updated version of pytype, so I've merged that.

Thenerdstation:
Not sure what's going on with the Google bot...

sidek:
I signed it!

PR #197, opened 2019-08-20
Base node

MichaelMarien:


tensornetwork/network_components.py
```diff
@@ -46,36 +46,36 @@ class Node:
   an arbitrary dimension.
   """
 
-  def __init__(self, tensor: Tensor, name: Text, axis_names: List[Text],
-               network: TensorNetwork) -> None:
-    """Create a node for the TensorNetwork.
+  def __init__(self, name: Text, axis_names: List[Text], network: TensorNetwork,
+               shape: Optional[Tuple[int]] = None) -> None:
```
Thenerdstation: I'm not sure `Optional` makes sense here. Is there ever a case where the shape of a node should be unknown?

Thenerdstation: Actually nevermind. I now see that `node.shape` is a property.

tensornetwork/network_components.py
```diff
@@ -118,49 +118,63 @@ def add_edge(self,
...
+  @abstractmethod
   def set_tensor(self, tensor):
-    self.tensor = tensor
+    return
 
   @property
+  @abstractmethod
   def shape(self):
-    return self.network.backend.shape_tuple(self._tensor)
+    return self._shape
```
Thenerdstation: Raise an error if `_shape` is None.

tensornetwork/network_components_test.py
```diff
@@ -35,6 +37,12 @@ def fixture_double_node_edge(backend):
   return DoubleNodeEdgeTensor(node1, node2, edge1, edge12, tensor)
 
 
+@pytest.fixture(name='copy_node')
+def fixture_copy_node(backend):
```
Thenerdstation: I did not realize you could nest fixtures like this. Cool!

tensornetwork/network_components.py
MichaelMarien: Done!

tensornetwork/network_components.py
MichaelMarien: Yeah, I was doubting myself for a while, but with raising an error, I think this is ok. Also note this is an abstract class and cannot be directly instantiated anyhow.

PR #213, opened 2019-08-27
backend methods

mganahl:


Thenerdstation:
What might be a better idea is to just have the `TensorNetwork` store it's own dtype. This could be an 
optional argument to the `TensorNetwork` creation that would automatically be deduced once the first node is added.

I've actually been meaning to do this, but I think we should do this now.

mganahl:
sounds good.  I assume you are thinking of soemthing like this:
```python
def __init__(..., dtype=None):
  ...
  self.dtype = dtype
def add_node(...):
  tensor = self.backend.convert_to_tensor(tensor)
  if dtype is None:
    self.dtype = tensor.dtype
```

mganahl:
Or rather, we just convert each following node to the same dtype as the first one.

Thenerdstation:
Yep! Basically what you describe, but rather than converting I'd rather throw an error. (Loud, slightly tedious error messages are much more preferable to silent bugs).

mganahl:
OK, I can add this to this  PR (since I need it there anyway)

mganahl:
I think I'll make `dtype` a required argument to the backend methods. 
If  we add `self.dtype` to TensorNetwork, I'll need to add it to add `dtype` to the backend as 
well if I want the above code to just work, but I can't provide any `dtype` to the backend at initialization time.

mganahl:
actually strike  that, should be fine

mganahl:
@Thenerdstation do you know which dtypes jax actually supports? If I do
`jax.random.uniform((2,2), dtype=np.float64)` I get a `float32` array back.

Thenerdstation:
JAX is weird in that the dtypes are determined by an environment variable (See top of `tensornetwork_test.py`

mganahl:
Hey @Thenerdstation could you have a look at this? I have added a `dtype` attribute to all backends, and added a `dtype` property to `TensorNetwork`. The way its done now is to have 
`backend.dtype` initialized to `None` at the initialization of `TensorNetwork` (that's hard coded in the `config.py` file for now), and set it to `backend.convert_to_tensor(tensor).dtype` at the first call to `TensorNetwork.add_node(tensor)`.

mganahl:
Actually, wait for a bit, just noticed one more problem

mganahl:
@Thenerdstation this is OK from my side

tensornetwork/backends/numpy/numpy_backend.py
```diff
@@ -17,25 +17,22 @@
...
-    except ImportError:
-      raise AssertionError("numpy is not installed.")
     from tensornetwork.backends.numpy import decompositions
     self.np = numpy
     self.decompositions = decompositions
     self.name = "numpy"
+    self.dtype = dtype
 
-  def tensordot(self, a: Tensor, b: Tensor, axes: Sequence[Sequence[int]]):
+  def tensordot(self, a: Tensor, b: Tensor, axes: Sequence[Tuple[int, ...]]):
```
Thenerdstation: Why was this changed to a tuple?

tensornetwork/backends/numpy/numpy_backend_test.py
```diff
@@ -75,10 +75,10 @@ def test_sqrt():
 
 def test_diag():
   backend = numpy_backend.NumPyBackend()
-  a = backend.convert_to_tensor(np.array([[1, 2, 3], [4, 5, 6]]))
+  a = backend.convert_to_tensor(np.array([[1.0, 2, 3], [4, 5, 6]]))
```
Thenerdstation: Please make all of these values floats instead of just one.

tensornetwork/backends/numpy/numpy_backend_test.py
```diff
@@ -123,3 +123,106 @@ def test_convert_bad_test():
...
+def test_eye_dtype():
+  backend = numpy_backend.NumPyBackend(dtype=np.float64)
+  dtype = np.float32
+  a = backend.eye(N=4, M=4, dtype=dtype)
+  assert a.dtype == dtype
+
+
+def test_ones_dtype():
+  backend = numpy_backend.NumPyBackend(dtype=np.float64)
+  dtype = np.float32
```
Thenerdstation: Please parameterize the dtype so we make sure it works for all of the supported values.

tensornetwork/backends/pytorch/pytorch_backend.py
```diff
@@ -85,7 +86,12 @@ def diag(self, tensor: Tensor) -> Tensor:
     return self.torch.diag(tensor)
 
   def convert_to_tensor(self, tensor: Tensor) -> Tensor:
-    return self.torch.as_tensor(tensor)
+    result = self.torch.as_tensor(tensor)
+    if self.dtype is not None and result.dtype is not self.dtype:
```
Thenerdstation: Add this check before doing the conversion.

tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -18,7 +18,7 @@
 from typing import Optional, Any, Sequence, Tuple
 from tensornetwork.backends import base_backend
 from tensornetwork.backends.tensorflow import decompositions
-
+import tensorflow as tf
```
Thenerdstation: Do not include this at the top. We don't want to assume everyone has tensorflow installed when we switch to having JAX be the default backend.

tensornetwork/backends/tensorflow/tensorflow_backend_test.py
```diff
@@ -113,3 +115,106 @@ def test_einsum():
...
+
+
+def test_zeros():
+  backend = tensorflow_backend.TensorFlowBackend()
+  a = backend.zeros((4, 4))
+  np.testing.assert_allclose(np.zeros((4, 4)), a)
+
+
+def test_randn():
+  backend = tensorflow_backend.TensorFlowBackend()
```
Thenerdstation: A lot of this code is very redudent. Can you please parameterize these tests to reduce code copying.

tensornetwork/tensornetwork_test.py
```diff
@@ -1366,7 +1367,16 @@ def test_switch_backend():
 def test_svd_consistency(backend):
   if backend == "pytorch":
     pytest.skip("Complex numbers currently not supported in PyTorch")
-  net = tensornetwork.TensorNetwork(backend=backend)
+  if backend == 'numpy':
```
Thenerdstation: This shouldn't be needed if the `TensorNetwork` automatically sets its backend dependent upon the first node added.

tensornetwork/tensornetwork_test.py
```diff
@@ -1618,3 +1631,76 @@ def test_network_copy_identities(backend):
...
+
+
+def test_network_numpy_backend_dtype_2():
+  dtype = tf.float32
+  with pytest.raises(TypeError):
+    tensornetwork.TensorNetwork(backend='numpy', dtype=dtype)
+
+
+def test_network_numpy_backend_dtype_3():
+  dtype = np.float32
```
Thenerdstation: Please parameterize these tests.

tensornetwork/network.py
```diff
@@ -1125,8 +1170,8 @@ def split_node_full_svd(
       right_name: Optional[Text] = None,
       left_edge_name: Optional[Text] = None,
       right_edge_name: Optional[Text] = None,
-  ) -> Tuple[network_components.BaseNode, network_components.
-             BaseNode, network_components.BaseNode, Tensor]:
+  ) -> Tuple[network_components.BaseNode, network_components
```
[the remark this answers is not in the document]
mganahl: that's my autoformatter

tensornetwork/backends/pytorch/pytorch_backend.py
mganahl: Is there any advantage? The reason it's like that is that `tensor` can be a `np.ndarray` (usually is for me), but `self.dtype` is a `torch.dtype`. For example, `torch.float64`  compares `False` to `np.float64`, even though those `dtypes` are compatible

Thenerdstation: Aw I see your point now. Ok, lets keep it as is then.

mganahl:
Hey @Thenerdstation I made most changes (except the formatting). Can we postpone the formatting issue for now? I'll fix the autoformatter and make the changes with the other PR #218

tensornetwork/backends/pytorch/pytorch_backend_test.py
```diff
@@ -74,9 +75,9 @@ def test_sqrt():
 
 def test_diag():
   backend = pytorch_backend.PyTorchBackend()
-  b = backend.convert_to_tensor(np.array([1, 2, 3]))
+  b = backend.convert_to_tensor(np.array([1.0, 2, 3]))
```
Thenerdstation: Please make all of these values floats

tensornetwork/backends/pytorch/pytorch_backend_test.py
```diff
@@ -110,3 +111,105 @@ def test_outer_product():
   expected = np.array([[[[[2.0, 2.0], [2.0, 2.0]]]], [[[[2.0, 2.0], [2.0,
```
Thenerdstation: Can you do `np.array([2.0] * 8).reshape((1, 1, 2, 2, 2))`  instead? Or whatever shape it is. I can't tell right now which is why you should change it. It also helps avoid the ugly `[[[[`

tensornetwork/backends/tensorflow/tensorflow_backend.py
Thenerdstation: Please still fix this

tensornetwork/backends/tensorflow/tensorflow_backend_test.py
```diff
@@ -91,7 +93,7 @@ def test_convert_to_tensor():
 
 def test_trace():
   backend = tensorflow_backend.TensorFlowBackend()
-  a = backend.convert_to_tensor(np.array([[1, 2, 3], [4, 5, 6]]))
+  a = backend.convert_to_tensor(np.array([[1, 2, 3], [4, 5, 6.0]]))
```
Thenerdstation: Make all values floats

tensornetwork/backends/numpy/numpy_backend_test.py
mganahl: Done

tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -88,7 +80,12 @@ def diag(self, tensor: Tensor) -> Tensor:
     return self.tf.linalg.diag(tensor)
 
   def convert_to_tensor(self, tensor: Tensor) -> Tensor:
-    return self.tf.convert_to_tensor(tensor)
+    result = self.tf.convert_to_tensor(tensor)
+    if self.dtype is not None and result.dtype is not self.dtype:
```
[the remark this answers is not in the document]
mganahl: Same problem as for pytorch backend. `np.ndarray.dtype` is of type `np.dtype`, not `type` (as is e.g `np.float64`). `==` comparison  of `dtype` with e.g. `tf.float64` raises `TypeError`, and  `is` comparison of course gives `False`

tensornetwork/backends/tensorflow/tensorflow_backend_test.py
```diff
@@ -74,9 +76,9 @@ def test_sqrt():
 
 def test_diag():
   backend = tensorflow_backend.TensorFlowBackend()
-  b = backend.convert_to_tensor(np.array([1, 2, 3]))
+  b = backend.convert_to_tensor(np.array([1, 2, 3.0]))
   actual = backend.diag(b)
-  expected = np.array([[1, 0, 0], [0, 2, 0], [0, 0, 3]])
+  expected = np.array([[1, 0, 0], [0, 2, 0], [0, 0, 3.0]])
```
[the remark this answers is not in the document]
mganahl: Done

tensornetwork/config.py
```diff
@@ -11,5 +11,48 @@
...
+import torch
+default_backend = "tensorflow"
+#for backwards compatibility default dtypes have to be `None`
+#changing this will cause tests to fail due backend.convert_to_tensor(tensor)
+#raising TypeErrors when incoming `tensor` has a dtype different from
+#backend.dtype.
 
-default_backend = "tensorflow"
\ No newline at end of file
+default_dtypes = {
```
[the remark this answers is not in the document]
mganahl: added it

tensornetwork/backends/tensorflow/tensorflow_backend_test.py
mganahl: fixed

tensornetwork/backends/pytorch/pytorch_backend_test.py
mganahl: Done

tensornetwork/tensornetwork_test.py
mganahl: Done

Issue #192, opened 2019-08-16
CopyNode is redundantly instantiated


Bonnevie:
Considering that `CopyNode` is extremely sparse, it seems a waste to instantiate it physically. In fact, this just became an issue for me since it turns out that Numpy does not allow tensors with more than 32 dimensions (probably reasonable), which prevents a `CopyNode` of that order (unproblematic) from being constructed. Use case-wise this crops up when using tensor networks for probabilistic models.

Is the tensor just being built to satisfy the `Node` superclass? Or is there a deeper reason?

Controversially, the core `Node` class should maybe not contain tensors by default to facilitate the construction of structured nodes (e.g. subnetworks, symmetric nodes, etc). This seems to be touched upon in #62.

Thenerdstation:
We can (and should) actually do this pretty easily now. 

We set `node.tensor` to be a `@property` already, so all we would need to do is override it in the `CopyNode` class to be a lazy creation, and if someone uses the `contract_copy_node` method, then we never create the tensor at all.

On a higher level though, sparse and semetric Tensors are really poorly supported at the moment. We had a guy who was working on it (author on the other issue) but he is out on paternity leave right now.

Thenerdstation:
Haha, we have a TODO to do exactly what you want.
https://github.com/google/TensorNetwork/blob/f86d8357b6365f31a0b8dc80967aff05cae5bae8/tensornetwork/network_components.py#L299

I likely won't have time to get to this until late August/September, but I agree this should get fixed.

MichaelMarien:
@Thenerdstation I have some time next week, if you want, you can assign the lazy-evaluation of copy-node to me.

mganahl:
@Thenerdstation, just curious, what's the plan for supporting (block) sparsity?

Bonnevie:
Making the tensor lazy makes great sense, but I am a bit unsure how to handle the call to `super()` in the `CopyNode` constructor? Since the standard `Node` class requires a `tensor` in its constructor, we'll either need to feed it a dummy or make it an optional parameter (or make the tensor node a subclass of some `AbstractNode` class). The dummy option seems less than ideal IMO.

Thenerdstation:
> what's the plan for supporting (block) sparsity?

We welcome PRs. 

I'm a little busy with a research project right now, so don't expect any big updates from me until September

Thenerdstation:
@MichaelMarien Sure! Go ahead and try to take this up.

There are 2 ways of going about it. 

[...]

Issue #86, opened 2019-05-31
Tensor network with symmetries


mganahl:
Some tensor networks can have global symmetries like conserved U(1) of SU(2) charge. 
The symmetry imposes a sparse block structure on the individual tensors, which can be exploited to speed up contractions and decompositions, and reduce memory requirements. 

One way of implementing symmetries is to wrap the blocks together with some additional information into a new class, and provide an API for basic operations. A (non exhaustive)
list of operations that would be needed are:
* contraction of two block-sparse tensors
* reshaping (i.e. merging and splitting indices) of a block-sparse tensor
* matrix decompositions (i.e . SVD, QR, eigen-decomps, ...)
* conjugation, addition, ... (all vector space operations)

We could add a new class `SparseNode`,  and a method `net.add_sparse_node(sparse_tensor)`,
and delegate contraction details to the node classes, such that `node1.tensordot(node2, edges)` works for both `Node` and `SparseNode`

I assume that this would result in changes in a lot of places in the API ...

viathor:
This makes sense and is a nice generalization of `net.add_copy_node()` and the related contraction code. One thing I'm not sure about is whether we need one or more types of sparse tensors with different symmetries (and different contraction code to exploit them).

PR #216, opened 2019-08-27
Flexible add node

chaserileyroberts:
Allow adding a node object from `add_node` instead of requiring a tensor. 

We will deprecate `add_copy_node` in the future in favor of doing it this way from now on.

fixes https://github.com/google/TensorNetwork/issues/62

relevant to https://github.com/google/TensorNetwork/issues/192 and https://github.com/google/TensorNetwork/issues/86

### DIFF
--- a/tensornetwork/network.py
+++ b/tensornetwork/network.py
@@ -150,7 +150,7 @@ class TensorNetwork:
 
     Args:
       value: Either the concrete tensor or an existing `Node` object that
-        has no accociated `TensorNetwork`. If a concrete tensor is given,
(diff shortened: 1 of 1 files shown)


mganahl:
it would be nice if `add_node` could also add `node`s from existing networks. `add_node` could just create a new `Node` if `node.network` is not `None`, with the `tensor` shared between the two nodes

Thenerdstation:
Just do `net.add_node(node.tensor)`

mganahl:
allright, no problem!

PR #218, opened 2019-08-29
Bug fix in split_node_rq

mganahl:
`split_node_rq` is currently doing the wrong thing; instead of returning `r, q` it returns `q, r`.

PR #225, opened 2019-09-03
Add a simple FiniteMPS class

mganahl:


mganahl:
Actually, I think I'll remove this PR again for now. It depends on two other PRs #224 and #218 which I locally merged into this one, so this might cause some annoying conflicts. I'll  request another PR once the other two are in

Issue #220, opened 2019-08-30
better error message for reorder_edges


mganahl:
`Node.reorder_edges(edges)` should give a clearer error message when one or more of the passed eges in `edges`do not belong to `Node`. Currently this code 
```python 
net=tn.TensorNetwork()
a = net.add_node(np.random.rand(3,3))
b = net.add_node(np.random.rand(3,3))
c = net.add_node(np.random.rand(3,3))

c.reorder_edges([b[0], a[0]])
```
throws
```ValueError: Given edge order does not match expected edges. Found: [<tensornetwork.network_components.Edge object at 0x146f9a978>, <tensornetwork.network_components.Edge object at 0x14a02fb00>], Expected: [<tensornetwork.network_components.Edge object at 0x14a02fb00>, <tensornetwork.network_components.Edge object at 0x14a02fc88>]```

It should be changed to something like 
``` 
ValueError: some edges in `edges` do not belong to `Node`
```

MichaelMarien:
@Thenerdstation : Housekeeping :)  this is fixed I believe?

Thenerdstation:
Yes it is.

In the future, be sure to do "fixes #ISSUE_NUM" in the comment of the PR so that it'll automatically close it.

PR #226, opened 2019-09-03
Reoder edges error

MichaelMarien:
Fix for #220. Let error print whether edges are missing, or there are additional ones. Still print out the object references as they can be useful for debugging.

Added some tests.


tensornetwork/network_components_test.py
```diff
@@ -208,6 +208,37 @@ def test_node_has_nondangling_edge_true(double_node_edge):
...
+  assert node[2] == e0
+
+
+def test_node_reorder_edges_raise_error_wrong_edges(single_node_edge):
+  node = single_node_edge.node
+  e0 = node[0]
+  e1 = node[1]
+  e2 = node[2]
+  edge = Edge(name="edge", node1=node, axis1=0)
+  with pytest.raises(ValueError):
```
Thenerdstation: Can you regex check for "Additional" and "Missing"?

tensornetwork/network_components.py
```diff
@@ -181,17 +181,24 @@ def reorder_edges(self, edge_order: List["Edge"]) -> "BaseNode":
...
     if not hasattr(self, '_tensor'):
       raise AttributeError("Please provide a valid tensor for this Node.")
 
-    if set(edge_order) != set(self.edges):
+    extra_edges = set(edge_order).difference(set(self.edges))
+    if extra_edges:
       raise ValueError("Given edge order does not match expected edges. "
-                       "Found: {}, Expected: {}".format(edge_order, self.edges))
+                       "Additional edges that do not belong to node found: "
+                       "{}".format(extra_edges))
```
Thenerdstation: Would it make more sense to print the edge names instead? It would be easier for humans to read, but if the names are conflicting it might become confusing. What is your opinion?

MichaelMarien: Well, it's what I did first, but then I also noticed the conflicting issue :(. 
We can go for the hybrid approach and print both name and object reference:
`Additional edges that do not belong to node found: Edge "my_edge_name" (tensornetwork.network_components.Edge object at 0x1a5c318f50>), Edge "my_second_edge_name' (another hash)`

Thenerdstation: We might be able to do that by changing the `__repr__` method on edges.

Thenerdstation:
I think the error message is fine for now (we can upgrade it later). Please just add the regex on the raise check and we can pull this in.

MichaelMarien:
> I think the error message is fine for now (we can upgrade it later). Please just add the regex on the raise check and we can pull this in.

Done!

PR #231, opened 2019-09-04
Remove dependencies

### LINKED ISSUE #227: Config imports torch and tensorflow
`config.py` imports torch and tensorflow, causing errors when a user doesn't have both installed.

We can mitigate this by adding try/except statements around the imports and uses.

MichaelMarien:
fixes #227 . Make sure tensor flow, pytorch and jax are not required. 
Note that at the moment I kept tensorflow as default backend and in requirements.txt, but already made tf optional in the code to prepare move to diff backend

### DIFF
--- a/tensornetwork/backends/backend_test.py
+++ b/tensornetwork/backends/backend_test.py
@@ -11,8 +11,6 @@ def clean_backend_import():
   sys.modules.pop('tensornetwork.backends.jax.jax_backend', None)
   sys.modules.pop('tensornetwork.backends.tensorflow.tensorflow_backend', None)
   sys.modules.pop('tensornetwork', None)
-  sys.modules.pop('tensornetwork.network', None)
-  sys.modules.pop('tensornetwork.network_components', None)
   yield # use as teardown
   sys.modules.pop('tensornetwork', None)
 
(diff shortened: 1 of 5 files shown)


tensornetwork/backends/backend_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@pytest.mark.usefixtures('no_backend_dependency')
+def test_config_pytorch_missing():
+  import tensornetwork.config
+  with pytest.raises(ImportError):
+    import torch
+
+
+@pytest.mark.usefixtures('no_backend_dependency')
```
Thenerdstation: Can you add a test where we just import `tensornetwork` with all of the backends removed and make sure that doesn't raise an error? That should cover 90% of the mistakes we make again in the future.

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -18,20 +18,23 @@
...
 import numpy
+
+
 Tensor = Any
 
 
 class JaxBackend(numpy_backend.NumPyBackend):
   """See base_backend.BaseBackend for documentation."""
 
+  try:
```
Thenerdstation: When would this code run? When the file is imported or when the class is used?

MichaelMarien: Hmmm, I'm confused. From the top of my mind, I'd say when the file is imported, right? Unless there's a diff python2-3 I'm not aware of. But his should cause tests too fail, are we still missing a test, or do the tests not work?

MichaelMarien: Ok nevermind, tests work but we should just add more tests of the ones you advised. Not just test import tensornetwork but all relevant submodules. I'll add those.

However, I have a small issue now. The reason I (mistakenly) moved them out of the init is that for pytorch, the typing in init: `def __init__(self, dtype: Optional[torch.dtype] = None):` forces me to already know torch...

Thenerdstation: Just make them `Any`s.

Thenerdstation: Just importing `tensornetwork` should be fine, don't worry about full module converage since all of the things the user is expected to have direct access to is already imported in `__init__`

MichaelMarien: Except for the backends! And that would have caught the bug you noticed. I also added most important submodules, doesn't hurt, right?

Thenerdstation: Might be annoying if we do a refactor, but refactors are annoying anyway so it's fine.

MichaelMarien: Actually, let me check what is already covered by the init of tensornetwork and the other tests and put it to a minimum, no need introducing extra annoyances

PR #235, opened 2019-09-05
Correcting errors in setup.py

kshithijiyer:


setup.py
```diff
@@ -31,14 +31,14 @@
 ]
 
 setup(
-    name='tensornetwork',
+    name='TensorNetwork',
     version=__version__,
     url='http://github.com/google/TensorNetwork',
-    author='The TensorNetwork Developers',
-    author_email=' chaseriley@google.com',
+    author='TensorNetwork Developers',
```
Thenerdstation: Do not remove the "The"

setup.py
```diff
@@ -31,14 +31,14 @@
 ]
 
 setup(
-    name='tensornetwork',
+    name='TensorNetwork',
```
Thenerdstation: Do not change this

kshithijiyer: But why?

Thenerdstation: Because it broke our pip install.

I had to fix it last time you changed it.
https://github.com/google/TensorNetwork/commit/ac0c770b2b3ac89bbab5689c3b1ac12d00f3c380

kshithijiyer: This is one mystery worth solving.

Thenerdstation: It's really not. Don't worry about it, we have much more important tasks to work on.

setup.py
kshithijiyer: We'll have to pick one change this to either "The TensorNetwork Developers" or "The TensorNetwork Authors". I see either of these being used and places.

Issue #240, opened 2019-09-09
Importing TensorNetwork is slow


gecrooks:
>python -m timeit "import tensornetwork as tn; tn.set_default_backend('numpy'); net = tn.TensorNetwork()"
1 loop, best of 5: 4.12 usec per loop

TensorNetwork is very slow to import. Squinting at the code, it looks like all the backends (numpy, tensorflow, torch, jax) are imported every time, irrespective of the backend that is actually used. In particular, importing TensorFlow is very slow.

Thenerdstation:
I think this PR should've fixed it? https://github.com/google/TensorNetwork/pull/231

We will push a new release soon once https://github.com/google/TensorNetwork/pull/236 is approved

gecrooks:
No, backend_factory imports everything, so I don't think #231 will fix the problem.

mganahl:
actually, it looks like tensorflow is still imported at import tensornetwork

Thenerdstation:
@mganahl How did you test that?

mganahl:
I get and import error if tensorflow isn't installed

Thenerdstation:
Could paste the trace on the other issue?

mganahl:
I  fixed the tensordot2 import in this PR #243 . Apparently importing as a class variable caused
tensorflow to be imported in tensordot2. 

I actually don't see that any of tensorflow, jax or torch get imported at tensorflow import (importing tensorflow in a clean env with only numpy, graphviz, h5py and opt_einsum installed works fine).

MichaelMarien:
@mganahl thanks for fixing. Depending on the outcome of #241 I can change the test to fix this behaviour (core reason is that I do not cleanup the import of `tensornetwork.backends.tensorflow` in the tests.

PR #241, opened 2019-09-09
Only import backend dependencies when needed. Fix for issue #240

gecrooks:
Before:

> time python -c "import tensornetwork as tn; tn.set_default_backend('numpy'); net = tn.TensorNetwork()"

real	0m1.990s
user	0m1.636s
sys	0m0.320s

After:
> time python -c "import tensornetwork as tn; tn.set_default_backend('numpy'); net = tn.TensorNetwork()"

real	0m0.543s
user	0m0.426s
sys	0m0.103s


tensornetwork/backends/backend_test.py
```diff
@@ -56,18 +56,3 @@ def test_config_tensorflow_missing_can_import_config():
   import tensornetwork.config
   with pytest.raises(ImportError):
     import tensorflow as tf
-
-
-@pytest.mark.usefixtures('no_backend_dependency')
```
Thenerdstation: Why did you delete this?

tensornetwork/backends/backend_factory.py
```diff
@@ -16,25 +16,30 @@
...
-    "jax": jax_backend.JaxBackend,
-    "shell": shell_backend.ShellBackend,
-    "pytorch": pytorch_backend.PyTorchBackend
-}
+import importlib
 
 
 def get_backend(name, dtype):
+
+  _BACKENDS = {
```
Thenerdstation: We can't do this, as it breaks imports internally. (We'll talk about this in person).

tensornetwork/backends/backend_test.py
gecrooks: It doesn't really test anything, as far as I can tell. All it's testing is that the mock import, defined in the same test file, raises an import error. So rather than fix the test for the other changes, it seems more sensible to just delete it

gecrooks: Actually, the same can be said for all the other tests in that same file.

Thenerdstation: Then we need to fix them.

Thenerdstation: The issue is we have no good tests for ensuring that when certain backends aren't installed, we can still use TensorNetwork fine for the other installed backends. These mock imports should have fixed this issue

tensornetwork/backends/backend_factory.py
Thenerdstation: Actually hold on let met see if that is actually true.

tensornetwork/backends/backend_test.py
Thenerdstation: Can you something like this? It will ensure we can still use TensornNetwork, but will also ensure we raise a smart error if one of the backends is missing.

```python
import tensornetwork as tn
tn.TensorNetwork("numpy")
with pytest.raises(ImportError):
  tn.TensorNetwork("tensorflow")
```

gecrooks: OK, updated import tests. Note that I've moved the imports so that an import error is thrown as soon as the backend module is imported, rather than waiting until the backend class is instantiated.

tensornetwork/backends/backend_factory.py
Thenerdstation: Confirmed this is broken internally.

tensornetwork/backends/shell/shell_backend.py
```diff
@@ -19,6 +19,9 @@
 import functools
 import operator
 from tensornetwork.backends import base_backend
+from ..numpy import numpy_backend
```
Thenerdstation: Doesn't doing this defeat the purpose? We are now importing all of the libraries when using Shell.

gecrooks: I'm unclear what the shell backend is for. If there is documentation, I haven't found it yet. But in the comments about dtypes the code states "# shell supports everything". So yes, you have to import everything to create the list of acceptable date types when you use the shell backend.

tensornetwork/backends/backend_test.py
MichaelMarien: Dont agree with this? The import should NOT raise an error? Also, for me, at least very naively, the tests do work (i.e. when I put an `import torch` in the init of the package, the tests fail. Not sure all is covered, but something is?

tensornetwork/backends/shell/shell_backend.py
Thenerdstation: Shell backend is a future feature. It's being used to help keep track of the real shape of a tensor for our jax_tpu backend (that is still in development) and was planned to be used to help calculate contraction paths before we integrated with opt_einsum

tensornetwork/backends/backend_test.py
gecrooks: Import errors should be raised on import, not (generally speaking) when you instantiate some class.

MichaelMarien: I agree in principle. But now we need to make sure none of the modules are imported by main package and this is I guess where the internally broken issue mentioned by @Thenerdstation comes in (just trying to understand that one)?

Thenerdstation: The main issue is that our internal build doesn't like dynamic imports. I hit this error when I tried using your branch internally.

```
tensornetwork/backends/backend_factory.py in get_backend(name, dtype)
     35   module_name, class_name = _BACKENDS[name]
     36   package_name = 'tensornetwork.backends'
---> 37   backend_module = importlib.import_module(module_name, package_name)
     38   backend_class = getattr(backend_module, class_name)
     39   dtypes = getattr(backend_module, 'supported_dtypes')


ModuleNotFoundError: No module named 'tensornetwork.backends.numpy'
```

MichaelMarien: @gecrooks  
> Actually, the same can be said for all the other tests in that same file.
Is there still an additional issue now with the tests?

gecrooks: No, I think I now see what they are trying to do.

gecrooks: OK, I don't need to use importlib. Can use conditional imports instead. Just less elegant.

MichaelMarien: @gecrooks 

> No, I think I now see what they are trying to do.

Thanks for confirming. They are a bit hacky & fragile though as you really need to cleanup all traces of all imported submodules... If you know of a more elegant way, I'd be happy to learn!

Thenerdstation: Still a no
```
tensornetwork/backends/backend_factory.py in get_backend(name, dtype)
     27     dtypes = tensorflow_backend.supported_dtypes
     28   elif name == "numpy":
---> 29     from .numpy import numpy_backend
     30     backend_class = numpy_backend.NumPyBackend
     31     dtypes = numpy_backend.supported_dtypes

ModuleNotFoundError: No module named 'tensornetwork.backends.numpy'
```

If the imports aren't at the top, they don't get included internally.

@gecrooks, just let me take the lead on this. There are a lot of subtle issues that need to be fixed.

Thenerdstation:
Closing as it breaks internal build.

PR #243, opened 2019-09-09
fix tensorflow import error

mganahl:
moved tensorflow import in TensorFlowBackend to __init__
importing it as class variable prompts import error

### DIFF
--- a/tensornetwork/backends/tensorflow/tensorflow_backend.py
+++ b/tensornetwork/backends/tensorflow/tensorflow_backend.py
@@ -26,18 +26,21 @@ Tensor = Any
 
 class TensorFlowBackend(base_backend.BaseBackend):
   """See base_backend.BaseBackend for documentation."""
-  from tensornetwork.backends.tensorflow import tensordot2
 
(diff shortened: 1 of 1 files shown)


Thenerdstation:
This is broken internally. 

The problem is is that if any code isn't imported simply by `import tensornetwork` with the exception of importing the backend libraries (`numpy`, `tensorflow`, etc) then that code doesn't get picked up by the bazel par_binary. https://google.github.io/subpar/subpar.html 

I'm still thinking about how to work around this...

MichaelMarien:
Can you clarify? Trying to look into root cause of this but I'm a bit confused atm.

Thenerdstation:
Basically our internal build system really does not like dynamic imports.

@MichaelMarien @mganahl @gecrooks I'll take the lead on this one guys. It's super fragile at the moment and I need to make sure everything works both on the pip and internally.

gecrooks:
That's fixable. Just less elegant.

Thenerdstation:
Closing this PR as it breaks our stuff internally.

PR #253, opened 2019-09-11
Complex initialization for randn + 

mganahl:
For complex dtypes `backend.randn` currently initializes only the real part with gaussian random
numbers. This PR implements initialization for the imaginary part as well.

It also fixes several bugs in backend tests, where dtype preservation wasn't correctly tested. 

This PR also sets the default dtype for the jax backend to `float64`.
The jax backend got its own `randn` function now, using `jax.random` module.

I've also added tests for the jax backend in  jax_backend_test.py.

I've added an optional `seed` argument to all `randn` backend functions


tensornetwork/backends/jax/jax_backend.py
```diff
@@ -49,4 +48,28 @@ def convert_to_tensor(self, tensor: Tensor) -> Tensor:
...
+  def randn(self,
+            shape: Tuple[int, ...],
+            dtype: Optional[numpy.dtype] = None,
+            seed: Optional[int] = None) -> Tensor:
+    if not seed:
+      seed = numpy.random.randint(0, 2**63)
+    key = self.jax.random.PRNGKey(seed)
+    if not dtype:
+      dtype = self.dtype
+    if not dtype:
```
Thenerdstation: Why are you doing this check twice?

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -49,4 +48,28 @@ def convert_to_tensor(self, tensor: Tensor) -> Tensor:
...
+      dtype = numpy.float64
+    if dtype is self.np.complex128:
```
Thenerdstation: Why did you duplicate this code for these two cases?

tensornetwork/backends/jax/jax_backend.py
mganahl: self.dtype can also be None

tensornetwork/backends/jax/jax_backend.py
mganahl: differentn dtypes in jax.random.normal

tensornetwork/backends/jax/jax_backend.py
Thenerdstation: do `dtype = self.dtype if self.dtype is not None else numpy.float64` instead.

tensornetwork/backends/jax/jax_backend.py
Thenerdstation: There is no need to duplicate the code. Refactor it by having the different types be variables. (Duplicating code is almost always bad practice since it can cause a lot of issues when you need to change/fix something later and you miss the duplication.)

PR #263, opened 2019-09-22
Fix tests

### LINKED ISSUE #262: ncon_interface tests fail

MichaelMarien:
fixes #262 

* removed all occurrences of setting default backend in tests and replaced with fixture
* `jax.config.update("jax_enable_x64", True)` is still set globally in tests, not sure if this is expected
* to make this future proof, it can be a good idea to do an explicit autoused fixture that resets all backend and configs to defaults, @Thenerdstation , wdyt?

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -16,9 +16,33 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 import pytest
+import jax
+import tensornetwork
+import tensorflow as tf
 
 
 @pytest.fixture(name="backend", params=["numpy", "tensorflow",
                                         "jax", "pytorch"])
 def backend_fixure(request):
   return request.param
+
+
+@pytest.fixture(autouse=True)
+def reset_default_backend():
+  tensornetwork.set_default_backend("numpy")
+  yield
+  tensornetwork.set_default_backend("numpy")
+
+
+@pytest.fixture(autouse=True)
+def enable_jax_64():
+  jax.config.update("jax_enable_x64", True)
+  yield
+  jax.config.update("jax_enable_x64", True)
+
+
+@pytest.fixture(autouse=True)
+def tf_enable_v2_behaviour():
+  tf.compat.v1.enable_v2_behavior()
+  yield
+  tf.compat.v1.enable_v2_behavior()
--- a/tensornetwork/__init__.py
+++ b/tensornetwork/__init__.py
@@ -12,4 +12,4 @@ from tensornetwork.utils import load
 
 def set_default_backend(backend: Text, dtype: Optional[Type] = None) -> None:
   config.default_backend = backend
-  config.default_dype = dtype
+  config.default_dtype = dtype
--- a/tensornetwork/ncon_interface_test.py
+++ b/tensornetwork/ncon_interface_test.py
@@ -30,10 +30,12 @@ def test_sanity_check(backend):
 def test_order_spec(backend):
   a = np.ones((2, 2))
 
-  result = ncon_interface.ncon([a, a], [(-1, 1), (1, -2)], out_order=[-1, -2], backend=backend)
+  result = ncon_interface.ncon(
+      [a, a], [(-1, 1), (1, -2)], out_order=[-1, -2], backend=backend)
   np.testing.assert_allclose(result, np.ones((2, 2)) * 2)
 
-  result = ncon_interface.ncon([a, a], [(-1, 1), (1, -2)], con_order=[1], backend=backend)
+  result = ncon_interface.ncon(
+      [a, a], [(-1, 1), (1, -2)], con_order=[1], backend=backend)
   np.testing.assert_allclose(result, np.ones((2, 2)) * 2)
 
   result = ncon_interface.ncon([a, a], [(-1, 1), (1, -2)],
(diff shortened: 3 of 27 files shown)


Thenerdstation:
> to make this future proof, it can be a good idea to do an explicit autoused fixture that resets all backend and configs to defaults, @Thenerdstation , wdyt?

Yes good idea

tensornetwork/ncon_interface_test.py
```diff
@@ -23,121 +23,267 @@
...
 
 
-def test_sanity_check():
+def test_sanity_check_numpy_backend():
+  result = ncon_interface.ncon([np.ones(
+      (2, 2)), np.ones((2, 2))], [(-1, 1), (1, -2)], backend='numpy')
+  np.testing.assert_allclose(result, np.ones((2, 2)) * 2)
+
+
+def test_order_spec_numpy_backend():
```
Thenerdstation: Add `backend` as an argument to each of these tests. It'll automatically test all 4 of our main backends.

tensornetwork/ncon_interface_test.py
```diff
@@ -23,121 +23,267 @@
...
+  a = np.ones((2, 2))
+
+  result = ncon_interface.ncon([a, a], [(-1, 1), (1, -2)], out_order=[-1, -2],
+                               backend='numpy')
```
Thenerdstation: Instead of doing just numpy, test all of the backends via my comment earlier.

MichaelMarien:
> > to make this future proof, it can be a good idea to do an explicit autoused fixture that resets all backend and configs to defaults, @Thenerdstation , wdyt?
> 
> Yes good idea

Added global fixtures to the config of pytest. Currently they set tensorflow v2 behaviour and jax 64 support.

tensornetwork/ncon_interface_test.py
MichaelMarien: Did this. One test fails for Jax. I can somehow see why, but don't understand why it does not happen for NumPy. It goes deeper into the contraction methods. I can't quick fix it, so I think it is better to make a separate issue.

MichaelMarien:
I changed quite some test files now. Pointing out to make sure it doesn't get lost in them that I corrected what I believe to be a typo in the config.py file, check it out. 

For two test files, Git shows that I fully "changed" them, I tried reverting that commit but doesn't help. Moreover my IDE just shows the one-line changes I make when I address the commit. Not sure why it happens or what I can do about it.

conftest.py
```diff
@@ -16,9 +16,33 @@
...
+
+@pytest.fixture(autouse=True)
+def enable_jax_64():
+  jax.config.update("jax_enable_x64", True)
+  yield
+  jax.config.update("jax_enable_x64", True)
+
+
+@pytest.fixture(autouse=True)
+def tf_enable_v2_behaviour():
```
Thenerdstation: Nice! Thanks for adding these!

tensornetwork/ncon_interface_test.py
Thenerdstation: We should really fix it then, as a failure of JAX with ncon is a big dea.

tensornetwork/__init__.py
```diff
@@ -12,4 +12,4 @@
 
 def set_default_backend(backend: Text, dtype: Optional[Type] = None) -> None:
   config.default_backend = backend
-  config.default_dype = dtype
+  config.default_dtype = dtype
```
Thenerdstation: Oh no....

Thenerdstation: How did pylint/pytype not catch this?

tensornetwork/ncon_interface_test.py
Thenerdstation: Actually, I changed my mind and I see your point.

Martin has some fixes for `ncon_interface_test` in his PR https://github.com/google/TensorNetwork/pull/264. Would you be ok if you revert the `ncon_interface_test` file for this PR? I really like your other fixes and I think we should have them, but ncon is a little messier than anyone expected.

MichaelMarien: I'll revert this, let's use Martin's fix!

Thenerdstation: Thank you!

Thenerdstation:
Ok, well good news is that the tests we expected to fail now actually fail.

You've done a lot and I really appreciate it. Let me quick push a fix on your branch and I'll pull this in.

Thenerdstation:
The `ncon_interface_test` is fixed and now tests all of our backends. There is a single failure for JAX, but it's not our fault.

```
_________________________________________________________ test_outer_product[jax] __________________________________________________________

backend = 'jax'

    def test_outer_product(backend):
      a = np.array([1, 2, 3])
      b = np.array([1, 2])
      res = ncon_interface.ncon([a, b], [(-1,), (-2,)], backend=backend)
      np.testing.assert_allclose(res, np.kron(a, b).reshape((3, 2)))
      res = ncon_interface.ncon([a, a, a, a], [(1,), (1,), (2,), (2,)],
>         backend=backend)

ncon_interface_test.py:112: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ncon_interface.py:124: in ncon
    res_node = tn.outer_product_final_nodes(out_edges)
network.py:665: in outer_product_final_nodes
    final_node = self.outer_product(final_node, node)
network.py:588: in outer_product
    new_tensor = self.backend.outer_product(node1.tensor, node2.tensor)
backends/numpy/numpy_backend.py:104: in outer_product
    return self.np.tensordot(tensor1, tensor2, 0)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = DeviceArray(14, dtype=int64), b = DeviceArray(14, dtype=int64), axes = 0

    @_wraps(onp.tensordot)
    def tensordot(a, b, axes=2):
      _check_arraylike("tensordot", a, b)
      if not (ndim(a) >= 1 and ndim(b) >= 1):
        msg = "tensordot requires a.ndim and b.dim to be at least 1, got {} and {}."
>       raise TypeError(msg.format(ndim(a), ndim(b)))
E       TypeError: tensordot requires a.ndim and b.dim to be at least 1, got 0 and 0.

../../../.local/lib/python3.6/site-packages/jax/numpy/lax_numpy.py:1738: TypeError
```

I'll bring this up on the JAX git, they're pretty fast at fixing things.

PR #264, opened 2019-09-23
Add support for `Node` with no network

mganahl:
This is the first in a series of PRs that adds support for a free `Node` object.
This PR implements basic operations for `Node` objects that have no associatesd `TensorNetwork`.
`TensorNetwork` is still supported in this PR


tensornetwork/__init__.py
```diff
@@ -1,6 +1,9 @@
 from __future__ import absolute_import
+
 from tensornetwork.network import TensorNetwork
-from tensornetwork.network_components import Node, Edge, CopyNode
+from tensornetwork.network_components import Node, Edge, CopyNode, BaseNode
+#pylint: disable=line-too-long
+from tensornetwork.network_components import conj, transpose, contract, contract_copy_node, contract_between, outer_product, outer_product_final_nodes, split_node, split_node_qr, split_node_rq, split_node_full_svd, reachable, contract_parallel, check_connected, check_correct, connect, flatten_edges, flatten_all_edges, flatten_edges_between, get_all_nondangling, get_all_nodes, get_all_edges, get_parallel_edges, get_shared_edges, _contract_trace
```
Thenerdstation: Can you put these in a different file other than `network_components`? Perhaps `network_operations` or something similar?

tensornetwork/contractors/stochastic_contractor.py
```diff
@@ -78,12 +78,18 @@ def contract_trace_edges(
...
+    # makes sure node_edges points to the original edges
+    # even after contracting the trace
+    node_edges = [e for e in node.edges]
+    node_dims = list(node.get_tensor().shape)
+    for edge, dim in zip(node_edges, node_dims):
+      if (not edge.is_disabled) and (edge.node1 is edge.node2):
         if edge not in trace_edges:
           # Contract trace edge
-          new_node = net.contract(edge)
+          new_node = net.contract(edge, name=node.name)
```
Thenerdstation: I see, so contracting a trace edge will name the "new" edge have the same name. I guess this is better than what we were doing before, though if someone sets the node `via contract(edge, name="...")`, we should default to doing that.

tensornetwork/ncon_interface_test.py
```diff
@@ -18,121 +18,148 @@
...
-tf.compat.v1.enable_v2_behavior()
 from tensornetwork import ncon_interface
 from tensornetwork.contractors.naive_contractor import naive
+from tensornetwork.backends import backend_factory
+from jax.config import config
+config.update("jax_enable_x64", True)
+tf.compat.v1.enable_v2_behavior()
+
+
+#TODO: fix jax backend and add to fixture
```
Thenerdstation: We should fix that now. What was the issue?

tensornetwork/network_components.py
```diff
@@ -844,12 +1040,1369 @@ def _load_edge(cls, edge_data: h5py.Group,
...
+
+  return nodes
+
+
+def get_all_edges(nodes: Collection[BaseNode]) -> Set[Edge]:
+  """Return the set of edges of all nodes."""
+  edges = set()
+  for node in nodes:
+    edges |= set(node.edges)
+  return edges
```
Thenerdstation: Let's move all of these methods to a different file to prevent this file from being 2,400 lines long (!)

Thenerdstation:
Looks good. The tests pass which is nice. The only travis error is it doesn't like `Collections` being used as a type. I think we can switch this to either `List` or `Sequence`.

mganahl:
Allright, let me quickly do that. I'll switch it to Union[List,Set]

tensornetwork/__init__.py
mganahl: OK, can try, this will cause some cyclic impoprt issues with network_operations.connect and disconnect. I think I will have to keep those in network_components.py

Thenerdstation: That's fine by me.

tensornetwork/ncon_interface_test.py
Thenerdstation: Remove this comment as it's mostly fixed now.

tensornetwork/contractors/stochastic_contractor.py
mganahl: Good point. I just changed _contract_trace to do the same thing.

Thenerdstation:
I'll give this a full review once the build passes.

Looks like the only issue right now is
```
File "/home/travis/build/google/TensorNetwork/tensornetwork/tests/network_components_free_test.py", line 491, in test_copy_node_load: No attribute 'copy_node_dtype' on tensornetwork.network_components.BaseNode [attribute-error]
```

mganahl:
I just fixed this one, will push in a second

mganahl:
btw pytype got really slow with this PR, no clue why

tensornetwork/network_components.py
mganahl: this is a bit finicky due to the cyclic import issues. `__xor__, __or__ `and `__matmul__` require
`connect, contract_between, outer_product, _contract_trace` which in turn require
all the helpers `_flatten_edges_..., _remove_edges_...`.
I moved as many functions as possible for now into network_operations.py.

mganahl:
OK I just pushed the latest stuff. This builds on my machine, so hopefully it will go through on travis

Thenerdstation:
I'll push any quick fixes if need be

mganahl:
There are still a bunch of changes and tests that are waiting to be moved into master. Since this PR is already so large I'd suggest to open another PR once this is through.

Thenerdstation:
Agreed.

tensornetwork/__init__.py
```diff
@@ -1,6 +1,13 @@
 from __future__ import absolute_import
+
 from tensornetwork.network import TensorNetwork
-from tensornetwork.network_components import Node, Edge, CopyNode
+from tensornetwork.network_components import Node, Edge, CopyNode, BaseNode
+#pylint: disable=line-too-long
+from tensornetwork.network_operations import conj, transpose, split_node, split_node_qr, split_node_rq, split_node_full_svd, reachable, check_connected, check_correct, get_all_nodes, get_all_edges
+
+from tensornetwork.network_components import contract, contract_copy_node, contract_between, outer_product, outer_product_final_nodes, contract_parallel, flatten_edges, get_all_nondangling, flatten_all_edges, flatten_edges_between, get_parallel_edges, get_shared_edges, _contract_trace
```
Thenerdstation: Don't include the private `_contract_trace` method.

tensornetwork/contractors/greedy_contractor_test.py
```diff
@@ -19,6 +19,8 @@
 import numpy as np
 from tensornetwork.contractors import greedy_contractor
 from tensornetwork import network
+import tensorflow as tf
+tf.enable_v2_behavior()
```
Thenerdstation: This isn't required anymore after https://github.com/google/TensorNetwork/commit/113015b9ff034d38432e30a663f048067f25c6de

tensornetwork/network.py
```diff
@@ -16,20 +16,21 @@
...
 from __future__ import print_function
-import collections
 import h5py
 # pylint: disable=line-too-long
 from typing import Any, Sequence, List, Set, Optional, Union, Text, Tuple, Type, Dict, BinaryIO
 import numpy as np
-import weakref
 from tensornetwork import config
-from tensornetwork import network_components
+#pylint: disable=useless-import-alias
```
Thenerdstation: Why did you disable this check?

tensornetwork/network.py
```diff
@@ -76,6 +77,7 @@ def _new_node_name(self, name: Optional[Text]) -> Text:
   def dtype(self) -> Type[np.number]:
     return self.backend.dtype
 
+  # pylint: disable=redefined-outer-name
```
Thenerdstation: Why did you disable this?

tensornetwork/network_components_test.py
```diff
@@ -433,21 +469,29 @@ def test_copy_node_load(tmp_path, backend):
     node_group = node_file.create_group('node_data')
     node_group.create_dataset('signature', data=node.signature)
     node_group.create_dataset('name', data=node.name)
+    node_group.create_dataset(
+        'copy_node_dtype', data=np.dtype(node.copy_node_dtype).name)
```
Thenerdstation: Why should the `copy_node_dtype` be different from the dtype of the node? Seems redundant.

tensornetwork/network_operations.py
```diff
@@ -0,0 +1,656 @@
...
+        continue
+      reachable_nodes.add(next_node)
+      _reachable(next_node)
+
+  _reachable(node)
+
+  return reachable_nodes
+
+
+def reachable_iterative(node: BaseNode) -> Set[BaseNode]:
```
Thenerdstation: Why do you have 3 versions of the exact some thing? This is really unneeded. Just use the iterative version.

tensornetwork/network_operations.py
```diff
@@ -0,0 +1,656 @@
...
+  Returns:
+    A list of `BaseNode` objects that can be reached from `node`
+    via connected edges.
+  Raises:
+    ValueError: If an unknown value for `strategy` is passed.
+  """
+  if isinstance(nodes, BaseNode):
+    nodes = [nodes]
+  reachable_nodes = {nodes[0]}
+  if strategy == 'recursive':
```
Thenerdstation: There is no reason to have strategies here. Just pick one.

tensornetwork/tests/network_components_free_test.py
```diff
@@ -0,0 +1,806 @@
+import numpy as np
+import tensorflow as tf
+import pytest
+from collections import namedtuple
+import h5py
+from tensornetwork.network_components import Node, CopyNode, Edge
+import tensornetwork as tn
+
+tf.compat.v1.enable_v2_behavior()
```
Thenerdstation: Remove this as it is not needed after https://github.com/google/TensorNetwork/pull/263

tensornetwork/tests/tensornetwork_test.py
```diff
@@ -640,6 +645,10 @@ def test_weakref(backend):
     e.node2
 
 
+# This test no longer tests weakref behaviour
```
Thenerdstation: Weak refs are still in the code though right?

tensornetwork/visualization/graphviz.py
```diff
@@ -21,6 +21,8 @@
 from tensornetwork import network
 
 
+#pylint: disable=no-member
```
Thenerdstation: Why is this here twice?

tensornetwork/tests/tensornetwork_test.py
mganahl: Yes, still there

mganahl:
`ncon` actually works seamlessly with `Node` as well now, which is pretty nice

tensornetwork/network_components_test.py
mganahl: that's a technical detail: the current implementation of `CopyNode` instantiates
`CopyNode._tensor` when calling `CopyNode.tensor` property. This instantiation
is always done in `numpy`, irrespective of the actual backend, and the `copy_node_dtype` is used as dtype if this instantiation. `CopyNode.tensor` then calls `backend.convert_to_tensor(numpy_tensor)` on the numpy tensor to create the actual tensor object returned by `CopyNode.tensor`. The dtype of the `numpy_tensor` has to match 
the dtype of `backend`, and the lazy solution to this problem is to let the use pass this numpy dtype. The better thing would be to determine the numpy-dtype from `backend.dtype`. That's a bit ugly to code but not a real problem. I'll add it

Thenerdstation:
Are you ok if I push some commits to this PR so we get this in faster?

mganahl:
Absolutely, go for it!

mganahl:
Hey @Thenerdstation, did you start on this already? I'm busy right now with another project which should be done by mid next week. Should be free to work on the PR then

tensornetwork/__init__.py
mganahl: added new file network_operations.py

tensornetwork/ncon_interface_test.py
mganahl: removed

tensornetwork/network_components.py
mganahl: I cut this down as much as possible for now. If it's still too long we need to find a good way of
breaking this apart without causing circular imports

tensornetwork/__init__.py
mganahl: removed

tensornetwork/network.py
mganahl: how can I make network_components available?

tensornetwork/network.py
mganahl: that's an artifact i forgot to remove. It's gone

tensornetwork/network_operations.py
mganahl: Kept the deque version

tensornetwork/tests/network_components_free_test.py
mganahl: done

tensornetwork/visualization/graphviz.py
mganahl: removed

mganahl:
Hey @Thenerdstation, I can start to work on this again (I fixed some comments and added a few more tests yesterday night). How should we proceed?

tensornetwork/network_operations.py
```diff
@@ -0,0 +1,575 @@
...
+  Raises:
+    ValueError: If an unknown value for `strategy` is passed.
+  """
+
+  if isinstance(nodes, BaseNode):
+    nodes = {nodes}
+
+  reachable_nodes = set()
+  for node in nodes:
+    reachable_nodes |= _reachable(node)
```
Thenerdstation: This will run on O(n^2) instead of O(n) for the worst case if there is more than 1 node given to `reachable`

We'll fix it later, I'm just leaving this comment to show that I noticed it.

Thenerdstation:
I'm going to fix my one comment and then pull this in.

Great work Martin!

mganahl:
Thanks! I just noticed that some docstrings are still outdated. I'll go through them and fix them

Thenerdstation:
> Thanks! I just noticed that some docstrings are still outdated. I'll go through them and fix them

Do that in a different PR. This is big enough as is

mganahl:
allright!

mganahl:
oh man, I just upgraded as well ...

Thenerdstation:
Though we should switch all of our code to be TF 2.0 compliant. We'll do that in a later PR.

PR #271, opened 2019-10-01
Removed TensorNetwork from tests/

Thenerdstation:
Removed the use of TensorNetwork in several of the files of `tests/`

Added `copy` and `remove_node` operations to work outside of a TensorNetwork.

Finally, I removed the weakrefs from the code. We may consider adding a `node.deallocate()` method in the future to free up memory manually.

The serialization tests still need to be fixed.

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -20,7 +20,6 @@ import collections
 from typing import Any, Dict, List, Optional, Set, Text, Tuple, Type, Union, \
   overload, Sequence
 import numpy as np
-import weakref
 from abc import ABC
 from abc import abstractmethod
 import h5py
@@ -98,7 +97,7 @@ class BaseNode(ABC):
     if axis_names is not None:
       self.add_axis_names(axis_names)
     else:
-      self._axis_names = None
+      self._axis_names = [str(i) for i in range(len(shape))]
 
     self._signature = -1
 
@@ -939,9 +938,9 @@ class Edge:
     if self.is_disabled:
       raise ValueError(
           'Edge has been disabled, accessing node1 is no longer possible')
-    if self._node1() is None:
+    if self._node1 is None:
       raise ValueError("node1 for edge '{}' no longer exists.".format(self))
-    return self._node1()
+    return self._node1
 
   @property
   def node2(self) -> Optional[BaseNode]:
@@ -950,9 +949,9 @@ class Edge:
           'Edge has been disabled, accessing node2 is no longer possible')
     if self._is_dangling:
       return None
-    if self._node2() is None:
+    if self._node2 is None:
       raise ValueError("node2 for edge '{}' no longer exists.".format(self))
-    return self._node2()
+    return self._node2
 
   @node1.setter
   def node1(self, node: BaseNode) -> None:
@@ -960,7 +959,7 @@ class Edge:
       raise ValueError(
           'Edge has been disabled, setting node1 is no longer possible')
     # pylint: disable=attribute-defined-outside-init
-    self._node1 = weakref.ref(node)
+    self._node1 = node
 
   @node2.setter
   def node2(self, node: Optional[BaseNode]) -> None:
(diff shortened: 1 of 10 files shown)


tensornetwork/network_operations.py
```diff
@@ -91,6 +92,84 @@ def transpose(node: BaseNode,
...
+  """
+  broken_edges_by_name = {}
+  broken_edges_by_axis = {}
+  for i, name in enumerate(node.axis_names):
+    if not node[i].is_dangling() and not node[i].is_trace():
+      edge1, edge2 = disconnect(node[i])
+      new_broken_edge = edge1 if edge1.node1 is not node else edge2
+      broken_edges_by_axis[i] = new_broken_edge
+      broken_edges_by_name[name] = new_broken_edge
+  node.disable()
```
mganahl: I think we don't want to disable the node. The user might still want to use it for something else.

tensornetwork/network_components.py
```diff
@@ -1254,7 +1253,7 @@ def flatten_edges(edges: List[Edge],
     if axis_names:
       node.axis_names = [axis_names[n] for n in range(len(node.edges))]
     else:
-      node.axis_names = None
+      node.axis_names = ["__unnamed_edge__" for n in range(len(node.edges))]
```
mganahl: we should stick to the same convention as in BaseNode.__init__

mganahl: or at least make axis_names unique (looks like a typo though)

tensornetwork/network_operations.py
mganahl: We could consider to remove `BaseNode.disable()`

Thenerdstation: Good point. `refresh_edges` maybe?

tensornetwork/network_components.py
Thenerdstation: Good point. Switched to `str(n)`

tensornetwork/network_operations.py
mganahl: I think we can just remove `node.disable()` can leave it at that. The only non-trivial case is if 
`node` has trace edges. As a user, I would expect them to be left untouched
since I was only requesting to disconnect `node` from any other nodes.
This is exactly what's happening then.

tensornetwork/network_components.py
```diff
@@ -1075,9 +1074,9 @@ def disconnect(self,
     if self.is_dangling():
       raise ValueError("Cannot break dangling edge {}.".format(self))
     if not edge1_name:
-      edge1_name = '__unnamed_edge__'
+      edge1_name = '__broke_edge1__{}'.format(self.name)
```
sleichen: Not a fan of "break," "broke," or "broken" for describing edge that have been disconnected. Either "disconnected" or "cut" would be preferable. I would go with "disconnected" to avoid confusion. Docstring should also be changed to avoid using "break."

tensornetwork/network_operations.py
```diff
@@ -91,6 +92,84 @@ def transpose(node: BaseNode,
...
+
+def remove_node(node: BaseNode
+               ) -> Tuple[Dict[Text, Edge], Dict[int, Edge]]:
+  """Remove a node from the network.
+
+  Args:
+    node: The node to be removed.
+
+  Returns:
+    broken_edges_by_name: A Dictionary mapping `node`'s axis names to
```
sleichen: Same comment about "broken." The connotation with "broken" makes it sound like something has gone wrong.

tensornetwork/network_operations.py
Thenerdstation: Good point. Done.

tensornetwork/network_components.py
Thenerdstation: Done, replaced `broken` with `disconnected`

tensornetwork/network_operations.py
Thenerdstation: Done.

PR #272, opened 2019-10-01
remove `TensorNetwork` from `ncon`

mganahl:
1. removes the `TensorNetwork` dependency from `ncon`
2. adds support for using `ncon` on `BaseNode`


tensornetwork/ncon_interface.py
```diff
@@ -17,19 +17,23 @@
 from __future__ import division
 from __future__ import print_function
 import warnings
-from typing import Any, Sequence, List, Optional, Union, Text, Tuple, Dict
+from typing import Any, Sequence, List, Optional, Union, Text, Tuple, Dict, Set
 from tensornetwork import network
 from tensornetwork import network_components
-
+# pylint: disable=useless-import-alias
```
Thenerdstation: If they're useless why did you add them?

tensornetwork/ncon_interface.py
```diff
@@ -17,19 +17,23 @@
...
+import tensornetwork.config as config
+# pylint: disable=useless-import-alias
+import tensornetwork.backends.backend_factory as backend_factory
 Tensor = Any
 
 
-def ncon(tensors: Sequence[Tensor],
+def ncon(tensors: Sequence[Union[network_components.BaseNode, Tensor]],
```
Thenerdstation: I still don't see the value in having `BaseNode`s as an input to `ncon`

tensornetwork/ncon_interface.py
```diff
@@ -72,10 +76,30 @@ def ncon(tensors: Sequence[Tensor],
         `tensornetwork.config.default_backend`.
 
     Returns:
-      A `Tensor` resulting from the contraction of the tensor network.
+      The result of the contraction. The result if returned in a `Node` 
+        if all elements of `tensors` are `BaseNode` objects, else
+        it is returned as a `Tensor` object.
```
Thenerdstation: Your typing hint suggests that it will always be a `Tensor`.

tensornetwork/ncon_interface.py
Thenerdstation: (Just do `from tensornetwork import config`)

tensornetwork/ncon_interface_test.py
```diff
@@ -16,20 +16,39 @@
 from __future__ import print_function
 import pytest
 import numpy as np
+import tensornetwork as tn
```
Thenerdstation: Prefer to not import `tensornetwork` directly in tests outside of the `tests/` directory. just use the modules directly like you do below.

tensornetwork/ncon_interface_test.py
```diff
@@ -38,14 +57,38 @@ def test_order_spec(backend):
...
-                               out_order=[-1, -2])
+                               out_order=[-1, -2],
+                               backend=backend)
+
   np.testing.assert_allclose(result, np.ones((2, 2)) * 2)
 
 
+def test_node_order_spec(backend):
+  node = tn.Node(np.ones((2, 2)), backend=backend)
+  result = ncon_interface.ncon([node, node], [(-1, 1), (1, -2)],
```
Thenerdstation: Passing two of the exact same node seems like a really bad test case. 

Also, what should `ncon` do if two nodes are connect? Right now it ignores them and `ncon` just uses the underlying tensor, but that feels sketchy too.

tensornetwork/ncon_interface.py
mganahl: fixed

tensornetwork/ncon_interface.py
mganahl: it just feels more streamlined. With `Node` our API provides a wrapper for `Tensor` so that we can do backend independent tensor contractions. The way I use it is to build my code around `Node` objects, not `Tensor` (apart from initialization). So it seems nice not to have to access `Node.tensor` when calling `ncon`.

tensornetwork/ncon_interface.py
mganahl: fixed it

tensornetwork/ncon_interface_test.py
mganahl: fixed

tensornetwork/ncon_interface_test.py
mganahl: I think that's what it should do.  a call to `ncon` has to specify the whole contraction or throw an exception otherwise. This should override any preexisting connections on nodes. In essence, it's like using `Node.tensor` except that you don't have to access `Node.tensor` yourself.

tensornetwork/ncon_interface_test.py
```diff
@@ -148,12 +290,26 @@ def test_contraction(backend):
...
+  res = ncon_interface.ncon([a, a, a], [(-1, 1, 2), (1, 2, 3), (3, -2, -3)],
+                            backend=backend)
+  res_np = tensor.reshape((2, 4)) @ tensor.reshape((4, 2)) @ tensor.reshape(
+      (2, 4))
   res_np = res_np.reshape((2, 2, 2))
-  np.testing.assert_allclose(res, res_np)
+  np.testing.assert_allclose(res.tensor, res_np)
+
+
+# def test_backend_network(backend):
```
Thenerdstation: Delete commented out code

(Why was this deleted)?

pylintrc
```diff
@@ -53,6 +53,7 @@ load-plugins=
 #
 # CHANGED:
 disable=
+    bad-option-value,
```
Thenerdstation: Why did you add this?

mganahl: see #277. I'll remove it again

tensornetwork/ncon_interface_test.py
mganahl: added it back

pylintrc
Thenerdstation: Actually keep it. Just wanted to know why it was added 😊.

PR #273, opened 2019-10-01
Node contractors

mganahl:
Add support for contraction path finding for `Iterable[BaseNode]`


tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -16,15 +16,77 @@
 
 import functools
 import opt_einsum
+import tensornetwork as tn
```
Thenerdstation: Don't import the base tensornetwork module like because you'll get a cyclic import. Prefer to just import what you need

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -16,15 +16,77 @@
...
+  Returns:
+    Final node after full contraction.
+  """
+  nodes_set = set(nodes)
+  tn.check_connected(nodes_set)
+  edges = tn.get_all_nondangling(nodes_set)
+  #output edge order has to be determinded before any contraction
+  #(edges are refreshed after contractions)
+  if output_edge_order is None:
+    output_edge_order = list(
```
Thenerdstation: We really should add a `get_all_dangling` method.

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -16,15 +16,77 @@
...
+
+  Returns:
+    Final node after full contraction.
+  """
+  nodes_set = set(nodes)
+  tn.check_connected(nodes_set)
+  edges = tn.get_all_nondangling(nodes_set)
+  #output edge order has to be determinded before any contraction
+  #(edges are refreshed after contractions)
+  if output_edge_order is None:
```
Thenerdstation: Also, why are we not raising an error when the `output_edge_order` isn't given and the number of output edge's is > 1?

Setting it to be the difference of two sets gives it a non-deterministic order and can cause really subtle bugs.

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -16,15 +16,77 @@
...
+    raise ValueError("output edges are not all dangling.")
+
+  for edge in edges:
+    if not edge.is_disabled:  #if its disabled we already contracted it
+      if edge.is_trace():
+        nodes_set.remove(edge.node1)
+        nodes_set.add(tn.contract_parallel(edge))
+
+  if not tn.get_all_nondangling(nodes_set):
+    # There's nothing to contract.
```
Thenerdstation: We should ensure that there is only 1 node in this set too.

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -16,15 +16,77 @@
...
+  path, nodes = utils.get_path(nodes_set, algorithm)
+  for a, b in path:
+    new_node = nodes[a] @ nodes[b]
+    nodes.append(new_node)
+    nodes = utils.multi_remove(nodes, [a, b])
+
+  # if the final node has more than one edge,
+  # output_edge_order has to be specified
+  final_node = nodes[0]  # nodes were connected, we checked this
+  if (len(final_node.edges) > 1) and (output_edge_order is None):
```
Thenerdstation: Since you set `output_edge_order` to be a list it it's `None` earlier in the code, this check never hits. 

Please move this check to before contracting any nodes.

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -73,17 +135,29 @@ def base(net: TensorNetwork,
...
+    nodes: Union[TensorNetwork, Iterable[BaseNode]],
+    output_edge_order: Sequence[Edge] = None,
+    memory_limit: Optional[int] = None):  # -> Union[BaseNode, TensorNetwork]:
   """Optimal contraction order via `opt_einsum`.
 
   This method will find the truly optimal contraction order via
   `opt_einsum`'s depth first search algorithm. Since this search is
   exhaustive, if your network is large (n>10), then the search may
-  take longer than just contracting in a suboptimal way.
+nn  take longer than just contracting in a suboptimal way.
```
Thenerdstation: Why the "nn"?

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -73,17 +135,29 @@ def base(net: TensorNetwork,
...
 
-  Args:
+  nArgs:
```
Thenerdstation: Seems like a typo. Plz fix

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -156,12 +231,57 @@ def greedy(net: TensorNetwork,
...
   """
   alg = functools.partial(opt_einsum.paths.greedy, memory_limit=memory_limit)
-  return base(net, alg, output_edge_order)
+  return base(nodes, alg, output_edge_order)
 
 
-def auto(net: TensorNetwork,
-         output_edge_order: Sequence[Edge] = None,
-         memory_limit: Optional[int] = None) -> TensorNetwork:
+def _auto_nodes(
```
Thenerdstation: I think you duplicated the code. Please revise the code to avoid the copying and pasting. (If we decide to increase or decrease these cutoffs, we'd have to change it twice and its easy to miss.)

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -205,10 +325,22 @@ def auto(net: TensorNetwork,
...
-           optimizer: Any,
-           output_edge_order: Sequence[Edge] = None,
-           memory_limit: Optional[int] = None) -> TensorNetwork:
+def auto(
+    nodes: Union[TensorNetwork, Iterable[BaseNode]],
+    output_edge_order: Sequence[Edge] = None,
+    memory_limit: Optional[int] = None):  # -> Union[BaseNode, TensorNetwork]:
+
+  if isinstance(nodes, TensorNetwork):
+    return _auto_network(nodes, output_edge_order, memory_limit)
```
Thenerdstation: Would it not just be possible to do `_auto_nodes(net.nodes_set,...)` instead of having 2 different methods?

tensornetwork/ncon_interface.py
```diff
@@ -17,19 +17,23 @@
...
-
+# pylint: disable=useless-import-alias
+import tensornetwork.config as config
+# pylint: disable=useless-import-alias
+import tensornetwork.backends.backend_factory as backend_factory
 Tensor = Any
 
 
-def ncon(tensors: Sequence[Tensor],
+def ncon(tensors: Sequence[Union[network_components.BaseNode, Tensor]],
```
Thenerdstation: I don't see the value in having `Nodes`s as an input to ncon.

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
mganahl: that's a good point. I think we should throw an error. I'll fix this

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
mganahl: that's checked above by `tn.check_connected(nodes_set)`. If `nodes_set` had more than one node in it and `get_all_nondangling(nodes_set)` returned an empty `Iterable` then `tn.check_connected(nodes_set)`  above
will have already raised an exception.

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
mganahl: has become obsolete

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
mganahl: typo

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
mganahl: fixed

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -62,9 +126,9 @@ def base(net: TensorNetwork,
   if (len(final_node.edges) <= 1) and (output_edge_order is None):
     output_edge_order = list((net.get_all_edges() - net.get_all_nondangling()))
   elif (len(final_node.edges) > 1) and (output_edge_order is None):
```
Thenerdstation: I think you can remove this check now since you raise it earlier?

tensornetwork/contractors/opt_einsum_paths/utils.py
```diff
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Helper methods for `path_contractors`."""
-
+import tensornetwork as tn
```
Thenerdstation: Same comment as before

For tests outside of `tests/`, prefer to just import the modules you actually need.

Thenerdstation: And on that note, everything INSIDE of `tests/` should ONLY have `import tensornetwork as tn` so we can make sure the public API behaves as we expect it to behave.

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
mganahl: This check actually needs to be there. I changed the check in`_base_nodes` before. So, I also tried to remove the code duplication of `_base_network` and `_base_node`, but it turns out to lead to some annoying typing problems (essentially, pytype seems not to like functions returning different types) which I couldn't easily fix. So I left it as is for now and added a todo.

mganahl:
@Thenerdstation that's it so far from my side.

Thenerdstation:
> Please don't move the move the test files.
> 
> Unit tests should stay next to their representative file. Only unit tests that test the public API should be under `tests/`

Oh I see, they ARE public API tests.

I take this back now. This is fine as is.

PR #280, opened 2019-10-02
Fixed 1/2 of tests in examples

Thenerdstation:
Fixes for FFT and SAT tensor network examples.

Simple MERA and wave functions still need to be upgraded.

PR #282, opened 2019-10-02
Removed Naive and Stochastic contractors

Thenerdstation:
* Remove Naive and Stochastic contractors.
* Made ncon_network return a list of nodes in the order they were given.
* Added get_all_dangling() methods

Build will pass once #280 is in.


gecrooks:
There's still a `tf.enable_v2_behavior()` in `tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py` that should presumable be removed.

PR #286, opened 2019-10-03
Upgrade to tensorflow 2.0, the new standard

gecrooks:


Thenerdstation:
Congrats, you found another bug in TF.

Set the failing test to be skipped.

gecrooks:
Well that's odd. The test passes with tensorflow 2.0.0-alpha0, but fails with tensorflow 2.0.0 !?

PR #305, opened 2019-10-07
Fix dtypes

### LINKED ISSUE #301: NumpyBackend.dtype returns wrong type
The return type of `NumpyBackend.dtype` is currently given by the type of the `dtype` argument at initialization, but it should always yield a `numpy.dtype` object:
```python 
be = NumpyBackend(dtype=np.float64)
be.dtype #this is a `type` object, i.e. a `np.float64`
```
The fix is 

```python 
class NumPyBackend(base_backend.BaseBackend):
  """See base_backend.BaseBackend for documentation."""

  def __init__(self, dtype: Optional[numpy.dtype] = None):
    super(NumPyBackend, self).__init__()
    self.np = numpy
    self.name = "numpy"
    self._dtype =dtype# -> self._dtype = numpy.dtype(dtype)

```

MichaelMarien:
fixes #301


tensornetwork/backends/jax/jax_backend.py
```diff
@@ -44,29 +44,34 @@ def convert_to_tensor(self, tensor: Tensor) -> Tensor:
...
-          key, shape,
-          dtype=real_dtype) + complex_dtype(1j) * self.jax.random.normal(
-              key_2, shape, dtype=real_dtype)
 
-    if dtype is self.np.complex128:
+      real_part = self.jax.random.normal(key, shape, dtype=real_dtype)
+      complex_part = self.jax.random.normal(key_2, shape, dtype=real_dtype)
+      unit = (np.complex64(1j) if complex_dtype == np.dtype(np.complex64)
+              else np.complex128(1j))
+      return real_part + unit*complex_part
```
Thenerdstation: Add spaces around the `*`

MichaelMarien: done! strange pylint does not catch this

MichaelMarien:
@Thenerdstation we have some warnings in build, 
`Installing collected packages: tensorflow-estimator, google-pasta, protobuf, termcolor, wrapt,
astor, grpcio, absl-py, keras-applications, werkzeug, markdown, tensorboard, gast, keras-preprocessing, tensorflow, decorator, networkx, importlab, ninja, pyyaml, typed-ast, pytype, torch, fastcache, jax, scipy, jaxlib
313Successfully installed absl-py-0.8.0 astor-0.8.0 decorator-4.4`
not sure if we should be worried

Thenerdstation:
That's not a concern

PR #310, opened 2019-10-08
Added "Free Node" API support for the graph_viz visulalization. #296

### LINKED ISSUE #296: Add "Free Node" API support for the graph_viz visulalization.
Right now, `tn.to_graphviz` only supports the `TensorNetwork` object, which has been deprecated. It should support passing in an iterable collection of nodes.

longdead-ai:
In the changed function both tha attributes are initated with none given value not passed by the user. I both are null at the same time then it raises a value error.

Fixes #296

### DIFF
--- a/tensornetwork/visualization/graphviz.py
+++ b/tensornetwork/visualization/graphviz.py
@@ -14,13 +14,14 @@
 """Implementation of TensorNetwork Graphviz visualization."""
 
 import graphviz
-from typing import Optional, Text
+from typing import Optional, Text, Sequence
 from tensornetwork import network
 
 
 #pylint: disable=no-member
-def to_graphviz(net: network.TensorNetwork,
+def to_graphviz(net: network.TensorNetwork = None,
                 graph: Optional[graphviz.Graph] = None,
+                nodes: Optional[Sequence] = None,
                 include_all_names: bool = False,
                 engine: Text = "neato") -> graphviz.Graph:
   """Create a graphviz Graph that is isomorphic to the given TensorNetwork.
@@ -29,6 +30,9 @@ def to_graphviz(net: network.TensorNetwork,
     net: A `TensorNetwork`.
     graph: An optional `graphviz.Graph` object to write to. Use this only
       if you wish to set custom attributes for the graph.
+    nodes: An optinal Sequence object(list, numpy.array, set,etc). If only
+    nodes are known and if the Tensornet passed is none. This should be 
+    passed if net is none.
     include_all_names: Whether to include all of the names in the graph.
       If False, all names starting with '__' (which are almost always just
       the default generated names) will be dropped to reduce clutter.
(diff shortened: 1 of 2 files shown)


tensornetwork/visualization/graphviz.py
```diff
@@ -40,36 +44,47 @@ def to_graphviz(net: network.TensorNetwork,
...
-            label="",
-            _attributes={"style": "invis"})
-        graph.edge(
-            "{}_{}".format(node.signature, i),
-            str(node.signature),
-            label=edge_label)
+  if not net:
+    for node in net.nodes_set:
+      if not node.name.startswith("__") or include_all_names:
+        label = node.name
```
Thenerdstation: I don't understand this part.

If the network is `None`, why would you iterate over `net.nodes_set`?

tensornetwork/visualization/graphviz.py
```diff
@@ -40,36 +44,47 @@ def to_graphviz(net: network.TensorNetwork,
...
+              str(edge.node2.signature),
+              label=edge_label)
+  else:
+    if not nodes:
+      for node in nodes:
+        if not node.name.startswith("__") or include_all_names:
+          label = node.name
+        else:
+          label = ""
+        graph.node(str(node.signature), label=label)
```
Thenerdstation: It seems like you're forgetting to set the edges here too.

tensornetwork/visualization/graphviz.py
```diff
@@ -14,13 +14,14 @@
...
-from typing import Optional, Text
+from typing import Optional, Text, Sequence
 from tensornetwork import network
 
 
 #pylint: disable=no-member
-def to_graphviz(net: network.TensorNetwork,
+def to_graphviz(net: network.TensorNetwork = None,
                 graph: Optional[graphviz.Graph] = None,
+                nodes: Optional[Sequence] = None,
```
Thenerdstation: `Optional[Sequence[BaseNode]]`

Thenerdstation: I also don't like having separate `net` and `nodes` arguments. Can you modify this so that either can be passed to the first argument?

tensornetwork/visualization/graphviz.py
```diff
@@ -29,6 +30,9 @@ def to_graphviz(net: network.TensorNetwork,
     net: A `TensorNetwork`.
     graph: An optional `graphviz.Graph` object to write to. Use this only
       if you wish to set custom attributes for the graph.
+    nodes: An optinal Sequence object(list, numpy.array, set,etc). If only
```
Thenerdstation: Don't support numpy array of nodes.

tensornetwork/visualization/graphviz.py
Thenerdstation: Also by merging the args into one, it gets rid of the need for the `None` checks.

tensornetwork/visualization/graphviz.py
```diff
@@ -40,36 +44,47 @@ def to_graphviz(net: network.TensorNetwork,
...
-        # to connect to.
-        graph.node(
-            "{}_{}".format(node.signature, i),
-            label="",
-            _attributes={"style": "invis"})
-        graph.edge(
-            "{}_{}".format(node.signature, i),
-            str(node.signature),
-            label=edge_label)
+  if not net:
```
Thenerdstation: I think you might have worked a little too hard here. You should just be able to set the `nodes` variable if a network was passed and iterate over that instead of having duplicate code separated by `if` statements.

So it would look something like this.

```python
if net is not None:
  nodes = net.nodes_set
for node in nodes:
   ...
```

tensornetwork/visualization/graphviz.py
longdead-ai: Sure, can I just Assert for the datatype and proceed?

tensornetwork/visualization/graphviz.py
MichaelMarien: Typo 'optinal' in line 33

Thenerdstation:
Closing this as it was fixed in #333

PR #311, opened 2019-10-09
Support copying induced subgraph

### LINKED ISSUE #266: Support induce subgraph in the copy function
Now that we are moving away from using `TensorNetwork` object, we should still allow copying a network via passing a collection of nodes to a `tensornetwork.copy(nodes)` function. 

Along with this, it would be nice if when we passed a subset of all the nodes that it would create an induced subgraph copy of the given nodes.

By this I mean that if `a` and `b` are connected by an edge and we do `tensornetwork.copy({a})`, then the returned set of nodes should only be a copy of `a` and the edge that used to connect `a` and `b` becomes a dangling edge.

lingxz:
Fixes #266

### DIFF
--- a/tensornetwork/network_operations.py
+++ b/tensornetwork/network_operations.py
@@ -96,7 +96,7 @@ def copy(nodes: Iterable[BaseNode],
   """Copy the given nodes and their edges.
 
   This will return a dictionary linking original nodes/edges 
-  to their copies.
+  to their copies. If nodes A and B are connected but only A is passed in to be copied, the edge between them will become a dangling edge.
 
   Args:
     nodes: An `Iterable` (Usually a `List` or `Set`) of `Nodes`.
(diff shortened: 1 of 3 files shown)


tensornetwork/network_operations.py
```diff
@@ -131,18 +132,17 @@ def copy(nodes: Iterable[BaseNode],
     node1 = edge.node1
     axis1 = edge.node1.get_axis_number(edge.axis1)
 
-    if not edge.is_dangling():
+    if not edge.is_dangling() and edge.node2 in node_dict:
```
Thenerdstation: It might be the case that the edge is only connected to a node in the node set via `edge.node2`. I'm not sure that's handled here.

tensornetwork/tests/network_test.py
```diff
@@ -78,6 +78,20 @@ def test_tnwork_copy_identities(backend):
     assert not edge_dict[edge] is edge
 
 
+def test_tnwork_copy_subgraph(backend):
+  a = tn.Node(np.random.rand(3, 3, 3), name='a', backend=backend)
+  b = tn.Node(np.random.rand(3, 3, 3), name='b', backend=backend)
+  c = tn.Node(np.random.rand(3, 3, 3), name='c', backend=backend)
+  a[0] ^ b[1]
+  edge2 = b[2] ^ c[0]
```
Thenerdstation: If you add another test where `edge2 = c[0] ^ b[2]` does it still work?

tensornetwork/network_operations.py
lingxz: no it wasn't handled. didn't notice, fixed it now, thanks!

tensornetwork/tests/network_test.py
lingxz: added a added a test for it and works now.

PR #313, opened 2019-10-10
Added support for contracting a subgraph

### LINKED ISSUE #289: Support induced subgraphs with contractors
Now that contractors work with contracting sets of nodes instead of the `TensorNetwork` object, it would be really nice to allow contract a subgraph of a network by only passing some of the network's nodes.

Basically the idea is that if a `Node` is connected to another `Node` that isn't in the passed set of nodes, then treat the edges between these two nodes as if they were normal dangling edges.

Thenerdstation:
Fixes #289

### DIFF
--- a/tensornetwork/contractors/opt_einsum_paths/path_contractors.py
+++ b/tensornetwork/contractors/opt_einsum_paths/path_contractors.py
@@ -21,7 +21,6 @@ from tensornetwork.network import TensorNetwork
 from tensornetwork.network_operations import check_connected, get_all_edges, get_subgraph_dangling
 # pylint: disable=line-too-long
 from tensornetwork.network_components import get_all_nondangling, contract_parallel
-    
 from tensornetwork.network_components import Edge, BaseNode
 from tensornetwork.contractors.opt_einsum_paths import utils
 from typing import Any, Optional, Sequence, Iterable, Union
--- a/tensornetwork/contractors/opt_einsum_paths/path_contractors.py
+++ b/tensornetwork/contractors/opt_einsum_paths/path_contractors.py
@@ -17,9 +17,11 @@
 import functools
 import opt_einsum
 from tensornetwork.network import TensorNetwork
-from tensornetwork.network_operations import check_connected, get_all_edges
+# pylint: disable=line-too-long
+from tensornetwork.network_operations import check_connected, get_all_edges, get_subgraph_dangling
 # pylint: disable=line-too-long
 from tensornetwork.network_components import get_all_nondangling, contract_parallel
+    
 from tensornetwork.network_components import Edge, BaseNode
 from tensornetwork.contractors.opt_einsum_paths import utils
 from typing import Any, Optional, Sequence, Iterable, Union
@@ -48,20 +50,21 @@ def _base_nodes(nodes: Iterable[BaseNode],
   """
   nodes_set = set(nodes)
   check_connected(nodes_set)
-  edges = get_all_nondangling(nodes_set)
+  edges = get_all_edges(nodes_set)
   #output edge order has to be determinded before any contraction
   #(edges are refreshed after contractions)
   if output_edge_order is None:
-    output_edge_order = list(
-        (get_all_edges(nodes) - get_all_nondangling(nodes)))
+    output_edge_order = list(get_subgraph_dangling(nodes))
     if len(output_edge_order) > 1:
       raise ValueError("The final node after contraction has more than "
-                       "one dangling edge. In this case `output_edge_order` "
+                       "one remaining edge. In this case `output_edge_order` "
                        "has to be provided.")
 
-  if set(output_edge_order) != (
-      get_all_edges(nodes) - get_all_nondangling(nodes)):
-    raise ValueError("output edges are not all dangling.")
+  if set(output_edge_order) != get_subgraph_dangling(nodes):
+    raise ValueError(
+        "output edges are not equal to the remaining "
+        "non-contracted edges of the final node."
+    )
 
   for edge in edges:
     if not edge.is_disabled:  #if its disabled we already contracted it
@@ -69,7 +72,7 @@ def _base_nodes(nodes: Iterable[BaseNode],
         nodes_set.remove(edge.node1)
         nodes_set.add(contract_parallel(edge))
 
-  if not get_all_nondangling(nodes_set):
+  if len(nodes_set) == 1:
     # There's nothing to contract.
     return list(nodes_set)[0]
 
--- a/tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py
+++ b/tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py
@@ -96,3 +96,28 @@ def test_custom_sanity_check(backend):
   optimizer = PathOptimizer()
   final_node = path_contractors.custom(nodes, optimizer)
   np.testing.assert_allclose(final_node.tensor, np.ones(5) * 2.0)
+
+
+def test_subgraph_contraction(backend, path_algorithm):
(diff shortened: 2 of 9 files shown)


tensornetwork/contractors/opt_einsum_paths/utils.py
```diff
@@ -42,7 +41,7 @@ def _get_path_nodes(nodes: Iterable[BaseNode], algorithm: Algorithm
   sorted_nodes = sorted(nodes, key=lambda n: n.signature)
 
   input_sets = [set(node.edges) for node in sorted_nodes]
-  output_set = get_all_edges(nodes) - get_all_nondangling(nodes)
+  output_set = get_subgraph_dangling(nodes)
```
viathor: This seems to be a change in behavior.

Suppose we have two nodes A and B connected by a single edge and let nodes={A}. Then all edges is {(A, B)} and all non-dangling edges is also {(A, B)}, so before this PR, ouput_set would be empty. However, get_subgraph_dangling(nodes) will return the non-empty set {(A, B)} since the edge (A, B) is "relatively dangling".

Is this intentional?

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -17,9 +17,11 @@
 import functools
 import opt_einsum
 from tensornetwork.network import TensorNetwork
-from tensornetwork.network_operations import check_connected, get_all_edges
+# pylint: disable=line-too-long
+from tensornetwork.network_operations import check_connected, get_all_edges, get_subgraph_dangling
 # pylint: disable=line-too-long
 from tensornetwork.network_components import get_all_nondangling, contract_parallel
+
```
viathor: nit: probably no need for the newline here

tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py
```diff
@@ -96,3 +96,28 @@ def __call__(self, inputs, output, size_dict, memory_limit=None):
...
   np.testing.assert_allclose(final_node.tensor, np.ones(5) * 2.0)
+
+
+def test_subgraph_contraction(backend, path_algorithm):
+  a = tn.Node(np.ones((2, 2)), backend=backend)
+  b = tn.Node(np.ones((2, 2)), backend=backend)
+  c = tn.Node(np.ones((2, 2)), backend=backend)
+  a[0] ^ b[0]
+  c[0] ^ b[1]
+  result = path_algorithm({a, b}, [a[1], b[1]])
```
viathor: Could you add an assert verifying that this intermediate result is correct?

tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py
```diff
@@ -96,3 +96,28 @@ def __call__(self, inputs, output, size_dict, memory_limit=None):
   optimizer = PathOptimizer()
   final_node = path_contractors.custom(nodes, optimizer)
   np.testing.assert_allclose(final_node.tensor, np.ones(5) * 2.0)
+
+
+def test_subgraph_contraction(backend, path_algorithm):
+  a = tn.Node(np.ones((2, 2)), backend=backend)
+  b = tn.Node(np.ones((2, 2)), backend=backend)
+  c = tn.Node(np.ones((2, 2)), backend=backend)
```
viathor: This test would be more robust if the tensors were more interesting, e.g.

```python
a = 10 + tn.Node(np.arange(4).reshape((2, 2)), backend=backend)
a = 20 + tn.Node(np.arange(4).reshape((2, 2)), backend=backend)
```

The present form of the tensors allows certain bugs to pass through unnoticed, e.g. if we were inadvertently contracting b with c (instead of a with b).

tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py
```diff
@@ -96,3 +96,28 @@ def __call__(self, inputs, output, size_dict, memory_limit=None):
...
+  a = tn.Node(np.eye(2), backend=backend)
+  b = tn.Node(np.eye(2), backend=backend)
+  c = tn.Node(np.eye(2), backend=backend)  
+  d = tn.Node(np.eye(2), backend=backend)
+  a[0] ^ b[1]
+  b[0] ^ c[1]
+  c[0] ^ d[1]
+  d[0] ^ a[1]
+  ab = path_algorithm({a, b}, list(tn.get_subgraph_dangling({a, b})))
+  cd = path_algorithm({c, d}, list(tn.get_subgraph_dangling({c, d})))
```
viathor: ditto about verifying the intermediate tensors `ab` and `cd`

tensornetwork/contractors/opt_einsum_paths/utils.py
Thenerdstation: Yes, this is intentional and the point of the PR. So that edge between A and B would not be contracted, and is effectively treated as a "relatively dangling" edge.

This allows you to find optimal contractions of subnetworks to the tensor network. For example, in the [quantum supremacy Nasa paper](https://arxiv.org/pdf/1905.00444.pdf), they contract their network into 3 sections before combining those section together. With this PR, you can now pass only the nodes in say section D and it will find an optimal contraction path just for those nodes, leaving all of the nodes that are "relatively dangling" untouched.

tensornetwork/tests/network_operations_test.py
```diff
@@ -246,3 +246,29 @@ def test_reachable_disconnected_2(backend):
                            nodes[1]])) == {nodes[0], nodes[1], nodes[2]}
   nodes[2][1] ^ nodes[3][0]  #connect third and fourth node
   assert set(tn.reachable([nodes[0], nodes[1]])) == set(nodes)
+
```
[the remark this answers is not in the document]
Thenerdstation: surprised the linter didn't notice...

tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py
Thenerdstation: Added

tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py
Thenerdstation: Added

tensornetwork/contractors/opt_einsum_paths/path_contractors_node_test.py
Thenerdstation: Added

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
Thenerdstation: Removed

tensornetwork/tests/network_operations_test.py
```diff
@@ -246,3 +246,29 @@ def test_reachable_disconnected_2(backend):
...
+def test_subgraph_disconnected_nodes(backend):
+  a = tn.Node(np.eye(2), backend=backend)
+  b = tn.Node(np.eye(2), backend=backend)
+  c = tn.Node(np.eye(2), backend=backend)
+  a[0] ^ b[0]
+  b[1] ^ c[1]
+  edges = tn.get_subgraph_dangling({a, c})
+  assert edges == {a[0], a[1], c[0], c[1]}
+
```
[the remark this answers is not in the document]
Thenerdstation: Done

tensornetwork/contractors/opt_einsum_paths/utils.py
viathor: Right, makes sense. Thanks.

PR #321, opened 2019-10-15
Typo fix

Thenerdstation:
words are hard.


Thenerdstation:
Yes good point. I'll fix it

Issue #196, opened 2019-08-20
Outer product doesn't work for nodes with 0 legs


Thenerdstation:
Outer product doesn't work for nodes with 0 edges on (I believe) all backends.

This should be fairly easy to fix, just switch to a scalar multiplication if either of the Tensors have rank 0.

Thenerdstation:
#330 fixed this.

PR #326, opened 2019-10-16
Return scalar multiplication in outer product if rank of one tenser i…

jziub:
This intends to fix [#196 ](https://github.com/google/TensorNetwork/issues/196)

### DIFF
--- a/tensornetwork/backends/base_backend.py
+++ b/tensornetwork/backends/base_backend.py
@@ -324,3 +324,16 @@ class BaseBackend:
     """
     raise NotImplementedError(
         "Backend '{}' has not implemented eighs_lanczos.".format(self.name))
+
+  def multiply(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
+    """
+      Return the default multiplication of `tensor`.
+      A backend can override such implementation.
+      Args:
+        tensor1: A tensor.
+        tensor2: A tensor.
+      Returns:
+        Tensor
+    """
+    raise NotImplementedError(
+        "Backend '{}' has not implemented multiply.".format(self.name))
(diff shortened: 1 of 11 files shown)


jziub:
@googlebot I signed it!

jziub:
@googlebot I fixed it.

tensornetwork/backends/base_backend.py
```diff
@@ -324,3 +324,15 @@ def eigsh_lanczos(
...
+    """
+      Return the default multiplication of `tensor`.
+      A backend can override such implementation.
+      Args:
+        tensor1: A tensor.
+        tensor2: A tensor.
+      Returns:
+        Tensor
+    """
+    return tensor1 * tensor2
```
MichaelMarien: I'd prefer to raise an NotImplementedError here instead and add it to all the separate backends even if the implementation is identical. Anyway, definitely we need a test for this wherever it ends up.

tensornetwork/tests/network_test.py
```diff
@@ -230,6 +230,16 @@ def test_outer_product(backend):
   assert d.name == "D"
 
 
+def test_outer_product_without_legs(backend):
+  a = tn.Node(np.ones((2, 4, 5)), name="A", backend=backend)
+  b = tn.Node(np.ones(()), name="B", backend=backend)
+  c = tn.outer_product(a, b, name="C")
+
```
MichaelMarien: perhaps test the other order also the arguments of outer product?

tensornetwork/backends/base_backend.py
Thenerdstation: +1

tensornetwork/tests/network_test.py
Thenerdstation: +1

jziub:
Closed this pull request as the commits were overwritten. Will send in a new pull request regarding this. Sorry about the confusion!

PR #332, opened 2019-10-18
Added ability to pass Edge instance to reachable method

mrunderline:
@Thenerdstation, it's almost done! feel free to comment or ask any changes.

### DIFF
--- a/tensornetwork/network_operations.py
+++ b/tensornetwork/network_operations.py
@@ -594,7 +594,7 @@ def reachable(inputs: Union[BaseNode, Iterable[BaseNode], Edge, Iterable[Edge]])
   """
   Computes all nodes reachable from `node` or `edge.node1` by connected edges.
   Args:
-    inputs: A `BaseNode`/`Edge` or collection of `BaseNodes`/`BaseEdges`
+    inputs: A `BaseNode`/`Edge` or collection of `BaseNodes`/`Edges`
   Returns:
     A list of `BaseNode` objects that can be reached from `node`
     via connected edges.
--- a/tensornetwork/network_operations.py
+++ b/tensornetwork/network_operations.py
@@ -606,6 +606,8 @@ def reachable(inputs: Union[BaseNode, Iterable[BaseNode], Edge, Iterable[Edge]])
     inputs = {inputs}
   elif isinstance(inputs, Edge):
     inputs = {inputs.node1}
+  elif isinstance(inputs, list) and all(isinstance(_, Edge) for _ in inputs):
+    inputs = {_.node1 for _ in inputs}
   return _reachable(set(inputs))
 
 
--- a/tensornetwork/network_operations.py
+++ b/tensornetwork/network_operations.py
@@ -590,11 +590,11 @@ def _reachable(nodes: Set[BaseNode]) -> Set[BaseNode]:
   return seen_nodes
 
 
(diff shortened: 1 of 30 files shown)


tensornetwork/network_operations.py
```diff
@@ -590,21 +590,23 @@ def _reachable(nodes: Set[BaseNode]) -> Set[BaseNode]:
...
 
 
-def reachable(nodes: Union[BaseNode, Iterable[BaseNode]]) -> Set[BaseNode]:
+def reachable(inputs: Union[BaseNode, Iterable[BaseNode], Edge, Iterable[Edge]]) -> Set[BaseNode]:
   """
-  Computes all nodes reachable from `node` by connected edges.
+  Computes all nodes reachable from `node` or `edge.node1` by connected edges.
   Args:
-    nodes: A `BaseNode` or collection of `BaseNodes`
+    inputs: A `BaseNode`/`Edge` or collection of `BaseNodes`/`BaseEdges`
```
Thenerdstation: No such thing as BaseEdges

mrunderline: Ops! it's been changed and corrected!

tensornetwork/network_operations.py
```diff
@@ -590,21 +590,23 @@ def _reachable(nodes: Set[BaseNode]) -> Set[BaseNode]:
...
   Raises:
     ValueError: If an unknown value for `strategy` is passed.
   """
 
-  if isinstance(nodes, BaseNode):
-    nodes = {nodes}
-  return _reachable(set(nodes))
+  if isinstance(inputs, BaseNode):
+    inputs = {inputs}
+  elif isinstance(inputs, Edge):
```
Thenerdstation: Do you want to also support the case where it is an iterable of `Edge`s?

mrunderline:
I think it's almost done.
A very very simple test added; cause i'm still beginner at it 😅

tensornetwork/tests/network_operations_test.py
```diff
@@ -218,6 +218,13 @@ def test_reachable(backend):
   _ = [nodes[n][0] ^ nodes[n + 1][1] for n in range(len(nodes) - 1)]
   assert set(nodes) == tn.reachable(nodes[0])
 
+  edges = []
```
Thenerdstation: Dont append this to an existing test, Instead, create a new test.

Also, I would highly recommend not creating an `Edge` object directly like this, as the `Node`s that the edge is trying connect won't exist in those `Node`s. Instead, you should create edges via the `tn.connect` operation. (You can also use the `^` symbol if you don't like typing.) See the other tests or our tutorial on `tensornetwork.readthedocs.io` for examples.

mrunderline: thanks for your attention and kind comments.
a new commit pushed, please check it out.

tensornetwork/network_operations.py
```diff
@@ -590,21 +590,25 @@ def _reachable(nodes: Set[BaseNode]) -> Set[BaseNode]:
...
+    inputs = {inputs.node1}
+  elif isinstance(inputs, list) and all(isinstance(_, Edge) for _ in inputs):
```
Thenerdstation: Can you change the `_` to `x`? Generally `_` is only used when we want to ignore the variable, but in this case we are actually using it.

Thenerdstation: Same just below.

PR #333, opened 2019-10-18
Nuke tensornetwork

### LINKED ISSUE #312: Nuke `TensorNetwork`
We should nuke `TensorNetwork` soon, as nearly all features support free nodes at this point.

MichaelMarien:
fixes #312 

Special attention to bucket_contractor where I added a rather naive alternative implementation. Dealing with partial contractions is a bit tricky with just list of nodes as compared to network.


MichaelMarien:
I didn't pay attention to them locally, and see all the examples tests are failing. It's still quite some effort to get the examples up-to-date. Should I also take this up in this PR?

Thenerdstation:
Looks like its simple MERA and wavefunction tests that are failing. @amilsted Do you think you would have time to upgrade these to the new API?

amilsted:
> Looks like its simple MERA and wavefunction tests that are failing. @amilsted Do you think you would have time to upgrade these to the new API?

This was in the stuff I sent to @mganahl, but I guess it got lost? Anyway, the code is writ - where do you want me to put it?

Thenerdstation:
> Anyway, the code is writ - where do you want me to put it?

Just make a new PR and replace the code.

amilsted:
D'oh. Yes, will do.

mganahl:
sorry @amilsted, I actually never pulled it in!

tensornetwork/network.py
```diff
@@ -1,1083 +0,0 @@
...
-    """
-    logging.warn(
-        "The TensorNetwork object has been DEPRECATED and will be removed "
-        "in future releases.\n"
-        "Updating your code is fairly easy. For a detailed tutorial, please "
-        "visit "
-        "https://medium.com/@keeper6928/upgrading-your-tensornetwork-code-b032f0ab3dd4"
-    )
-    if backend is None:
-      backend = config.default_backend
```
Thenerdstation: Maybe instead of totally nuking `TensorNetwork`, we have it raise a `DeprecationWarning` https://docs.python.org/3/library/exceptions.html.

This gives users a slightly cleaner heads up. It will also link to the upgrade post and help them port their code.

tensornetwork/backends/backend_test.py
```diff
@@ -3,6 +3,8 @@
 import sys
 import pytest
 import numpy as np
+import tensornetwork as tn
+from tensornetwork import connect, contract
```
Thenerdstation: Perfer not to import `tensornetwork` directly in unit tests outside of `tests/` directory (and prefer to ONLY do `import tensornetwork as tn` inside of `tests/`)

tensornetwork/network.py
MichaelMarien: Yeah, fair enough. However, some parts of the code still won't work, i.e. contractors only accept iterables of nodes in this PR. Perhaps then we should not have nuked anything and deprecated everything :)?

tensornetwork/backends/backend_test.py
MichaelMarien: Can you clarify the reasons? I don't mind changing, but looks like there's a subtlety to learn here I'm not fully aware of.

Thenerdstation: Not really a subtlety more of a structure thing I'm weakly trying to enforce. Tests next to files are supposed to be totally self contained, just testing that one module. Thus, anything you need in the test should be imported in manually.

However, tests in `tests/` should be public facing user test. Basically, test the behavior of the library as a whole so that specific use cases are less likely to break. Think of difference like integration vs unit testing.

Like I said it's a weak enforcement, and that rule is likely broken somewhere in the codebase.

tensornetwork/network.py
Thenerdstation: What's wrong with contractors just accepting an iterable of `Node`s? I've been using the for personal research for a while and I haven't run into any issues. (though sometimes I use a `tn.reachable(edge.node1)` hack, that should be fixed with #332).

MichaelMarien: Nothing wrong, on the contrary is the only supported way in this PR. What I'm concerned about is that we confuse users by
1. not nuking the TensorNetwork class, but adding a deprecation warning
2. nuking some of the functionalities around the TensorNetwork object (like being able to pass them to a contractor) which the PR is removing without any warning.

tensornetwork/backends/backend_test.py
MichaelMarien: got it.

tensornetwork/network.py
Thenerdstation: Aw I see your concern. The `DeprecationWarning` is actually an exception in python (terrible name, but we don't have control over that). All of the functionality around and inside of `TensorNetwork` will be removed.

Basically, I only want `TensorNetwork` to look like this:

```python
class TensorNetwork():
  def __init__(...):
    raise DeprecationWarning("The TensorNetwork object has been...")
```
So when someone updates their tensornetwork version and tries to run their code, it should immediately crash with this message:
```python
>>> tn.TensorNetwork()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  ...
DeprecationWarning: The TensorNetwork object has been...
```

Make sense? Sorry for the confusion.

MichaelMarien: Makes sense! I was indeed thinking about a deprecation decorator that leaves functionality intact for now.

MichaelMarien: Added this

tensornetwork/backends/backend_test.py
MichaelMarien: Tried to fix it in most places

PR #335, opened 2019-10-21
Support reduce density (#291).

### LINKED ISSUE #291: Implement tn.reduced_density operation
I want an easy want to calculate reduced density operations. The API would look like.

```python
nodes_dict, edges_dict = tn.reduced_density(traced_out_edges)
```

Underneath, we would copy/conjugate all of the nodes in reachable by the `traced_out_edges` and return a dictionary of the old nodes/edges to the copied nodes/edges. We would then connect all of the edges in `traced_out_edges` to their copies.

jziub:
This intends to fix #291.


tensornetwork/network_operations.py
```diff
@@ -713,3 +713,28 @@ def contract_trace_edges(node: BaseNode) -> BaseNode:
...
+  # Copy and conjugate all reachable nodes.
+  new_nodes, new_edges = copy(old_nodes, True)
+
+  # Add each edge to the copied nodes as new edge.
+  for traced_out_edge in traced_out_edges:
+    for new_node in new_nodes:
+      # What if a traced_out_edge is already an edge of the traget node?
+      # What axis should be set here axis1 or axis2?
+      # Do we need to update the nodes traced_out_edge connected to?
+      new_node.add_edge(traced_out_edge, traced_out_edge.axis1)
```
Thenerdstation: Nope. Not quiet what you need to do.

I stead, just do 

```
for traced_out_edge in traced_out_edges:
  new_edge[traced_out_edge] ^ traced_out_edge
```

The `^` operation is the same as doing `tn.connect`

tensornetwork/network_operations.py
```diff
@@ -713,3 +713,28 @@ def contract_trace_edges(node: BaseNode) -> BaseNode:
...
+
+  Args:
+    traced_out_edges:
+
+  Returns:
+
+  """
+
+  # Get all reachable nodes.
+  old_nodes = get_all_nodes(traced_out_edges)
```
Thenerdstation: You'll want to also do `tn.reachable(...)` so that you get all possible edges, not just some of them.

tensornetwork/network_operations.py
jziub: Got it. Please take another look!

jziub: One thing I am not sure is the dangling. It seems if the input is a connected graph. There will be chance that the ^ operation will fail when it tries to connect a traced_out_edge to its copy as the edge may not be dangling.

Thenerdstation: Good observation!

Yes, `traced_out_edges` should only be dangling edges and we should raise a very clear error in the case that it's not. Can you add and test that as well?

tensornetwork/tests/network_operations_test.py
```diff
@@ -274,3 +274,24 @@ def test_full_graph_subgraph_dangling(backend):
...
+
+def test_reduced_density(backend):
+  a = tn.Node(np.random.rand(3, 3, 3), name="A", backend=backend)
+  b = tn.Node(np.random.rand(3, 3, 3), name="B", backend=backend)
+  c = tn.Node(np.random.rand(3, 3, 3), name="C", backend=backend)
+  edges = tn.get_all_edges({a, b, c})
+
+  node_dict, edge_dict = tn.reduced_density([a[0], b[1], c[2]])
+
+  assert a[0].is_dangling() == False and \
```
Thenerdstation: Make these separate assert lines.

Basically, never use `\`. I'm not sure why our pylint doesn't complain...

jziub: Done.

tensornetwork/network_operations.py
jziub: Done.

Thenerdstation:
Code looks great! Though now I'm questioning my decision on the return type of this function. 

@mganahl @MichaelMarien I would love your opinions on this.

tensornetwork/tests/network_operations_test.py
```diff
@@ -274,3 +274,37 @@ def test_full_graph_subgraph_dangling(backend):
   b[1] ^ c[1]
   edges = tn.get_subgraph_dangling({a, b, c})
   assert edges == {a[1], c[0]}
+
+
```
MichaelMarien: Can you add a test where we do the final contraction and materialise the reduced density matrix and check that it's what you expect?

MichaelMarien:
> Code looks great! Though now I'm questioning my decision on the return type of this function.
> 
> @mganahl @MichaelMarien I would love your opinions on this.

Return type makes sense to me. It definitely allows the user full flexibility. What's the alternative you had in mind?

tensornetwork/tests/network_operations_test.py
jziub: Sure. Could you elaborate or maybe point me to some sample code about final contraction and materialization? Thanks!

Thenerdstation: Don't worry about it. I'll push a test.

Thenerdstation:
I had no alternatives. Just wanted second opinion. Thanks! We'll stick with the current return type.

tensornetwork/network_operations.py
```diff
@@ -723,3 +723,30 @@ def contract_trace_edges(node: BaseNode) -> BaseNode:
...
+  Connects each given edge to its copy.
+  Args:
+    traced_out_edges: A list of dangling edges
+  Returns:
+    A tuple containing:
+      node_dict: A dictionary mapping the nodes to their copies.
+      edge_dict: A dictionary mapping the edges to their copies.
+  """
+
+  if len(list(filter(lambda x: not x.is_dangling(), traced_out_edges))) > 0:
```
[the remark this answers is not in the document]
jziub: Done.

tensornetwork/tests/network_operations_test.py
jziub: Thanks. Please cc'd me the PR. I am interested in that too.

Thenerdstation: I plan on pushing it on this PR. ;)

PR #336, opened 2019-10-21
Free simple_mera and wavefunctions of TensorNetwork objects.

amilsted:
This makes the examples and their tests survive the removal of the TensorNetwork object.


Thenerdstation:
Linting errors are being fixed with #337

PR #346, opened 2019-10-25
Update README.md

Thenerdstation:
Added logo.


README.md
```diff
@@ -1,4 +1,6 @@
 # TensorNetwork
```
sleichen: I think this can be removed

Thenerdstation: Done.

PR #358, opened 2019-10-31
Support algebraic operations for BaseNode 

### LINKED ISSUE #292: algebraic operations for BaseNode
It would be convenient to have addition, subtraction, multiplication by scalar and division by scalar for `BaseNode`s:

```python 
a, b =tn.Node(np.random.rand(3,3)), tn.Node(np.random.rand(3,3))
c=a + b
d = a*10
e = a/5
```

jziub:
This is to fix #292 .

### DIFF
--- a/tensornetwork/backends/shell/shell_backend.py
+++ b/tensornetwork/backends/shell/shell_backend.py
@@ -257,11 +257,13 @@ class ShellBackend(base_backend.BaseBackend):
     return ShellTensor((a * b).shape)
 
   def add(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
-    a = np.ones(tensor1.shape)
-    b = np.ones(tensor2.shape)
-    return ShellTensor((a + b).shape)
+    if tensor1.shape != tensor2.shape:
+      raise ValueError("Tensor shapes mismatch.")
+
+    return ShellTensor(tensor1.shape)
 
   def sub(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
-    a = np.ones(tensor1.shape)
-    b = np.ones(tensor2.shape)
-    return ShellTensor((a - b).shape)
+    if tensor1.shape != tensor2.shape:
+      raise ValueError("Tensor shapes mismatch.")
+
+    return ShellTensor(tensor1.shape)
--- a/tensornetwork/backends/shell/shell_backend.py
+++ b/tensornetwork/backends/shell/shell_backend.py
@@ -255,3 +255,13 @@ class ShellBackend(base_backend.BaseBackend):
     a = np.ones(tensor1.shape)
     b = np.ones(tensor2.shape)
     return ShellTensor((a * b).shape)
+
+  def add(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
+    a = np.ones(tensor1.shape)
+    b = np.ones(tensor2.shape)
+    return ShellTensor((a + b).shape)
+
+  def sub(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
+    a = np.ones(tensor1.shape)
+    b = np.ones(tensor2.shape)
+    return ShellTensor((a - b).shape)
--- a/tensornetwork/backends/tensorflow/tensorflow_backend.py
+++ b/tensornetwork/backends/tensorflow/tensorflow_backend.py
@@ -83,7 +83,10 @@ class TensorFlowBackend(base_backend.BaseBackend):
     return self.tf.linalg.diag(tensor)
 
   def convert_to_tensor(self, tensor: Tensor) -> Tensor:
-    result = self.tf.convert_to_tensor(tensor)
+    if isinstance(tensor, (float, int)):
+      result = self.tf.convert_to_tensor(tensor, dtype=self.tf.float64)
+    else:
+      result = self.tf.convert_to_tensor(tensor)
     if self.dtype is not None and result.dtype is not self.dtype:
       raise TypeError(
           "Backend '{}' cannot convert tensor of dtype {} to dtype {}".format(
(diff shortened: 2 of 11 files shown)


jziub:
This is an implementation of addition in numpy backend. Will add other operations if this is the right direction. 

For the returned node, does it need to be a copy of the left hand node? Or just a new node with the dangling edges?

Thenerdstation:
Hm, that's a good question. Perhaps it would only make sense to support modifying the tensor in place, so we would only support `a += b` type operations.

@mganahl what are your thoughts?

mganahl:
I think typically one would expect such an operation to return a new object, so I would say `__add__`  should return a new `Node` with new edges. We can still have `__iadd__` as well, in which case everything except `Node.tensor` stays the same as it was.

jziub:
Thanks for clarifying. Will return a brand new node then.

Thenerdstation:
Thanks for the PR!

Other projects will take my priority right now, so I will review this in ~2 weeks.

Thenerdstation:
Thanks for the PR!

Other projects will take priority for me on the short term, so I will review this ~2 weeks.

jziub:
No problem. Have the addition and subtraction implemented across backends.

tensornetwork/backends/shell/shell_backend.py
```diff
@@ -255,3 +255,13 @@ def multiply(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
...
     return ShellTensor((a * b).shape)
+
+  def add(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
+    a = np.ones(tensor1.shape)
+    b = np.ones(tensor2.shape)
+    return ShellTensor((a + b).shape)
+
+  def sub(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
+    a = np.ones(tensor1.shape)
+    b = np.ones(tensor2.shape)
```
Thenerdstation: Not sure I agree with how this is implemented. This will blow up if the shapes are huge (which would be the main reason to use the shell backend in the first place).

mganahl: I agree, just check if the shapes  the tensors match and then create a ShellTensor directly

jziub: Changed as suggested. Added another unit test to check mismatch shape. Please take another look.

tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -83,7 +83,10 @@ def diag(self, tensor: Tensor) -> Tensor:
     return self.tf.linalg.diag(tensor)
 
   def convert_to_tensor(self, tensor: Tensor) -> Tensor:
-    result = self.tf.convert_to_tensor(tensor)
+    if isinstance(tensor, (float, int)):
```
mganahl: what is this case covering?

tensornetwork/network_components.py
```diff
@@ -534,6 +534,34 @@ def get_tensor(self):
   def set_tensor(self, tensor):
     self.tensor = tensor
 
+  def __add__(self, other: Union[int, float, "BaseNode"]) -> "BaseNode":
+    if not hasattr(self, '_tensor'):
+      raise AttributeError("Please provide a valid tensor for this Node.")
+
+    if isinstance(other, BaseNode):
+      other_tensor = other.get_tensor()
```
mganahl: I thought we had deprecated get_tensor()?

tensornetwork/backends/tensorflow/tensorflow_backend.py
jziub: Given a scalar, we need to specify the data type to create a tensor in tensorflow.

tensornetwork/network_components.py
jziub: Replaced with tensor.

tensornetwork/backends/tensorflow/tensorflow_backend.py
mganahl: Just worried that explicit use of `tf.float64` could lead to problems if other dtypes are e.g. tf.float32

jziub: hmm, the input is constrained to float and int in python type, could that protect tf.float64?

mganahl: Can you tell me why you added this?

jziub: self.tf.convert_to_tensor(tensor) throws an error if the input is float/int.

mganahl: OK, but where in the code do you need to pass a float or int to backed.convert_to_tensor?

jziub: Following functions call backed.convert_to_tensor.  Test case is like np.ones((1, 2, 3)) + 2

def __sub__()
def __add__()

tensornetwork/backends/tensorflow/tensorflow_backend_test.py
```diff
@@ -257,3 +257,33 @@ def test_multiply(a, b, expected):
   tensor2 = backend.convert_to_tensor(b)
 
   np.testing.assert_allclose(backend.multiply(tensor1, tensor2), expected)
+
+
+@pytest.mark.parametrize("a, b, expected",
```
mganahl: please test this for all supported dtypes

tensornetwork/tests/network_test.py
```diff
@@ -537,3 +537,33 @@ def test_remove_node(backend):
   broken_edges_by_name, broken_edges_by_axis = tn.remove_node(b)
   assert broken_edges_by_name == {"0": a[0]}
   assert broken_edges_by_axis == {0: a[0]}
+
+
+@pytest.mark.parametrize("a, b, expected",
```
mganahl: please parametrize tests over different dtypes

mganahl: also check that dtypes are preserved (no implicit conversions)

jziub: Add test cases. Please take another look.

tensornetwork/backends/tensorflow/tensorflow_backend_test.py
jziub: Add test cases. Please take another look.

alewis:
Is this still relevant given the new API?

Issue #303, opened 2019-10-07
implement sparse symmetric eignesolver for jax backend


mganahl:

MichaelMarien:
I don't mind looking at this if it's not urgent. Probably won't have time before December.

Thenerdstation:
I believe this is done.

PR #359, opened 2019-11-01
WIP: jax eigsh lanczos

MichaelMarien:
This is still very much work in progress, but would like to capture some initial feedback of @Thenerdstation , @mganahl . While taking a first look at #303 I started playing a bit with the code.

It occurs to me that with some additional backend methods, we can have the current numpy code rather generic over the different frameworks. Feedback I would like to capture is whether the inclusion of `backend.dot`and `backend.eigh`methods is worth making the implementation of `eigsh_lanczos`backend independent (or almost, I'm very close already). There are some other changes already, like refactoring out the checks, but never mind these as this is definitely not a final design.


mganahl:
Hey @MichaelMarien looks good! I like the idea of having a backend independent lanczos. The reason  we decided to put it into the backend is that we want eigsh_lanczos to work under `jax.jit`  in jax, and under `tf.function` in tensorflow. This affects some control flow ops like while and comparisons as well.

MichaelMarien:
> The reason we decided to put it into the backend is that we want eigsh_lanczos to work under `jax.jit` in jax, and under `tf.function` in tensorflow. This affects some control flow ops like while and comparisons as well.

What exactly is it we'd like to jit? Is it the (action of the) linear operator we want the eigenvectors of?

mganahl:
Hey @MichaelMarien I think we  want to jit both the liner operator as well as the lanczos iteration itself.

MichaelMarien:
implemented in another PR #570 in a jitted way

PR #362, opened 2019-11-04
remove dtype from backends

### LINKED ISSUE #349: Backends have fixed dtypes
Why is it that, currently, backends have a dtype field? Judging by how it is used in `convert_to_tensor()`, it seems to be forced on all tensors if it is set (not `None`), with exceptions raised if, e.g. I try to create a tensor from real valued inputs (`float64`) when the dtype is complex (`complex128`).

This surprised me - I could imagine it being the "default" dtype for new tensors, but I didn't expect it to be enforced like this. I expect there is a good reason?

It seems that the backend object created for a new `Node` has the dtype field set, so this is the default behavior, I guess.

mganahl:
`dtype` attribute is removed from all backends.
`BaseNode.dtype` returns `BaseNode.tensor.dtype` now
fixes #349

### DIFF
--- a/tensornetwork/backends/backend_factory.py
+++ b/tensornetwork/backends/backend_factory.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from tensornetwork.backends.tensorflow import tensorflow_backend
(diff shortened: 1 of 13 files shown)


mganahl:
@Thenerdstation do we use `switch_backend` anywhere right now?

PR #374, opened 2019-11-07
remove dtype from shell backend

mganahl:


Thenerdstation:
The fact that this passed without modifying any tests is worrisome.

PR #379, opened 2019-11-09
Quantum abstractions

amilsted:
Promised PR for the quantum abstractions. This is still WIP, but I think it's ready for some more thorough feedback!


tensornetwork/quantum/quantum.py
```diff
@@ -0,0 +1,374 @@
...
+    out_edges = [edges_dict1[e] for e in self.out_edges]
+    ref_nodes = ([n for _, n in nodes_dict1.items()] +
+                 [n for _, n in nodes_dict2.items()])
+    ignore_edges = ([edges_dict1[e] for e in self.ignore_edges] +
+                    [edges_dict2[e] for e in other.ignore_edges])
+
+    return quantum_constructor(out_edges, in_edges, ref_nodes, ignore_edges)
+
+  def __mul__(self, other: "QuOperator"):
+    if self.is_scalar or other.is_scalar:
```
Thenerdstation: Would it make sense to also support normal scalars rather than just nodes that act as scalars?

tensornetwork/quantum/quantum.py
```diff
@@ -0,0 +1,374 @@
...
+      return self @ other
+    raise ValueError("Elementwise multiplication is only supported if at "
+                     "least one of the arguments is a scalar.")
+
+  def tensor_prod(self, other: "QuOperator"):
```
Thenerdstation: in the rest of the code this is called the `outer_product`. should we do that here too?

tensornetwork/quantum/quantum.py
```diff
@@ -0,0 +1,374 @@
...
+    Args:
+      contractor: A function that performs the contraction. Defaults to
+        `greedy`, which uses the greedy algorithm from `opt_einsum` to
+        determine a contraction order.
+      final_edge_order: Manually specify the axis ordering of the final tensor.
+    Returns:
+      The final tensor representing the operator.
+    """
+    if not final_edge_order:
+      # TODO: Perhaps support using no assigned ordering at all here?
```
Thenerdstation: I think we should always force a user to give the final edge order or have them explicitly turn that off like we do with `contractors`.

tensornetwork/quantum/quantum_test.py
```diff
@@ -0,0 +1,139 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+import numpy as np
+import tensornetwork as tn
+import quantum as qu
+
+
+def test_constructor():
```
Thenerdstation: Can you also include `backend` to all of the tests?

tensornetwork/quantum/quantum.py
amilsted: In this case there is an order implied by `in_edges` and `out_edges` that I think we can safely make the default (this is the default ordering I have picked). Given row-major memory layout, it is the ordering corresponding to the Kronecker product.

tensornetwork/quantum/quantum.py
amilsted: Yes. Was thinking of supporting that!

tensornetwork/quantum/quantum.py
```diff
@@ -0,0 +1,374 @@
...
+  def is_scalar(self):
+    return len(self.out_edges) == 0 and len(self.in_edges) == 0
+
+  @property
+  def is_vector(self):
+    return len(self.out_edges) > 0 and len(self.in_edges) == 0
+
+  @property
+  def is_adjoint_vector(self):
+    return len(self.out_edges) == 0 and len(self.in_edges) > 0
```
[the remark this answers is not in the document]
amilsted: Okay, I suppose I don't mind. What's your thinking?

tensornetwork/quantum/quantum.py
amilsted: Yeah, I wondered about this too...
I chose `tensor_prod` because this is the common parlance in QM. We could have both `tensor_product` and `outer_product` mean the same thing?

tensornetwork/quantum/quantum_test.py
amilsted: Ugh, do I have to??

😉 Will do.

tensornetwork/quantum/quantum.py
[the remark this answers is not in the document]
amilsted: Also, do you mean make them methods, or really functions (`def is_vector(op):`...)? Am happy with the former, but find the latter odd, since these functions only make sense for `QuOperator`s.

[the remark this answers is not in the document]
Thenerdstation: Just remove the `@property` and make them member methods.

tensornetwork/quantum/quantum.py
```diff
@@ -264,6 +261,12 @@ def tensor_prod(self, other: "QuOperator"):
 
     return quantum_constructor(out_edges, in_edges, ref_nodes, ignore_edges)
 
+  def outer_product(self, other: "QuOperator"):
+    """Outer product with another operator.
+    Synonym for `tensor_product()`.
+    """
+    return self.tensor_product(other)
+
```
Thenerdstation: I don't think unneeded aliases are good. Just pick one, I have no preference at this point.

amilsted: Ok.

amilsted:
I was thinking about splitting out some of the functionality into separate functions. For example, one could have a `multiply_operators(out_edges_1, in_edges_1, out_edges_2, in_edges_2, **kwargs)` serve as the implementation for `QuOperator.__mul__()`. It could then be used outside of the `QuOperator` framework. Comments? @Thenerdstation @mganahl

amilsted:
Also, currently it looks like identities (which are `CopyNode`s) get expanded into dense identity tensors prior to contraction. Is there a way to avoid this? If not, I guess we should have the automatic contractors detect `CopyNode` identities and eliminate them.

amilsted:
And another thing: Do we want to consider again the pros and cons of (re-)adding an "official" and generic `TensorNetwork` container? It could be a minimal base class for `QuOperator`, but would it be useful beyond that? I guess this question boils down to: Are there operations on *generic* tensor networks that would be nicer as overloaded operators or object methods (vs. the current functional interface). Right now I can't think of any...?

mganahl:
> And another thing: Do we want to consider again the pros and cons of (re-)adding an "official" and generic `TensorNetwork` container? It could be a minimal base class for `QuOperator`, but would it be useful beyond that? I guess this question boils down to: Are there operations on _generic_ tensor networks that would be nicer as overloaded operators or object methods (vs. the current functional interface). Right now I can't think of any...?

If there's a good use case then yes, otherwise not sure. Generally I prefer the functional coding.

Thenerdstation:
> Also, currently it looks like identities (which are CopyNodes) get expanded into dense identity tensors prior to contraction. Is there a way to avoid this?

Yes, you can do `tn.contract_copy_node(copy_node)` and it will contract all of the legs on that node relatively efficiently.

Thenerdstation:
Ping on updates for this.

amilsted:
Hi @Thenerdstation! I am pretty happy with how this is looking at the moment. The only real issue (a performance issue) is that there are some obstacles in the way of automatically eliminating CopyNode identity tensors from the network. This could/should be fixed in a separate PR.

Thenerdstation:
SGTM!

amilsted:
Okay, I changed my mind and have implemented the elimination of identity CopyNodes in `quantum.py` for now.

Also, if you want typing for return values, you might want to wait a little longer :)

amilsted:
Note: I had to add a feature to the main library to make the CopyNode thing work efficiently, namely the propagation of `CopyNode`s through `tn.copy()`.

amilsted:
Ok, this is ready to go again. @Thenerdstation You may want to take a look at `eliminate_identities()` and the changes to `tn.copy()`.

Thenerdstation:
Please add a GitHub issues about the `tn.copy()` behavior.

PR #381, opened 2019-11-11
Added __repr__ for Node and Edges (#353)

### LINKED ISSUE #353: Add better __repr__ for Node class
We don't have a good repr for the `Node` class yet.

drigio:
Fixes #353 .

Please review and let me know if there are any changes required.


drigio:
Sure! I may have to test the whole string I guess then? Will I be using ``double_node_edge`` and ``single_node_edge`` from the test script?

PR #380, opened 2019-11-09
added eigh to backends

mganahl:

PR #375, opened 2019-11-07
implementation of eigs for numpy backend

mganahl:
Other backends will raise an  error if `eigs` is called

PR #382, opened 2019-11-11
add index_update method to backends

mganahl:
`index_update` allows to assign values to a tensor using a "mask":

```python 
x = backend.randn((10,))
mask = x>0.1
value = 0.0
backend.index_update(x, mask, value) #zero out all values > 0.1
```


mganahl:
I took `index_update` from jax.If you don't like it I'm happy to rename.

mganahl:
This PR already has the following PRs merged:
#375 -> #380 -> #382 
I'd recommend merging this in the above order to avoid clashes.

Thenerdstation:
Why do you need this?

mganahl:
There are cases where I need to take a pseudo inverse of a diagonal matrix (for stability reasons)
by zeroing those elements of the inverse for which the matrix elements are smaller than a cutoff.

PR #402, opened 2019-12-05
Block sparse tensors

mganahl:
This is the first step at implementing block sparsity


mganahl:
Hey @Thenerdstation, @amilsted  I merged my initial code for block-sparsity in. I'll be staying low over the next two weeks, but fell free to do with it whatever you deem useful. It's only an experimental code so we can get a feel for it!

Issue #423, opened 2020-01-17
What to submit for a issue?


IllgamhoDuck:
https://github.com/google/TensorNetwork/blob/8a4f8eb4b0a4a3e57022f63877ae903ef42cfe4c/tensornetwork/network_components.py#L314-L315

I'm trying to use the tensor network and testing it right now.
I just found that this line of code doesn't have protection.

So if I do ..
```
node = tn.Node(np.eye(2))
print(node) # __unnamed_node__
node.set_name(node)
print(node)
```
Type error occured not when I assign the name and occured when I try to print it out.
Because I sended the node instance to the self.name.
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-28-cdce2f239ef3> in <module>
----> 1 print(node)

TypeError: __str__ returned non-string (type Node)
```

Is this a problem or not? Is this kind of protection needed?

```
assert type(self.name) == str, "the name should be string type"
```
I'm not sure about is this could be an issue or not so I just posted a question here.
I tried to look for a guideline but hard to get a sense of the scope of how far I could post or fix.
https://github.com/google/TensorNetwork/blob/master/CONTRIBUTING.md

Thenerdstation:
Yeah, you could add that check. It would be in the `__str__` definition of the Nodes class. Since it's pretty straight forward, would you be interested in trying to do a PR for this?

IllgamhoDuck:
@Thenerdstation Yes for sure. It's just one line of code so not really hard. Anyway, I'm digging on the codebase so for further I'll try to look at the protections and etc. I'm currently looking at the https://github.com/google/TensorNetwork/issues/292(algebraic operations for BaseNode).

PR #424, opened 2020-01-17
BaseNode / Edge class name type check protection add

IllgamhoDuck:
Fix https://github.com/google/TensorNetwork/issues/423

**BaseNode class** - Add protection to `name, axis_names`

Protected in 3 place
- Initialize stage - `__init__`
- Function use setting - `set_name / add_axis_names`
- Property - Add `@property` to `name` to protect direct adding `node.name = 123`
___
**Edge class** - Add protection to `name`

Protected in 3 place
- Initialize stage - `__init__`
- Function use setting - `set_name`
- Property


IllgamhoDuck:
@googlebot I signed it!

tensornetwork/network_components.py
```diff
@@ -122,9 +130,12 @@ def add_axis_names(self, axis_names: List[Text]) -> None:
     if len(axis_names) != len(set(axis_names)):
       raise ValueError("Not all axis names are unique.")
     if len(axis_names) != len(self.shape):
-      raise ValueError("axis_names is not the same length as the tensor shape."
-                       "axis_names length: {}, tensor.shape length: {}".format(
+      raise ValueError("Axis_names is not the same length as the tensor shape."
```
Thenerdstation: Don't worry about capitalization here since `axis_names` is a variable.

tensornetwork/network_components.py
```diff
@@ -373,6 +386,16 @@ def edges(self, edges: List) -> None:
                            self.name))
     self._edges = edges
 
+  @property
+  def name(self) -> None:
```
Thenerdstation: Return type should be `Text`

IllgamhoDuck: my bad resolved it!

MichaelMarien:
@IllgamhoDuck thanks a lot for the effort, looks great. In general, when type checking using the 'isinstance' method is prefered over 'type(...)==...'. If I remember correctly, it's faster and more importantly it supports subclassing. Unless you really don't want that, you should use the former (also, the linter agrees with me ;)).

More seriously @Thenerdstation . It seems that the torch=1.4.0 release broke the build, nothing to do with your changes. We could confirm it by pinning the version to 1.3.1. I don't have too much time right now, I might look into it tomorrow evening CET.

Thenerdstation:
Aw, good catch Michael on both fronts!

I'll investigate the pytorch issues. For now, let's just have the `requirements_travis.txt` force pytorch to be 1.3.1

IllgamhoDuck:
@MichaelMarien  thanks for informing! I'll fix it to `isinstance` and re push it

Thenerdstation:
Tests pass so I'll merge this once that linter issue is fixed.

MichaelMarien:
@IllgamhoDuck , build for python3.6 failed, but It might be just a connection issue. Can you retrigger the build by pushing a commit (some documentation or so?)

Thenerdstation:
Don't worry, I restarted the build. For some reason, travis breaks like this sometimes, but restarting it usually fixes it.

MichaelMarien:
> Don't worry, I restarted the build. For some reason, travis breaks like this sometimes, but restarting it usually fixes it.

Admin power ftw!

Issue #350, opened 2019-10-25
Making a vector from a list of scalars: Add stack()?


amilsted:
I want to take a list of scalar values and make a vector of it using backend operations. This seems to be tricky at the moment! My initial attempt was:
```python
scalars = [backend.convert_to_tensor(1.0), backend.convert_to_tensor(2.0)]
vec = backend.concat(scalars, axis=0)
```
This fails, since I cannot concatenate zero-dimensional arrays.

Perhaps we should have a backend.stack()?

Thenerdstation:
Yes, backend.concat is ONLY for shape calculations (we really should change the name).

I'd prefer to change backend.concat to backend.shape_concat.

PR #425, opened 2020-01-17
#350 make vec from scalars

Travmatth:
Addresses issue #350, renames existing `backend.concat` methods to `backend.shape_concat`  and implement `backend.concat` methods using backend `stack` method. If name `backend.shape_concat` is too similar to existing `backend.concat_shape` methods could alternatively fold `backend.concat` into existing method of same name as special case when `axis == 0`.  Did not squash commits.


tensornetwork/backends/base_backend.py
```diff
@@ -135,7 +135,12 @@ def rq_decomposition(
     raise NotImplementedError(
         "Backend '{}' has not implemented rq_decomposition.".format(self.name))
 
-  def concat(self, values: Sequence[Tensor], axis) -> Tensor:
+  def shape_concat(self, values: Sequence[Tensor], axis) -> Tensor:
+    """Concatenate a sequence of tensors together about the given axis."""
```
Thenerdstation: Can you change the description here to be explicitly only for shape calculations?

Travmatth: Added to the description, now read: 
```
"""Concatenate a sequence of tensors together about the given axis,
intended only for use in shape calculations"""
```

Thenerdstation:
Hmm. Still seems like there's a shape based error for the Pytorch backend. Can you investigate?

tensornetwork/backends/pytorch/pytorch_backend.py
```diff
@@ -69,9 +69,12 @@ def rq_decomposition(
   ) -> Tuple[Tensor, Tensor]:
     return decompositions.rq_decomposition(self.torch, tensor, split_axis)
 
-  def concat(self, values: Tensor, axis: int) -> Tensor:
+  def shape_concat(self, values: Tensor, axis: int) -> Tensor:
     return np.concatenate(values, axis)
 
+  def concat(self, values: Tensor, axis: int = 0) -> Tensor:
+    return np.stack(values, axis)
```
Thenerdstation: Aw, that would explain it. You'll need to use pytorch's concat equivalent here instead of numpy.

tensornetwork/backends/shell/shell_backend.py
```diff
@@ -107,14 +107,22 @@ def rq_decomposition(self, tensor: Tensor,
...
     return ShellTensor(new_shape)
 
+  def concat(self, values: Sequence[Tensor], axis: int = 0) -> Tensor:
+    new_shape = None
+    if axis == 0:
+      new_shape = ShellTensor(values)
+    else:
+      new_shape = self.shape_concat(values, axis)
+    return new_shape
+
```
Thenerdstation: I don't this this is correct. `ShellTensor` is a tensor type that only stores its shape and has no concrete values, so when you do a normal concatenation of multiple `ShellTensors`, you'll want to add the value of that specified axis together.

Though this function really isn't important at all. Just throw a `NotImplementedError` and we'll add it when we actually need it.

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -39,9 +39,12 @@ def convert_to_tensor(self, tensor: Tensor) -> Tensor:
     result = self.jax.jit(lambda x: x)(tensor)
     return result
 
-  def concat(self, values: Tensor, axis: int) -> Tensor:
+  def shape_concat(self, values: Tensor, axis: int) -> Tensor:
```
Thenerdstation: Doubly come to think of it, we don't need the axis number for `shape_concat` at all since we always use just -1. Can you remove this argument?

Travmatth:
It looks as though the Travis build for python 3.7 failed due to an error when downloading pytype. The Travis build for python 3.6 successfully completed.

PR #428, opened 2020-01-18
Fix default axis names in _flatten_trace_edges

jacksonwb:
Re-create the axis_names for the new edge configuration after all node edges are flattened in _flatten_trace_edges

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -1171,6 +1171,7 @@ def _flatten_trace_edges(edges: List[Edge],
   edge2 = Edge(node1=node, axis1=len(perm_front) + 1, name="TraceBack")
   node.edges = node.edges[:len(perm_front)] + [edge1, edge2]
   new_edge = connect(edge1, edge2, new_edge_name)
+  node.axis_names = [str(n) for n in range(len(node.edges))]
   # pylint: disable=expression-not-assigned
   [edge.disable() for edge in edges]  #disable edges!
   return new_edge
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -1155,6 +1155,8 @@ def _flatten_trace_edges(edges: List[Edge],
     The new edge that represents the flattening of the given edges.
   """
   node = edges[0].node1  # We are in the trace case, so this is the only node.
+  has_nongeneric_names = (set(node.axis_names) != 
+                          {str(n) for n in range(len(node.edges))})
   backend = node.backend
   # Flatten all of the edge's axes into a a single list.
   perm_back = [min(e.axis1, e.axis2) for e in edges]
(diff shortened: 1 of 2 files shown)


jacksonwb:
@googlebot I fixed it.

tensornetwork/network_components.py
```diff
@@ -1171,6 +1171,7 @@ def _flatten_trace_edges(edges: List[Edge],
   edge2 = Edge(node1=node, axis1=len(perm_front) + 1, name="TraceBack")
   node.edges = node.edges[:len(perm_front)] + [edge1, edge2]
   new_edge = connect(edge1, edge2, new_edge_name)
+  node.axis_names = [str(n) for n in range(len(node.edges))]
```
Thenerdstation: Hmm, I'm not sure this is what we actually want here. Edge names can sometimes be different from axis names. I think what we want is for all of the axes to keep their original name except for the ones that got flattened. And then for that name maybe go with a default name?

@amilsted @mganahl do you two have any opinions here?

jacksonwb: Yeah I see what you mean. We were ending up with nodes that had an `axis_names` list of a different size than the number of edges in the node which was breaking tests in my other PR. I thought about doing something similar to `flatten_edges`:

```
if axis_names:
      node.axis_names = [axis_names[n] for n in range(len(node.edges))]
    else:
      node.axis_names = [str(n) for n in range(len(node.edges))]
```

But that also seems to be not quite sufficient as it is only selecting the first n names instead of the correct n names. Also wasn't quite sure how we should determine the new axis name so I went with this solution to at least have the correct number of names.

Thenerdstation: I think a better strategy would be to make a copy of the names in a new list and then remove all of the names in the indices of `edge.axis1` and `edge.axis2` for all of the given edges. You can see an example of the multi remove here. https://github.com/google/TensorNetwork/blob/e213de9ceae4a4003469cb65f78879bcb9e02514/tensornetwork/contractors/opt_einsum_paths/utils.py#L23

tensornetwork/network_components.py
```diff
@@ -1171,7 +1173,12 @@ def _flatten_trace_edges(edges: List[Edge],
   edge2 = Edge(node1=node, axis1=len(perm_front) + 1, name="TraceBack")
   node.edges = node.edges[:len(perm_front)] + [edge1, edge2]
   new_edge = connect(edge1, edge2, new_edge_name)
-  node.axis_names = [str(n) for n in range(len(node.edges))]
+  if has_nongeneric_names:
+    node.axis_names = (node.axis_names[0:len(perm_front)] + 
+                       [node.axis_names[new_edge.axis1], 
+                        node.axis_names[new_edge.axis2]])
+  else:
+    node.axis_names = [str(n) for n in range(len(node.edges))]
```
jacksonwb: Here the axis names will be preserved if they have non-generic names. Grab non flattened axis names and then the axes of the first edges amongst the edges being flattened.

If the names are generic a new set of names will be generated instead of just using the generic names that have been scrambled by _flatten_trace_edges. But I can remove this if you think its better to not regenerate the generic names.

Thenerdstation:
Let me first discuss with my other collaborators for what we think should be the correct result here. Your code is good though!

Issue #427, opened 2020-01-17
Pytorch 1.4.0 breaks our build


Thenerdstation:
Needs to be investigated.

IllgamhoDuck:
I'll try to look up for this

MichaelMarien:
hi @IllgamhoDuck , not to stress you out, but will you get a chance to look at this in the near future? Would like to fix this myself otherwise, as otherwise these kind of things tend to get worse :).

IllgamhoDuck:
Hi @MichaelMarien! Sure I would love to dig on it :) I'll try to look at this problem thanks for asking!

PR #429, opened 2020-01-18
Changing requirements to accept higher versions

kshithijiyer:


MichaelMarien:
Let's wait until we fix #427

Thenerdstation:
Yeah, I don't think we'll be doing this for a while, so I'll just close this.

kshithijiyer:
Is it compatible with a lower version as well? If not then would consider to merge it after the #427 is fixed. Else I am good with closing this.

PR #430, opened 2020-01-18
Remove duplicate Dockerfile

kshithijiyer:
The same file is present in the root directory of the project.


Thenerdstation:
Can you delete the one at root instead?

kshithijiyer:
Done #431

Issue #203, opened 2019-08-21
Merge axes names after contraction


Thenerdstation:
As discussed in https://github.com/google/TensorNetwork/issues/179, we should merge axes names after contraction and throw an error if the user tries to access an ambiguous name.

MichaelMarien:
I can sure give this a try! One thing I'm immediately wondering is whether we should exploit a node's `name` to introduce uniqueness. I tend to reuse a name like "physical_index" for multiple MPS nodes, but I would give the nodes different names.

For instance, a Node `A`with axis names `[a, b, c]`and another node `B` with the same axis names. After contracting the `a` axis we can return a Node with axis names `[A_b, A_c, B_b, B_c]` ... What do you think?

Thenerdstation:
That might get messy once you start contracting multiple nodes together. Also, unless you specify a node name, the new node created after contraction will have a default name "__Node_X".

MichaelMarien:
Yes, you are right. I always specify node names, but unless we make this a required property (which we should not I think), we can't assume the user will do this and than it can become a complete jungle very quickly.

MichaelMarien:
Should we throw an error immediately when a contraction is done that leaves unambiguous axis, or delay this as long as possible (only for instance when user makes explicit reference to such axis name?)

Thenerdstation:
I think when the user does it explicitly.

MichaelMarien:
Currently we are throwing that error at Node creation time (see `add_axis_names` method)? Should we change this also and only throw errors when axis is requested not when names are added? Or treat nodes added and nodes created from contraction differently. The latter feels awkward to me.

MichaelMarien:
@Thenerdstation You want it when user does it explicitly? See my remake above :)

Thenerdstation:
Sorry for the late response.

I see what you mean by the latter feeling awkward, but right now I think it's the best compromise we can do. It makes sense to throw an error when a user writes it, as it can be easily fixed, but it's the best compromise we could come up with when the axis names collide after contraction.

jacksonwb:
[...]

PR #432, opened 2020-01-19
Add support for merging axis names after contraction(#203)

jacksonwb:
NOTE - This requires PR #428 to not fail tests  

* Add _merge_axis_names which reduces node name lists and excluded axes
  lists.
* _unsafe_add_axis_names was added to allow attaching axis_names lists
  with ambiguous names
* get_axis_number will throw a Value Error on access to an ambiguous
  axis name
* Add axis name merging to contract, contract_between, contract_trace,
  outer_product
* If one node in the contract has non-generic names, axis names will be
  merged, otherwise generic names will be generated
* Tests added in axis_names_test

PR #433, opened 2020-01-19
Add support for merging axis names after contraction(#203)

jacksonwb:
* Add _merge_axis_names which reduces node name lists and excluded axes
  lists.
* _unsafe_add_axis_names was added to allow attaching axis_names lists
  with ambiguous names
* get_axis_number will throw a Value Error on access to an ambiguous
  axis name
* Add axis name merging to contract, contract_between, contract_trace,
  outer_product
* If one node in the contract has non-generic names, axis names will be
  merged, otherwise generic names will be generated
* Tests added in axis_names_test

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -1911,9 +1911,7 @@ def contract_between(
     # Merge and update axis names if nodes use non-generic names
     if (not axis_names and 
         (node1.has_nongeneric_axis_names() or 
-         node2.has_nongeneric_axis_names()) and
-        backend.name in ['tensorflow', 'jax', 'numpy', 
-                         'pytorch']):
+         node2.has_nongeneric_axis_names())):
       new_node._unsafe_add_axis_names(
           _merge_axis_names(node1, node2, axes1, axes2))
 
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -109,23 +109,37 @@ class BaseNode(ABC):
     """
     self.signature = signature
 
-  def add_axis_names(self, axis_names: List[Text]) -> None:
-    """Add axis names to a Node.
+  def _unsafe_add_axis_names(self, axis_names: List[Text]) -> None:
+    """Add axis names to a Node without checking uniqueness.
+    
+    Used to attach merged axis names to a new node generated
+    from a contraction
 
     Args:
       axis_names: List of names for each of the tensor's axes.
 
     Raises:
-      ValueError: If there is a repeated name in `axis_names` or if the length
-        doesn't match the shape of the tensor.
+      ValueError: If the length doesn't match the shape of the tensor.
     """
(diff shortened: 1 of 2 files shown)


jacksonwb:
PR #432 required to pass tests

tensornetwork/network_components.py
```diff
@@ -1743,6 +1776,38 @@ def disconnect(edge,
   return edge.disconnect(edge1_name, edge2_name)
 
 
+def _merge_axis_names(
+    node1: BaseNode, 
+    node2: BaseNode, 
+    axes1: List[Text], 
+    axes2: List[Text]) -> None:
```
Thenerdstation: I think these should be `int`s based on how you use them below.

tensornetwork/network_components.py
```diff
@@ -1843,6 +1908,15 @@ def contract_between(
     new_tensor = backend.tensordot(node1.tensor, node2.tensor, [axes1, axes2])
     new_node = Node(
         tensor=new_tensor, name=name, backend=backend)
+    # Merge and update axis names if nodes use non-generic names
+    if (not axis_names and 
+        (node1.has_nongeneric_axis_names() or 
+         node2.has_nongeneric_axis_names()) and
+        backend.name in ['tensorflow', 'jax', 'numpy', 
+                         'pytorch']):
```
Thenerdstation: Why are you checking backend names here?

tensornetwork/network_components.py
```diff
@@ -1926,6 +2000,9 @@ def outer_product(node1: BaseNode,
   new_node = Node(
       tensor=new_tensor, name=name, axis_names=axis_names, backend=backend)
   additional_axes = len(node1.tensor.shape)
+  if not axis_names and \
```
Thenerdstation: Prefer to wrap things in `()` rather than using the `\`

tensornetwork/tests/axis_names_test.py
```diff
@@ -0,0 +1,121 @@
+import tensornetwork as tn
```
Thenerdstation: Again, awesome testing.

tensornetwork/network_components.py
jacksonwb: ah right, you are correct

tensornetwork/network_components.py
jacksonwb: Yeah I meant to run it by you, I noticed one of the tests failing, specifically `test_custom_backend` in `tensornetwork_test.py`. These axis merging rules will only work on backends implementing normal tensor behavior, and may fail on custom backends, so this was my solution to that failing test. Should I omit this, or is there a better way to check for that?

jacksonwb: OK yeah was misunderstanding something, fixed now

ronf6co:
I wonder when it will be reviewed, this is critical

Issue #339, opened 2019-10-22
Add `with tn.DefaultBackend("backend_name"):` support


Thenerdstation:
Currently, the only way to set a backend for a `Node` is either manually set it for each node or globally with `set_defualt_backend`. It would be nice to have a context manager instead, as it would be cleaner.

mchav:
Will work on this.

Thenerdstation:
Great! Go for it

mchav:
Discussed this offline and I'm not sure a context manager would be the way to do this. Especially since by design they are typically meant to manage resources. In this case we want to capture common state which is what a context manager does but only as a bi-product of handling a resource.

I think we should instead have a context object that holds any "global" state specific to node creation.

Thenerdstation:
In situations like the one above where classes that use/extend the library, having an easy way to make sure all nodes created have the same backend (vs pass `self.backend.name` to each created `Node`) would be preferable.

mchav:
Agreed. But I think:

```python
    # Optional context object.
    context = tn.NodeContext(self.backend.name)
    a = context.Node([1, 2])
    b = context.Node([3, 4])
    ...
```

Is more idiomatic than:

```python
    a = None
    b = None
    with tn.DefaultBackend("backend_name"):
        a = tn.Node([1, 2])
        b = tn.Node([1, 2])
    ...
```

For a couple of reasons.

* Context managers are usually used to [manage resources](https://book.pythontips.com/en/latest/context_managers.html) and this situation has no resource that needs to be managed or is in contention.
* The design seems brittle. For example suppose a new parameter, similar to backend, was added to the node class. You'd need to nest contexts to capture this new information.

```python
    a = None
    b = None
    with tn.DefaultBackend("backend_name"):
        with tn.OtherGlobal("param"):
            a = tn.Node([1, 2])
            b = tn.Node([1, 2])
    ...
```

So it seems this situation requires an auxillary object to store the information. I think the TensorNetwork object provided that but was somewhat misnamed and didn't have to be compulsory.

Thenerdstation:
#434

PR #434, opened 2020-01-20
Issue #339. with tn.DefaultBackend(backend): support

katolikyan:
I tried to implement this one. I believe the DefaultBackend class could be better but still :)

I also have tried to implement it in another way, more like a temporary change for `config.default_backend` in a `with` statement. In this way, it doesn't affect any code inside the `class Node`, but if someone will set `config.default_backend` inside a `with` statement that will cause a little mess. Here is the main idea how would another approach look like:

```python3
from typing import Text, Union
from contextlib import contextmanager
from tensornetwork.backends.base_backend import BaseBackend
import tensornetwork.config as config
from tensornetwork.__init__ import set_default_backend

@contextmanager
def DefaultBackend(backend: Union[Text, BaseBackend]) -> None:
    try:
        default_backend = config.default_backend
        set_default_backend(backend)
        yield None
    finally:
        set_default_backend(default_backend)
```


katolikyan:
@googlebot I signed it!

katolikyan:
@googlebot I fixed it.

tensornetwork/backend_contextmanager.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(self, backend: Union[Text, BaseBackend]) -> None:
+        DefaultBackend.backend = backend
+
+    def __enter__(self):
+        DefaultBackend.with_statement = True
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        DefaultBackend.with_statement = False
+        DefaultBackend.backend = None
```
Thenerdstation: Nice! I like the object oriented approach.

However, I think we should be using a stack based approach for finding the current backend. (For an example, see what we did with the `NodeCollection` [code](https://github.com/google/TensorNetwork/blob/83ff4af11dcd03d6c0db70e139c7137154dc015d/tensornetwork/network_components.py#L1978).

tensornetwork/tests/backend_contextmanager_test.py
```diff
@@ -0,0 +1,32 @@
...
+    with tn.DefaultBackend("numpy"):
+
+        assert tn.config.default_backend == "pytorch"
+
+        a = tn.Node(np.ones((10,)))
+        tn.set_default_backend("tensorflow")
+        b = tn.Node(np.ones((10,)))
+
+    assert (type(a.backend) == type(b.backend)) == True
+    assert tn.config.default_backend == "tensorflow"
```
Thenerdstation: We should also consider the case where we have nested `with` statements. You can imagine this happening for when a library needs to implement something purely in numpy or when a user wants to test their code in all backends.

Maybe add a test like this?

```python
with tn.DefaultBackend("tensorflow"):
  a = tn.Node(...)
  assert a.backend.name == "tensorflow" 
  with tn.DefaultBackend("numpy"):
     b = tn.Node(...)
     assert b.backend.name == "numpy"
  c = tn.Node(...)
  assert c.backend.name == "tensorflow"
```

katolikyan:
Thank you for the comments! I tried to implement it with a stack as you showed in the example.
I also have added a couple of tests.
But I am not sure if I should have written the `DefaultBackend` class inside the `network_components.py` and put the stack into the `ops.py` file.

tensornetwork/tests/backend_contextmanager_test.py
```diff
@@ -0,0 +1,53 @@
...
+  assert final_node.tensor == 10.0
+
+
+def test_contextmanager_interaption():
+  tn.set_default_backend("pytorch")
+  assert tn.config.default_backend == "pytorch"
+  with tn.DefaultBackend("numpy"):
+    assert tn.config.default_backend == "pytorch"
+    a = tn.Node(np.ones((10,)))
+    tn.set_default_backend("tensorflow")
```
Thenerdstation: using `set_default_backend` while there is a backend stack feels like a foot gun. Can you raise an error in this case instead?

tensornetwork/network_components.py
```diff
@@ -516,7 +517,11 @@ def __init__(self,
       backend = tensor.backend
       tensor = tensor.tensor
     if not backend:
-      backend = config.default_backend
+      default_backend = _default_backend_stack.get_current_backend()
+      if default_backend:
+        backend = default_backend.backend
+      else:
+        backend = config.default_backend
```
Thenerdstation: Can this logic be moved to the `DefaultBackend` code?

tensornetwork/backend_contextmanager.py
```diff
@@ -0,0 +1,28 @@
...
+    _default_backend_stack.stack.pop()
+
+class _DefaultBackendStack():
+  """A stack to keep track default backends context manager"""
+
+  def __init__(self):
+    self.stack = []
+
+  def get_current_backend(self):
+    return self.stack[-1] if self.stack else None
```
Thenerdstation: Move the logic in network_components to here.

tensornetwork/tests/backend_contextmanager_test.py
katolikyan: Sorry I didn't get what you mean by raising an error in this case. Prohibit using `set_default_backend` while in `with` statement?

katolikyan:
I moved the logic to the right place. and cleaned up a test a little to get rid of some unnecessary checks.

tensornetwork/tests/backend_contextmanager_test.py
```diff
@@ -0,0 +1,42 @@
...
+      assert b.backend.name == "numpy"
+    c = tn.Node(np.ones((10,)))
+    assert c.backend.name == "tensorflow"
+  d = tn.Node(np.ones((10,)))
+  assert d.backend.name == "numpy"
+
+def test_contextmanager_wrong_item():
+  a = tn.Node(np.ones((10,)))
+  try:
+    with tn.DefaultBackend(a):
```
Thenerdstation: Add `# pytype: disable=wrong-arg-types` to this line to pass the build.

tensornetwork/tests/backend_contextmanager_test.py
```diff
@@ -0,0 +1,42 @@
...
+      pass
+    assert False
+  except ValueError:
+    assert True
```
Thenerdstation: You should use `with pytest.raises(ValueError)` instead.

https://docs.pytest.org/en/latest/assert.html

tensornetwork/backend_contextmanager.py
```diff
@@ -0,0 +1,32 @@
+from typing import Text, Union
+from tensornetwork.backends.base_backend import BaseBackend
+import tensornetwork.config as config
```
Thenerdstation: Can we move the `set_default_backend` and everything in `config` to this file? No need to make these separate anymore and I like this filename more.

tensornetwork/__init__.py
```diff
@@ -15,6 +15,7 @@
 from tensornetwork.utils import load_nodes, save_nodes
 from tensornetwork.matrixproductstates.finite_mps import FiniteMPS
 from tensornetwork.matrixproductstates.infinite_mps import InfiniteMPS
+from tensornetwork.backend_contextmanager import DefaultBackend
 
 
 def set_default_backend(backend: Union[Text, BaseBackend]) -> None:
```
Thenerdstation: Add a check that if the default backend stack is empty, and if it's not, raise an `AssertionError`. 

This is to help avoid the confusing situation like: 

```python
with tn.DefaultBackend("tensorflow"):
  tn.set_default_backend("jax")
  a = tn.Node(...)
```
Should `a` be using jax or tensorflow? It's confusing and kinda a footgun, so let's just crash with a clear message instead.

katolikyan:
Hopefully, I haven't messed up the code. Removed `config` imports with the thought that it became useless in this case. But feel like maybe I shouldn't have done that.

Also, from my understanding, we don't need direct access to `default_backend` anymore, do we?
I moved it to the `_DefaultBackendStack` class, trying to avoid using `global`. I can quickly rewrite it as a global with something like this if needed :)

```python3
default_backend = "numpy"

def set_default_backend(backend: Union[Text, BaseBackend]) -> None:
  global default_backend
  if _default_backend_stack.stack:
    raise AssertionError("The default backend should not be changed "
                         "inside the backend context manager")
  if not isinstance(backend, (Text, BaseBackend)):
    raise ValueError("Item passed to set_default_backend "
                     "must be Text or BaseBackend")
  default_backend = backend
```

Thenerdstation:
Looks awesome! Thank you so much! I've been waiting on this one for a while. ^-^

katolikyan:
Glad to hear this :) it was nice to work on that issue, thank you! :)

PR #439, opened 2020-01-21
Algebraic operation add( + ), sub( - ), mul( * ), div( / ) for BaseNode class

### LINKED ISSUE #292: algebraic operations for BaseNode

IllgamhoDuck:
Fix #292

**[BaseNode class]** - add / sub / mul / truediv NotImplemented function Added

     [Node class] - add / sub / mul / truediv function added
     [CopyNode class] - overload the BaseNode add / sub / mul / truediv as NotImplemented

**[basebackend class]** - add / sub / mul / div NotImplemented function added

     [numpy / tensorflow / pytorch] - add / sub / mul / div function added
     [shell] - add / sub / div NotImplemented function added

Testing files
**[network_components_free_test]**

     Exception - Tensorflow is not tested when the operand is scalar
        1. Check add / sub / mul / div with int / float / Node
        2. Check implicit conversion
        3. Check the Type Error when type is not int / float / Node
        4. Check is the operand backend same
        5. Check is BaseNode has attribute _tensor

**[backend_test - numpy / tensorflow / pytorch]**

        check add / sub / mul / divide work for int / float / Node

TensorFlow doesn’t accept scalar value because `scalar value dtype` will be `32 bits` and `matrix` will be `64 bits`. There is no way to change the tensor’s dtype.

**Here is my suggestions while I think about doing this.**

1. **Add dtype attribute to BaseNode to handle TensorFlow**. 
TensorFlow treats the array as 64 bit and scalar as 32 bit. So we need to modify the BaseNode to accept the datatype.

2. **BaseNode tensor type check** 
There is no type check at the tensor itself so below throws no error.

tensor assigned for `str` type

    node = tn.Node(tensor='str')

But didn’t put protection checking the tensor type at the operator part. Because this responsibility is at tensor itself in my opinion. It will be good to put the protection when we put a value at the tensor. 

    node.tensor = value


tensornetwork/tests/network_components_free_test.py
```diff
@@ -316,6 +316,292 @@ def test_node_magic_matmul(backend):
...
+    np.testing.assert_almost_equal(result, expected)
+    assert float_node.dtype == result.dtype
+    expected = np.array([[1.5, 0.5], [-0.5, -1.5]])
+    result = (float_node - node1).tensor
+    np.testing.assert_almost_equal(result, expected)
+    assert float_node.dtype == result.dtype
+
+
+def test_node_and_scalar_sub_op(backend):
+  if backend != 'tensorflow':
```
Thenerdstation: Interesting, Why are you always excluding the TF backend? Are you testing with TF1 or TF2?

IllgamhoDuck: I'm testing on **TF2.1.0**


I posted the reason why I did it at the upper comment. TF doesn't accept algebraic between different dtype like `int32` with `int64` or `int64` with `float64` different from numpy or PyTorch. **So if I put a scalar value on the Tensor the TF will throw an error.**

    tensor(2) <- int32
    tensor(2.) <- float32
    tensor([2,2]) <- int64
    tensor([2.,2.]) <- float64

So I had to exclude the test that is held on different dtype in the case of TF. If you see in the test cases I allow TF to test when the dtype is the same.


I thought about making a test case especially for TF that raise an error with different dtype calculation but I think rather than that it would be good to support BaseNode to accept the dtype as a attribute :) So we can specify the TF tensor dtype so can handle the calculation with scalar type.

Or, can make a special `implicit dtype conversion for TF tensor` to **TF backend** that check the dtypes or operands and automatically make a conversion to make different dtype same.

    int32, int64 -> int64, int64
    int32, float64 -> float64, float64

In this case, I could get off that test case that TF excluding part.


If I am misunderstanding something please tell me :)

MichaelMarien: Hey @IllgamhoDuck, first of all, thanks for the effort.
If a test needs to be skipped, can you use `pytest.skip`with an appropriate reason, it makes it easier for later follow up and overall housekeeping?

Thenerdstation: Even easier would be to just wrap all of your scalars in a `np.asarray`. I prefer that over skipping tests.

IllgamhoDuck: @MichaelMarien Thks for the advice! I will next time :) I was missing the important point! 

@Thenerdstation Got it 👍 I will change the `dtype` with `np.asarray` for TF but will leave other backends just as it is right now.

IllgamhoDuck:
Each backend has different behavior.

|  <center>Backend </center> |  <center>different dtype calculation </center> |  <center>dtype implicit conversion</center> | <center>int / int -> float</center> |
|:--------|:--------:|:--------:|:--------:|
| **Numpy** | <center>O </center> | <center>X </center> | <center>O </center> |
| **Pytorch** | <center>O </center> | <center>X </center> | <center>X (int) </center> |
| **Tensorflow** | <center>X </center> | <center>X </center> | <center>O </center> |
| **Jax** | <center>O </center> | <center>O (32->64) </center> | <center>O </center> |

And in case of `int / int` it gives `float64` except Pytorch. Pytorch maintain the dtype by rounding the result.

So if the test case seems too messy then I will split the test cases for each backend. Or just simply add a `preprocessing input` / `postprocessing output` functions for each backend operator function to get rid of this problem. So the user using could use TN library without thinking too much about the different acts between backends.

PR #440, opened 2020-01-21
Return empty dict for empty sequence input to MPS left_envs and right_envs

AidanGG:
The `min` and `max` calls prevent input of empty sequences by raising `ValueError`, so this PR just enables these cases.

### DIFF
--- a/tensornetwork/matrixproductstates/finite_mps.py
+++ b/tensornetwork/matrixproductstates/finite_mps.py
@@ -29,25 +29,25 @@ Tensor = Any
 
 class FiniteMPS(BaseMPS):
   """
-  An MPS class for finite systems. 
+  An MPS class for finite systems.
 
   MPS tensors are stored as a list of `Node` objects in the `FiniteMPS.nodes`
   attribute.
-  `FiniteMPS` has a central site, also called orthogonality center. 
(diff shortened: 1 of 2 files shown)


AidanGG:
`left_envs` and `right_envs` do not have unit tests, so would you like me to write tests for their numerical correctness as well?

Thenerdstation:
Sure! You can do that in a later PR if you like. I'll never say no to more testing.

Thenerdstation:
I'll pull this in once the test is added (you can do the numerics one later).

AidanGG:
Okay, should be done now. I'll take a look at a numerics test later.

MichaelMarien:
Looks good to me, pytype just complains that you cannot call `left_envs`with boolean type. If you just change the argument in your test `assert mps.left_envs(False) == {}`from `False`to `[]`, it should work.

@Thenerdstation, what about turning of the linter and type checker for the test files?

Thenerdstation:
I still like the linter on test files as it helps me catch dumb errors sometimes. If you have a bunch of typing tests and you want to ignore all of them then use `#pytest: skip-file` at the top of the file.

AidanGG:
I added the cases for `[]`, `()` and `range(0)` to be thorough.

PR #442, opened 2020-01-21
Method slice_edge is added

### LINKED ISSUE #260: Add support for slicing an edge
Slicing an edge with something like

```python
net.slice_edge(edge, start_index, length)
```
Would be very useful for doing parallel contractions using JAX's `pmap`. Each slice of the edge would run on a different device and the final result would just be the sum of all of the slices, which is very easy to do in jax with `psum`.

recluse27:
Fixes #260

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -1975,6 +1975,36 @@ def outer_product(node1: BaseNode,
   return new_node
 
 
+def slice_edge(edge: Edge,
+               start: int,
+               length: int):
+  """Slice an `Edge`
+  """
+  backends = [node.backend for node in edge.get_nodes() if node is not None]
+  if not all([b.name == backends[0].name for b in backends]):
+    raise ValueError("Not all backends are the same.")
+  backend = backends[0]
+
+  expected_nodes = set(edge.get_nodes())
+
+  # dangling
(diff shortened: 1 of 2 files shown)


recluse27:
@googlebot I signed it!

tensornetwork/network_components.py
```diff
@@ -1977,6 +1977,36 @@ def outer_product(node1: BaseNode,
...
+
+    edge.node1.tensor = edge.node1.tensor[..., start:start + length]
+  
+  # standard and trace
+  if (len(expected_nodes) > 1):
+    if edge.node1.shape[-1] < start or edge.node2.shape[0] < start:
+      raise ValueError("Start index is out of the range")
+
+    edge.node1.tensor = edge.node1.tensor[..., start:start + length]
+    edge.node2.tensor = edge.node2.tensor[start:start + length, ...]
```
Thenerdstation: You'll need to add `slice` as a method to the `BaseBackend` and then implement it independently for each backend.

Also, it looks like you're always only slicing the last and first axes of the tensor. You'll need to slice based on the axis given by `edge.axis1` and `edge.axis2`. (for `node1` and `node2` respectively. 

Also be sure to add test cases!

PR #444, opened 2020-01-22
More Test!

MichaelMarien:
Gotta love them


MichaelMarien:
@Thenerdstation a lot of the tests I added check if errors are raised if wrong input is provided. Of course now pytype hates me :(. Can I disable it for this one file, or tests in general? Just encountered the same thing in #440 , although there it is easily fixed

Thenerdstation:
> Can I disable it for this one file, or tests in general? 

The preferred way of doing that would be to use `# pytype: disable=<SOME_ERROR>`
See https://google.github.io/pytype/user_guide.html

MichaelMarien:
So line-by-line is preferred, ok, no problem, was a bit in a lazy mood :p

Thenerdstation:
If you wanna be slightly lazy you can use `# pytype: skip-file` at the top of a test file. We already have this for the jax backend since it makes pytype explode.

Issue #344, opened 2019-10-24
Add `tn.copy_node`


Thenerdstation:
The `tn.copy` operation is a little API heavy for just copying a single node. For example, the shortest code you can write to copy a single node looks like:

```python
a = tn.Node(...)
a_copy = tn.copy(a)[0][a]
``` 

The `[0][a]` can be kinda messy, so it would be nice to have just a `tn.copy_node` operation to make this more obvious.

MichaelMarien:
Perhaps `tn.copy_node` is not the best naming convention, given we have a `CopyNode`class?

Thenerdstation:
oh shoot good point.

Thenerdstation:
I am open to suggestions. I agree we shouldn't use this name.

ayush2804:
Can I work on this issue?

Thenerdstation:
Go for it.

ayush2804:
ok@Thenerdstation

amilsted:
Since we have more than one type of `BaseNode` right now, should we maybe have `BaseNode.copy()` implementations for the subclasses? Then `tn.copy()` and `tn.copy_node()` can just call `node.copy()`, irrespective of the subclass.

MichaelMarien:
Yes!

Thenerdstation:
good idea!

jacksonwb:
Were you thinking BaseNode.copy() is an abstract method and then implement on Node and CopyNode classes? And then a call to this would always result in a new node with all dangling edges correct?

amilsted:
> Were you thinking BaseNode.copy() is an abstract method and then implement on Node and CopyNode classes? And then a call to this would always result in a new node with all dangling edges correct?

Yes to all of that!

Thenerdstation:
+1

Thenerdstation:
Closing as we now have a `node.copy()` method.

PR #447, opened 2020-01-25
Add a copy method to the node classes (#344)

jacksonwb:
This address issue #344 

* Add .copy() abstract method to the BaseNode Class

* Add copy method implementations to Node and CopyNode Class

* Add tests for both of these methods

* Change network operations copy function to use copy methods

* Add implementation to network_components_free_test TestNode class

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -503,6 +503,10 @@ class BaseNode(ABC):
         dtype=string_type,
         data=np.array([edge.name for edge in self.edges], dtype=object))
 
+  @abstractmethod
+  def copy(self) -> "BaseNode":
+    return
+
   def fresh_edges(self, axis_names: Optional[List[Text]] = None) -> None:
     if not axis_names:
       axis_names = self.axis_names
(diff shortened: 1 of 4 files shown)


tensornetwork/network_components.py
```diff
@@ -638,6 +642,27 @@ def get_tensor(self) -> Tensor:
...
+        new_edge = Edge(new_node, i, 
+                        name=edge.name,
+                        node2=new_node, 
+                        axis2=edge.axis2)
+        new_node.add_edge(new_edge, i)
+        new_node.add_edge(new_edge, edge.axis2)
+      else:
+        new_node.add_edge(Edge(new_node, i, name=edge.name), i)
+    return new_node
+
```
Thenerdstation: Might make sense to add a `conjugate` so that you can do `node.copy(conjugate=True)`.

tensornetwork/network_operations.py
```diff
@@ -126,27 +126,10 @@ def copy(nodes: Iterable[BaseNode],
...
-      else:
-        node_dict[node] = Node(
-            node.tensor,
-            name=node.name,
-            axis_names=node.axis_names,
-            backend=node.backend)
+    node_dict[node] = node.copy()
+    if isinstance(node, Node) and conjugate:
+      node_dict[node].tensor = node.backend.conj(node.tensor)
+
```
Thenerdstation: See above comment. Move this logic to the `node.copy()` method.

tensornetwork/network_components.py
jacksonwb: Do we want this to apply to the CopyNode also? Or should CopyNode ignore a conjugate argument?

Thenerdstation: Add the argument so the function is compatible, but a copy node is totally real, so a conjugation is a no-op. Just ignore it.

tensornetwork/network_components.py
```diff
@@ -753,6 +780,30 @@ def get_tensor(self) -> Tensor:
...
+      if edge in visited_edges:
+        continue
+      visited_edges.add(edge)
+      if edge.node1 == edge.node2:
+        new_edge = Edge(new_node, i, 
+                        name=edge.name,
+                        node2=new_node, 
+                        axis2=edge.axis2)
+        new_node.add_edge(new_edge, i)
+        new_node.add_edge(new_edge, edge.axis2)
```
Thenerdstation: Looks like you're adding a special case for traced edges here (Which is good! I agree this seems correct.) Can you add a test case for this then so we have 100% coverage?

jacksonwb: 👍

Thenerdstation:
Good work.

Issue #443, opened 2020-01-21
Unsafe "None" checking


MichaelMarien:
Throughout the library, a lot of initialisation check if an input is provided and compare it to the default value `None`. However usually this code looks like (example):

```class CopyNode(BaseNode):

  def __init__(self,
                     rank: int,
                     ...
                     backend: Optional[Text] = None,
    ...
    if not backend:
      backend = `config.default_backend
```

Values like `False,[], 0`,... actually pass this test. A better way of testing is `if backend is not None`.

PR #449, opened 2020-01-25
Fix unsafe None checks

jayanthchandra:
This PR address Issue #443

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -554,7 +554,7 @@ class Node(BaseNode):
       #always use the `Node`'s backend
       backend = tensor.backend
       tensor = tensor.tensor
-    if not backend:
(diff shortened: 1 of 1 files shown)


tensornetwork/network_components.py
```diff
@@ -554,7 +554,7 @@ def __init__(self,
       #always use the `Node`'s backend
       backend = tensor.backend
       tensor = tensor.tensor
-    if not backend:
+    if backend is not None:
```
MichaelMarien: whoops, I believe you mixed up? it should be `if backend is None` instead of `if backend is not None`

jayanthchandra: Yeah, Going through it i thought the check should be for `None`. The Issue mentioned `not None` to be added so went ahead with it. I will make that change.

PR #452, opened 2020-01-29
Block sparse merged

mganahl:
cleaned up code


mganahl:
shoot, I think this is because I fetched the upstream/master and merged it locally into my branch prior to submitting the PR ...

PR #473, opened 2020-02-19
add more stuff to charge.py

mganahl:


mganahl:
thats a bit more than 200 lines in charge.py, sry! a lot is docstrings though

tensornetwork/block_sparse/charge.py
```diff
@@ -114,3 +114,272 @@ def __repr__(self):
...
+    Args:
+      other: A `BaseCharge` object.
+    Returns:
+      BaseCharge: The result of fusing `self` with `other`.
+    """
+
+    # fuse the unique charges from each index, then compute new unique charges
+    comb_charges = fuse_ndarray_charges(self.unique_charges,
+                                        other.unique_charges, self.charge_types)
+    [unique_charges, charge_labels] = np.unique(
```
Thenerdstation: You can remove the brackets as python will automatically expand returned lists.

tensornetwork/block_sparse/charge.py
```diff
@@ -114,3 +114,272 @@ def __repr__(self):
...
+        comb_charges, return_inverse=True, axis=1)
+    charge_labels = charge_labels.reshape(self.unique_charges.shape[1],
+                                          other.unique_charges.shape[1]).astype(
+                                              np.int16)
+
+    # find new labels using broadcasting
+    charge_labels = charge_labels[(
+        self.charge_labels[:, None] +
```
Thenerdstation: Can you break this into multiple statements? It's hard to read currently

tensornetwork/block_sparse/charge.py
```diff
@@ -114,3 +114,272 @@ def __repr__(self):
...
+               charge_types: Optional[List[Type["BaseCharge"]]] = None) -> None:
+    super().__init__(charges, charge_labels, charge_types=[type(self)])
+
+  @staticmethod
+  def fuse(charge1, charge2) -> np.ndarray:
+    return np.add.outer(charge1, charge2).ravel()
+
+  @staticmethod
+  def dual_charges(charges) -> np.ndarray:
+    return charges * charges.dtype.type(-1)
```
Thenerdstation: What does type(-1) do?

tensornetwork/block_sparse/charge.py
```diff
@@ -114,3 +114,272 @@ def __repr__(self):
...
+  `[10, 100, 20, 200, 30, 300]`.
+  When using row-major ordering of indices in `BlockSparseTensor`,
+  the position of `degen1` should be "to the left" of the position of `degen2`.
+  Args:
+    degen1: Iterable of integers
+    degen2: Iterable of integers
+  Returns:
+    np.ndarray: The result of fusing `dege1` with `degen2`.
+  """
+  return np.reshape(degen1[:, None] * degen2[None, :],
```
Thenerdstation: This slicing trick won't work with a python list I believe.

tensornetwork/block_sparse/charge_test.py
```diff
@@ -42,3 +42,238 @@ def test_BaseCharge_copy():
...
+    i2 = fused_2 == target.charges[1, 0]
+    i3 = fused_3 == target.charges[2, 0]
+    #pylint: disable=no-member
+    nz_2 = np.nonzero(np.logical_and.reduce([i1, i2, i3]))[0]
+    return nz_1, nz_2
+
+  nz_1, nz_2 = run_test()
+  while len(nz_1) == 0:
+    nz_1, nz_2 = run_test()
+  assert np.all(nz_1 == nz_2)
```
Thenerdstation: This test looks like you're doing a lot, but this is the only assert and it seems kind of weak. Are there other checks you can do in the `run_test` function?

tensornetwork/block_sparse/charge_test.py
```diff
@@ -42,3 +42,238 @@ def test_BaseCharge_copy():
...
+
+  def run_test():
+    D = 2000
+    B = 6
+    O1 = np.random.randint(-B // 2, B // 2 + 1, D).astype(np.int16)
+    O2 = np.random.randint(-B // 2, B // 2 + 1, D).astype(np.int16)
+    P1 = np.random.randint(0, B + 1, D).astype(np.int16)
+    P2 = np.random.randint(-B // 2, B // 2 + 1, D).astype(np.int16)
+    Q1 = np.random.randint(1, B + 1, D).astype(np.int16)
+    Q2 = np.random.randint(1, B + 1, D).astype(np.int16)
```
Thenerdstation: In general it's a good idea to avoid random values in a test if you can help it. Non-seeded random values tend to cause tests to be flakey. Imagine having a test fail and then not be able to fix it because it depended on a specific set of random inputs.

I'll let it slide for now but keep it in mind to prefer to hard code things

tensornetwork/block_sparse/charge_test.py
```diff
@@ -42,3 +42,238 @@ def test_BaseCharge_copy():
...
+    i2 = fused_2 == target.charges[1, 0]
+    i3 = fused_3 == target.charges[2, 0]
+    #pylint: disable=no-member
+    nz_2 = np.nonzero(np.logical_and.reduce([i1, i2, i3]))[0]
+    return nz_1, nz_2
+
+  nz_1, nz_2 = run_test()
+  while len(nz_1) == 0:
+    nz_1, nz_2 = run_test()
+  assert np.all(nz_1 == nz_2)
```
Thenerdstation: What's the difference between the above couple tests? It looks like a lot of copying and pasting. Can use parameterize it instead?

tensornetwork/block_sparse/charge_test.py
mganahl: no problem, I'll fix this by seeding

tensornetwork/block_sparse/charge_test.py
mganahl: think so, i'll fix this

tensornetwork/block_sparse/charge.py
mganahl: fixed

tensornetwork/block_sparse/charge.py
mganahl: fixed

tensornetwork/block_sparse/charge.py
mganahl: type casting (casting -1 to whatever dtype charges have)

tensornetwork/block_sparse/charge.py
mganahl: absolutely not! I'll fix it

tensornetwork/block_sparse/charge_test.py
mganahl: This tests if charge fusion is correct by fusing charges and comparing the result with a numpy computed fusion. Most of this is initializing, and then recomputing everything directly in numpy to compare with.
nz_1 and nz_2 are large arrays that have to be identical, so I think the test is good. There's not really anything smaller to test here

tensornetwork/block_sparse/charge.py
Thenerdstation: Aw I see. Ok this should be fine then.

mganahl:
this should be good to go. Pls pull this in before #474

PR #474, opened 2020-02-19
more into charge.py

mganahl:
depends on #473


mganahl:
fixed merge conflict, should be good to go

Thenerdstation:
Looks good. Thanks for taking the time to do this in parts. You're code is getting much cleaner. :)

mganahl:
No problem! The code definitely benefits from doing it this way :)

PR #485, opened 2020-03-18
add _find_diagonal_sparse_blocks

mganahl:
+ tests


tensornetwork/block_sparse/block_tensor_test.py
```diff
@@ -223,3 +223,80 @@ def test_find_diagonal_sparse_blocks(num_legs, num_charges):
...
   np.testing.assert_allclose(unique_left, cs.charges)
+
+
+@pytest.mark.parametrize('num_legs', [2, 3, 4])
+@pytest.mark.parametrize('num_charges', [1, 2, 3])
+def test_find_transposed_diagonal_sparse_blocks(num_legs, num_charges):
+  np.random.seed(10)
+  order = np.arange(num_legs)
+  while np.all(order == np.arange(num_legs)):
+    np.random.shuffle(order)
```
Thenerdstation: Why do you have this in a while loop?

mganahl: wanted to avoid  to make my own choice. I can just hardcode the result of the shuffle if you want.

tensornetwork/block_sparse/block_tensor_test.py
```diff
@@ -223,3 +224,86 @@ def test_find_diagonal_sparse_blocks(num_legs, num_charges):
...
+      for c in range(num_charges)
+  ],
+                      axis=0)
+  fused = np.stack([
+      fuse_ndarrays([np_charges[n][c, :]
+                     for n in range(num_legs)])
+      for c in range(num_charges)
+  ],
+                   axis=0)
+
```
Thenerdstation: These are really hard to parse. Remember, there's no same in using multiple lines of code!

As a general rule of thumb, you generally want to avoid nested list comprehension, especially when it has to be broken into multiple lines like this. (We're explicitly not allowed to use it in Google's codebase). Prefer to use a normal for-loop for situations like this.

mganahl: sound goods, I'll change it!

PR #506, opened 2020-03-30
Final PR for block sparsity

mganahl:
last parts for block sparsity support:
module level import, tests + a few bug fixes


Thenerdstation:
This PR is over 900 lines of code and you are making significant changes that I disagree with. Please break this up into PRs of <200 lines so that it is easy to review.

tensornetwork/network_operations.py
```diff
@@ -293,14 +293,27 @@ def split_node(
...
   sqrt_s = backend.sqrt(s)
-  u_s = u * sqrt_s
-  # We have to do this since we are doing element-wise multiplication against
-  # the first axis of vh. If we don't, it's possible one of the other axes of
-  # vh will be the same size as sqrt_s and would multiply across that axis
-  # instead, which is bad.
-  sqrt_s_broadcast_shape = backend.shape_concat(
-      [backend.shape_tensor(sqrt_s), [1] * (len(vh.shape) - 1)], axis=-1)
-  vh_s = vh * backend.reshape(sqrt_s, sqrt_s_broadcast_shape)
+  if backend.name != 'symmetric':
```
Thenerdstation: We should NEVER have if statements based on the backed that don't raise errors. All of these difference should be handled by the backend classes. If they don't, we need to restructure the backends.

Thenerdstation:
Actually I take this back, most of the lines are tests.

tensornetwork/network_operations.py
mganahl: Let me see how to fix this

Thenerdstation: It looks like the only thing we nee to add is element wise multiplication.

tensornetwork/network_operations.py
```diff
@@ -293,14 +293,27 @@ def split_node(
...
+  else:
+    u_shape = u.shape
+    vh_shape = vh.shape
+    utmp = backend.reshape(u,
+                           (np.prod(u_shape[:len(u_shape) - 1]), u_shape[-1]))
+
+    vhtmp = backend.reshape(vh,
+                            (vh_shape[0], np.prod(vh_shape[1:len(vh_shape)])))
+    #NOTE (mganahl): we use mulitplication by diagonal matrix here
+    u_s = backend.reshape(utmp @ backend.diag(sqrt_s), u_shape)
```
Thenerdstation: Prefer to use `backend.tensordot(a, b, [[1], [0]])` over the matmul operator since that isn't guarantee to be overloaded.

tensornetwork/network_operations.py
mganahl: I'd like to avoid adding broadcasting support to `BlockSparseTensor` for now, because this could be a bit of a rabbit hole (I'm willing to add it later on though). The best alternative option I see for now is to add a `broadcast_left_multiply` and `broadcast_right_multiply` to the backends.  What do you think?

Thenerdstation: fine by me.

Thenerdstation:
Oh we're so close to getting this released and I'd hate to let this slide. Are you stuck on this PR?

mganahl:
I'm on it! the stuff I was submitting were minor things I found when working on this. Should be done really soon!

mganahl:
There are a few more PR coming still

Thenerdstation:
SG! I'm very excited for to release this to the public. :)

mganahl:
depends on  #514 
I removed the backend specific code from `network_operations.py`

Thenerdstation:
🥳

mganahl:
yippie

PR #510, opened 2020-03-31
Fix basecharge unique

mganahl:
add guarantee that `BaseCharge.unique()` returns ordered `BaseCharge`

### DIFF
--- a/tensornetwork/block_sparse/charge.py
+++ b/tensornetwork/block_sparse/charge.py
@@ -336,10 +336,10 @@ class BaseCharge:
     return obj
 
   def unique(self,
-             return_index: Optional[bool] = False,
-             return_inverse: Optional[bool] = False,
-             return_counts: Optional[bool] = False,
-             sort: Optional[bool] = True) -> Any:
+             return_index: bool = False,
+             return_inverse: bool = False,
+             return_counts: bool = False,
+             sort: bool = True) -> Any:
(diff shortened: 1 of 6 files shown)


tensornetwork/block_sparse/charge.py
```diff
@@ -306,35 +305,56 @@ def intersect(self, other, assume_unique=False, return_indices=False) -> Any:
...
+    return obj
+
   def unique(self,
-             return_index=False,
-             return_inverse=False,
-             return_counts=False) -> Any:
+             return_index: Optional[bool] = False,
+             return_inverse: Optional[bool] = False,
+             return_counts: Optional[bool] = False,
+             sort: Optional[bool] = True) -> Any:
```
Thenerdstation: Optional means these arguments can be None, is that what you actually want here?

mganahl: Oh, actually no. They shoud be `bool`

mganahl: removed the `Optional`.

mganahl: turns out that `None` is fine, put `Optional` back

Thenerdstation: There's a difference between fine and correct. Let's keep it as just bool.

PR #516, opened 2020-04-02
Z2charge

mganahl:


tensornetwork/block_sparse/charge.py
```diff
@@ -556,6 +556,42 @@ def random(cls, minval: int, maxval: int, dimension: int) -> BaseCharge:
...
+
+  @staticmethod
+  def dual_charges(charges) -> np.ndarray:
+    return charges
+
+  @staticmethod
+  def identity_charge() -> np.ndarray:
+    return np.int16(0)
+
+  #pylint: disable=arguments-differ
```
Thenerdstation, line 588: Its generally a bad idea to change the input signature of an overloaded function. Perhaps we just change the order of the arguments for `BaseCharge.random` to `dimension, min, max`?

Thenerdstation: And then you can set the defaults to `min=0, max=1` and just raise an error otherwise.

PR #517, opened 2020-04-02
set seed in tests

mganahl:
one of the tests in wavefunctions broke at some point. the cause was unclear because there was no seed set in the tests


mganahl:
sry, there is a lot of noise in this PR because of changing the formating to chromium style. I only added a tf.randon.set_seed(10) at two places

PR #520, opened 2020-04-03
added slice_edge

### LINKED ISSUE #260: Add support for slicing an edge

jensenjhwang:
Resolves #260

### DIFF
--- a/tensornetwork/backends/numpy/numpy_backend.py
+++ b/tensornetwork/backends/numpy/numpy_backend.py
@@ -37,6 +37,17 @@ class NumPyBackend(base_backend.BaseBackend):
   def transpose(self, tensor, perm):
     return self.np.transpose(tensor, perm)
 
+  def slice(self,
+            tensor: Tensor,
+            start_indices: Tuple[int, ...],
+            slice_sizes: Tuple[int, ...]) -> Tensor:
+    if len(start_indices) != len(slice_sizes):
+      raise ValueError("Lengths of start_indices and slice_sizes must be"
+                       "identical.")
+    obj = tuple(slice(start, start + size) for start, size
+                in zip(start_indices, slice_sizes))
+    return tensor[obj]
+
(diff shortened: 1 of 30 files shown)


jensenjhwang:
@googlebot I signed it!

tensornetwork/tests/network_components_free_test.py
```diff
@@ -1586,6 +1586,64 @@ def test_split_edge_different_backend_raises_value_error(single_node_edge):
...
+  assert new_edge.axis2 == 1
+  assert new_edge.dimension == 2
+
+  expected_tensor = np.array([[4, 5], [7, 8]])
+  np.testing.assert_allclose(expected_tensor, node.get_tensor())
+
+
+def test_slice_edge_dangling_edge(backend):
+  node = Node(np.arange(9).reshape(3, 3), backend=backend)
+  edge = Edge(node1=node, axis1=0)
```
Thenerdstation, line 1616: Just do `edge = node[0]`. Nodes automatically are given dangling edges upon creating.

tensornetwork/backends/numpy/numpy_backend.py
```diff
@@ -37,6 +37,17 @@ def reshape(self, tensor: Tensor, shape: Tensor):
   def transpose(self, tensor, perm):
     return self.np.transpose(tensor, perm)
 
+  def slice(self,
```
Thenerdstation, line 40: Be sure to also add this method to the `BaseBackend` abstract class aswell.

Thenerdstation:
Take a look at the linter and pytype errors. Once those are fixed this should be good to pull in.

PR #521, opened 2020-04-03
allow charge initialization form list

mganahl:


tensornetwork/block_sparse/charge.py
```diff
@@ -57,10 +57,11 @@ def __next__(self):
       raise StopIteration
 
   def __init__(self,
-               charges: np.ndarray,
+               charges: Union[List, np.ndarray],
                charge_labels: Optional[np.ndarray] = None,
                charge_types: Optional[List[Type["BaseCharge"]]] = None,
                charge_dtype: Optional[Type[np.number]] = np.int16) -> None:
+    charges = np.array(charges)
```
Thenerdstation, line 64: use `np.asarray` instead to prevent an unnecessary copy.

PR #522, opened 2020-04-03
Split node fix

### LINKED ISSUE #519: split_node_... mutates its input
`split_node_full_svd(node, [node[0], node[2]], [node[1]])` mutates the shape of `node`.

mganahl:
fixes #519

### DIFF
--- a/tensornetwork/network_operations.py
+++ b/tensornetwork/network_operations.py
@@ -284,6 +284,14 @@ def split_node(
     right_axis_names = None
 
   backend = node.backend
+  #store original edge order
+  orig_axes = []
+  for edge in left_edges + right_edges:
+    if edge.node1 is node:
+      orig_axes.append(edge.axis1)
+    else:
+      orig_axes.append(edge.axis2)
+
   node.reorder_edges(left_edges + right_edges)
 
   u, s, vh, trun_vals = backend.svd_decomposition(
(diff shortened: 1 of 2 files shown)


tensornetwork/network_operations.py
```diff
@@ -309,6 +317,12 @@ def split_node(
     edge.update_axis(i + len(left_edges), node, i + 1, right_node)
   connect(left_node.edges[-1], right_node.edges[0], name=edge_name)
   node.fresh_edges(node.axis_names)
+  #rerstore the original edge order of node
+  orig_edge_order = [None] * node.get_rank()
+  for n, o in enumerate(orig_axes):
+    orig_edge_order[o] = node.edges[n]
+  node.reorder_edges(orig_edge_order)
+
```
Thenerdstation, line 325: I'm a little against this solution. It does a second unnecessary transpose which could cause unintended slow downs, especially when running in a distributed setting. 

I think a better approach would be to add a `node.tensor_from_edge_order(edges)` which would return the transposed tensor but not modify `node`.

mganahl: good point. why not `Node.transpose(List[Union[int,Edge]])`?

mganahl: never mind, I think `tensor_from_edge_order` is better.

PR #532, opened 2020-04-14
simpler definition of custom charge

mganahl:
in `BaseCharge.__init__`, set `self.charge_types` to `[type(self)]` if `charge_types=None` is passed.
This eliminates the neccessity for the user to write their own __init__ in derived classes

### DIFF
--- a/tensornetwork/block_sparse/charge.py
+++ b/tensornetwork/block_sparse/charge.py
@@ -74,6 +74,8 @@ class BaseCharge:
       label_dtype = np.int16
     else:
       label_dtype = np.int32
+    if charge_types is None:
+      charge_types = [type(self)]
     self.charge_types = charge_types
     if charge_labels is None:
       if charges.shape[1] > 0:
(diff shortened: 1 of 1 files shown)


tensornetwork/block_sparse/charge.py
```diff
@@ -528,16 +530,6 @@ def isin(self, target_charges: Union[np.ndarray, "BaseCharge"]) -> np.ndarray:
 
 class U1Charge(BaseCharge):
   """Charge Class for the U1 symmetry group."""
-  def __init__(self,
```
Thenerdstation, line 531: Why are we removing the init function here and not for the Z2 and Modular charges?

mganahl: U1Charge doesn't perform any checks currently, the others do.

mganahl:
let's actually wait with this. I wan't test it a bit more

tensornetwork/block_sparse/charge.py
Thenerdstation: Oh I see, you're defaulting to `BaseCharge`'s init, got it.

Thenerdstation:
Oh no missed your comment. That's ok, feel free to add more testing in the next PR

mganahl:
np  :)

PR #533, opened 2020-04-14
Removal of the signature attribute

### LINKED ISSUE #446: Fix graphviz
Looks like this might be a bug:
Screen Shot 2020-01-23 at 15 53 00

### LINKED ISSUE #530: signature of node is always default as -1
I am wondering if this is the expected behavior for creating nodes since signature should be somewhat the unique id for the nodes. By default, all nodes created by tn.Node has signature -1, and this will have impact on several other utilities at least on `tn.to_graphviz` which depends on signature to draw different nodes. 

```python
s = [tn.Node(np.array([[1.0], [0]], dtype=np.complex64), name="l"),
     tn.Node(np.array([[1.0], [0]], dtype=np.complex64), name="r"), ]
print(s[0].signature, s[1].signature) # -1, -1
tn.to_graphviz(s)
```
The above snippet shows diagram with only one node instead of two due to the signature issue.

My suggestion: within ``__init__`` of tn.Node, the default signature should be generated as: 1. based on node name if set 2. a default global unique id if no node name is specified (perhas via class attr of tn.Node).

Thenerdstation:
Fixes #530 fixes #446

### DIFF
--- a/tensornetwork/contractors/opt_einsum_paths/utils.py
+++ b/tensornetwork/contractors/opt_einsum_paths/utils.py
@@ -38,13 +38,12 @@ def _get_path_nodes(
   Returns:
     The optimal contraction path as returned by `opt_einsum`.
   """
-  sorted_nodes = sorted(nodes, key=lambda n: n.signature)
-
-  input_sets = [set(node.edges) for node in sorted_nodes]
+  nodes = list(nodes)
+  input_sets = [set(node.edges) for node in nodes]
   output_set = get_subgraph_dangling(nodes)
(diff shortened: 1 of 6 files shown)


Thenerdstation:
I have confirmed this fixes the graphviz issue.
```python
import tensornetwork as tn
import numpy as np

a = tn.Node(np.ones((2, 2)), name="A")
b = tn.Node(np.ones((2, 2)), name="B")
a[0] ^ b[0]

tn.to_graphviz([a, b])
```
image

Thenerdstation:
Nope. And I believe I removed that already?

mganahl:
you're right, was already gone

PR #542, opened 2020-04-16
fix doc

mganahl:


mganahl:
should fix the issue with the block sparse tutorial

Thenerdstation:
Thanks for the quick fix!

PR #544, opened 2020-04-18
add sin cos exp log expm to backends

refraction-ray:
If this is ok, I can add more element wise and matrix ops later such as tan, cosh, sinm and so on. maybe we need to consider generating these methods dynamically since they follows the same pattern.

I implemented these methods in numpy, jax and tensorflow backends. I don't add torch one since I always need complex numbers, but impl with pytorch should be straightforward.

### DIFF
--- a/tensornetwork/backends/numpy/numpy_backend.py
+++ b/tensornetwork/backends/numpy/numpy_backend.py
@@ -17,6 +17,7 @@ from tensornetwork.backends import base_backend
 from tensornetwork.backends.numpy import decompositions
 import numpy
 import scipy
+from scipy import linalg
 Tensor = Any
 
 
@@ -27,8 +28,6 @@ class NumPyBackend(base_backend.BaseBackend):
     super(NumPyBackend, self).__init__()
     self.np = numpy
     self.sp = scipy
-    # pylint: disable=import-outside-toplevel
-    from scipy import linalg
     self.sp.linalg = linalg
     self.name = "numpy"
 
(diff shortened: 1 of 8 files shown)


tensornetwork/backends/numpy/numpy_backend.py
```diff
@@ -26,6 +26,10 @@ class NumPyBackend(base_backend.BaseBackend):
   def __init__(self):
     super(NumPyBackend, self).__init__()
     self.np = numpy
+    self.sp = scipy
+    # pylint: disable=import-outside-toplevel
+    from scipy import linalg
```
Thenerdstation, line 31: Can you move this import to the top?

tensornetwork/backends/tensorflow/tensorflow_backend_test.py
```diff
@@ -411,3 +411,41 @@ def test_sparse_shape():
...
+  matrix2 = getattr(tf.linalg, method)(matrix)
+  np.testing.assert_almost_equal(matrix1.numpy(), matrix2.numpy())
+
+
+@pytest.mark.parametrize("dtype,method",
+                         [(tf.float64, "expm"), (tf.complex128, "expm")])
+def test_matrix_ops_raises(dtype, method):
+  backend = tensorflow_backend.TensorFlowBackend()
+  matrix = backend.randn((4, 4, 4), dtype=dtype, seed=10)
+  with pytest.raises(ValueError):
```
Thenerdstation, line 447: Be sure to include the `match=` parameter for all `raises` tests. That way we ensure you're hitting the one you intend to hit.

PR #556, opened 2020-04-30
Adding DenseDecomp layer

bpenchas:
Adding a new DenseDecomp layer and associated tests.


tensornetwork/tn_keras/dense.py
```diff
@@ -0,0 +1,120 @@
+from keras.layers import Layer
+from keras import activations
+from keras import initializers
```
Thenerdstation, line 3: You should use `tf.keras` instead.

requirements_travis.txt
```diff
@@ -4,3 +4,4 @@ pytest
 torch>=1.1
 jax>=0.1.0
 jaxlib>=0.1.27
+Keras==2.3.1
```
Thenerdstation, line 7: You should be able to remove this if you tensorflow's keras instead.

Issue #553, opened 2020-04-29
conflict between pickle and thus multiprocessing with backend module assignment


refraction-ray:
In the current implementations of backends, we have the following thing:
```python
class SomeBackend():
    def __init__(self):
        self.np = numpy
```

Note numpy is a module, and of course such backend is included as one attr of `tn.Node`. 

The effect of such design is you will fail when you try:

```python
from multiprocessing import pool
def f(x): 
    return tn.Node(np.eye(x))
with Pool(3) as p:
        print(p.map(f, [1, 2, 3]))
```
the error is `can't pickle module objects` which is induced ultimately by `numpy` attr in backend class. This is due to the lack of ability to pickle modules in python. To verify this, one can either comment ``self.backend=backend`` in network_components or change all ``self.numpy`` to plain ``numpy`` in numpy_backend py. There will be no error for both modifications.

It is worth noting that other multiprocessing libraries such as pathos has no problem to interact with TensorNetwork as its core pickle library is dill which is more powerful. However, you may want to interact TN with other libraries with python builtin multiprocessing baked in. In such cases, TN fails the task.

Maybe we should just use plain ``np``, ``tf`` in backend files instead of ``self.np`` to overcome this?

Thenerdstation:
I believe we did it this way because we want to make sure we always support users who don't have all of the supported backends installed. If you just have a plain `import tensorflow` at the top instead of in side the class, you'll immediately crash upon importing TN if you do not have TF installed.

Thenerdstation:
If you have a workaround that solves both your issue and the import issue mentioned above, we'd gladly accept a PR!

Thenerdstation:
Aw yes it's all coming back to me as for why we did this. Checkout the discussion on https://github.com/google/TensorNetwork/pull/241#discussion_r322351287

Its a weird issue with how our build system works internally. If some files in a module aren't imported when the module itself is imported, then those files won't be included in the file `.par` package and you'll hit those runtime errors seen in that discussion.

mganahl:
We could try to  assign imported modules to global variables like so:
```python 
class Backend:

  def __init__(self):
    global jax
    try:
      import jax as jaxlib
    except ImportError:
      raise ImportError("jax not installed")
    jax = jaxlib


if __name__ == "__main__":
  backend = Backend()
  b = jax.numpy.zeros(10)
  print(b)
```
This could be potentially confusing though because `jax` will be at global scope even though the user never imported it.

[...]

PR #562, opened 2020-04-30
Change import behavior on TF and PyTorch backends.

Thenerdstation:
Progress on #553

### DIFF
--- a/tensornetwork/backends/pytorch/pytorch_backend.py
+++ b/tensornetwork/backends/pytorch/pytorch_backend.py
@@ -27,6 +27,7 @@ class PyTorchBackend(base_backend.BaseBackend):
 
   def __init__(self):
     super(PyTorchBackend, self).__init__()
+    # pylint: disable=global-variable-undefined
     global torchlib
     try:
       #pylint: disable=import-outside-toplevel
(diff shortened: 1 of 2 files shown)


mganahl:
I can review, let me know when its ready

Thenerdstation:
Should be good to go. Only error was a small linting issue.

refraction-ray:
I am still not sure what will go wrong in the following approach:

```python
# somebackend.py

try:
  import tensorflow as tf
  withtf = True
except ImportError:
  withtf = False

class SomeBackend(BaseBackend):
  def __init__(self):
    if not withtf:
      raise Exception("tf not installed")
  def methd(self, tensor):
    tf.methd(tensor)
 ```

Maybe the point here is to avoid unnecessary import overhead?

Thenerdstation:
Yeah in that case, if you have Jax, TF and PyTorch installed you're going to be importing all of them, which may not be what you want.

mganahl:
Hey @Thenerdstation are you working on fixing the other backends as well? If not I'm going to start a PR for lanczos for the jax backend

Thenerdstation:
I'm going to try to finish it up tonight.

PR #565, opened 2020-04-30
use single backend object

### LINKED ISSUE #561: switch to using a single instantiation of backend objects
We're currently using a new backend object every time we create e.g. a `Node`. 
To facilitate implementation of DMRG into backends like jax or tensorflow we should
switch to instantiating only one object per backend, and then keep reusing it .

mganahl:
fixes #561

### DIFF
--- a/tensornetwork/backends/backend_factory.py
+++ b/tensornetwork/backends/backend_factory.py
@@ -29,6 +29,9 @@ _BACKENDS = {
     "symmetric": symmetric_backend.SymmetricBackend
 }
 
+#we instantiate each backend only once and store it here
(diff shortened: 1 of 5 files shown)


mganahl:
I also took the time to fix a few typing issues in blocksparse that popped up on my platform. No idea why our build didn't catch them

Thenerdstation:
We're using an older version of pytype because the newer versions break and upgrading post-breaking is hard. :)

Looks like a single import test method is failing, likely because it's just grabbing the cached backend rather than running `__init__` and hitting the `import jax` line. I can fix this.

PR #566, opened 2020-04-30
Defining MPS and memory usage

sebgrijalva:
I'd like to add an example of how to define Matrix Product States using tensornetwork's MPS class and how the memory size of the objects is optimized with respect to just using numpy arrays.


mganahl:
nice works, thanks! I'll have a look

Thenerdstation:
I'll leave the decision to pull this in up to you martin.

mganahl:
Hey @sebgrijalva, this looks good in general. There are a few issues that should be addressed (consistent axis scales, some rephrasing), nothing major. Would you be OK if you took this and made it into a small MPS tutorial on RDT instead of a notebook?

sebgrijalva:
Hello @mganahl , thanks for the feedback. Sure, let me improve more carefully the file and commit it as a Read The Docs document (I assume this is what you meant). : )

sebgrijalva:
Hello, I have the .rst file ready, althouth I'm not sure that the math expressions show. Could you take a look at the file and tell me what you think?

sebgrijalva:
@googlebot I fixed it

sebgrijalva:
Will open a new pull request because there's a user mismatch for this one

PR #568, opened 2020-05-01
Fix backend issues with Jax and Numpy backends

Thenerdstation:


Thenerdstation:
@refraction-ray Once this is merged I'll push a new version of TN for you.

Thenerdstation:
@mganahl It's ready whenever you are.

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -49,37 +60,122 @@ def slice(self,
...
+    return tensor.shape
+
+  def shape_tuple(self, tensor: Tensor) -> Tuple[Optional[int], ...]:
+    return tensor.shape
+
+  def sparse_shape(self, tensor: Tensor) -> Tuple[Optional[int], ...]:
+    return self.shape_tuple(tensor)
+
+  def shape_prod(self, values: Tensor) -> Tensor:
+    return np.prod(values)
```
mganahl, line 104: is there a reason why this uses numpy instead of jax.numpy?

Thenerdstation: Literally the only reason this calls numpy at all is because python doesn't have its own `prod` method.

PR #570, opened 2020-05-01
Lanczos for jax

mganahl:


Thenerdstation:
Can we change this PR so that it doesn't change 15 files?

mganahl:
sure, I can do one backend at a time

mganahl:
Uhh, actually the problem is the linter. I had to adapt the signature of all eigsh_lanczos functions to the jax version to make the API consistent.

Thenerdstation:
Aw ok. didn't realize we were chaning the function signature. This is fine then.

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -16,11 +16,15 @@
...
 from tensornetwork.backends.numpy import decompositions
 import numpy as np
+from tensornetwork.backends.jax import jitted_functions
+from functools import partial
 
 Tensor = Any
 
-
 # pylint: disable=abstract-method
+# pylint: disable=attribute-defined-outside-init
```
Thenerdstation, line 25: Why was this added?

tensornetwork/tests/backend_factory_test.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2019 The TensorNetwork Authors
+#
```
Thenerdstation, line 2: I believe this file should be deleted.

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -299,11 +297,130 @@ def test_base_backend_eigs_not_implemented():
...
+  with pytest.raises(ValueError):
+    backend.eigsh_lanczos(lambda x: x, [], numeig=2, reorthogonalize=False)
+  with pytest.raises(ValueError):
+    backend.eigsh_lanczos(lambda x: x, [], shape=(10,), dtype=None)
+  with pytest.raises(ValueError):
+    backend.eigsh_lanczos(lambda x: x, [], shape=None, dtype=np.float64)
+  with pytest.raises(ValueError):
+    backend.eigsh_lanczos(lambda x: x, [])
+  with pytest.raises(TypeError):
+    backend.eigsh_lanczos(lambda x: x, [], initial_state=[1, 2, 3])
```
Thenerdstation, line 423: Can you add `pytest.raises(ValueError, match="...")` so that we assure ourselves we're hitting the right ValueError?

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -0,0 +1,104 @@
+from functools import partial
+
+
+def _generate_jitted_eigsh_lanczos(jax):
+  """
+    Helper function to generate jitted lanczos function used in eigsh_lanczos.
+    """
+
+  @partial(jax.jit, static_argnums=(3, 4, 5, 6))
+  def jax_lanczos(matvec, arguments, init, ncv, neig, landelta, reortho):
```
Thenerdstation, line 10: Please document these arguments.

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -222,6 +226,115 @@ def cmplx_random_uniform(complex_dtype, real_dtype):
...
+        raise ValueError("if no `initial_state` is passed, then `shape` and"
+                         "`dtype` have to be provided")
+      initial_state = self.randn(shape, dtype)
+
+    if not isinstance(initial_state, jnp.ndarray):
+      raise TypeError("Expected a `jax.array`. Got {}".format(
+          type(initial_state)))
+    if hasattr(self, '_A'):
+      #pylint: disable=access-member-before-definition
+      if self._A is not A:
```
Thenerdstation, line 328: Why are you setting the passed function as a attribute to the backend? Seems very sketchy...

mganahl: Yeah, it is kinda sketchy... The reason is that I want to avoid retracing. 
`_generate_jitted_eighs_lanczos` jits the function `jax_lanczos` in jitted_functions.py. The first argument to `jax_lanczos` is `matvec`, a  callable that does the matrix-vector product. that function `matvec` has to be wrapped into `jax.tree_util.Partial`, otherwise `jax.jit` will complain. Now, `matvec` is typically passed by the user (name `A` there), but I'd like to avoid having users wrapping their `matvecs` into `jax.tree_util.Partial` because that breaks backend independence of their code. Thus I wrap it in the backend function `eigsh_lanczos`. Unfortunately, if I just wrap it at every call to `eigs_lanczos` that would trigger jitting of `jax_lanczos` at every  call, something I want to avoid. So at the first call to `eigsh_lanczos` I wrape `A` into `jax.tree_util.Partial` and store both the passed `A` and the wrapped `A`. If the user calls `eigsh_lanczos` again with the same `A`  I can check if I already wrapped `A`, and I can just pass the previously wrapped `A` (stored in `self._APartial`) to `jax_lanczos`, thus avoiding rejitting. Couldn't really find a better way to accomplish that

Thenerdstation: Can we do this instead?

Let's have a dictonary for this module called `_CACHED_FUNCTIONS = {}` and use that to store the methods. That way we avoid adding an unnecessary attribute to the backend AND we can cache more than one method at a time.

mganahl: that's actually a great idea, let me do that

tensornetwork/backends/jax/jax_backend.py
mganahl: removed it (not that `_jaxlan` is still defined outside init)

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -222,6 +226,115 @@ def cmplx_random_uniform(complex_dtype, real_dtype):
...
+      args: List[Tensor],
+      initial_state: Optional[Tensor] = None,
+      shape: Optional[Tuple] = None,
+      dtype: Optional[Type[np.number]] = None,
+      num_krylov_vecs: Optional[int] = 200,
+      numeig: Optional[int] = 1,
+      tol: Optional[float] = 1E-8,
+      delta: Optional[float] = 1E-8,
+      ndiag: Optional[int] = 20,
+      reorthogonalize: Optional[bool] = False) -> Tuple[List, List]:
```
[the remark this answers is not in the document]
mganahl: Based on experience

mganahl:
Thanks for the suggestion above, it's much better that way!

mganahl:
The tests I added should actually catch any issue with the numerics

Thenerdstation:
Right, that's also I didn't check the code by hand. :)

PR #572, opened 2020-05-01
enormous yapf

mganahl:
yapfed the whole repo to be consistent with the latest version of yapf under "yapf" style

### DIFF
--- a/tensornetwork/backends/backend_test.py
+++ b/tensornetwork/backends/backend_test.py
@@ -152,9 +152,6 @@ def test_basic_network_without_backends_raises_error():
     Node(np.ones((2, 2)), backend="pytorch")
 
 
-[]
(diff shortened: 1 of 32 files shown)


tensornetwork/backends/backend_test.py
```diff
@@ -150,8 +150,11 @@ def test_basic_network_without_backends_raises_error():
     Node(np.ones((2, 2)), backend="tensorflow")
   with pytest.raises(ImportError):
     Node(np.ones((2, 2)), backend="pytorch")
+
+
 []
```
Thenerdstation, line 155: ???

mganahl: hmm, where did this guy  come from ??

PR #575, opened 2020-05-02
Test finite mps

MichaelMarien:
* added some tests for finite mps (now at 100% cov)
* removed check_canconical implementation in finite_mps as it is identical to parent class


tensornetwork/matrixproductstates/finite_mps.py
```diff
@@ -132,22 +132,6 @@ def canonicalize(self, normalize: Optional[bool] = True) -> np.number:
     self.position(len(self.nodes) - 1, normalize=False)
     return self.position(pos, normalize=normalize)
 
-  def check_canonical(self) -> Tensor:
```
Thenerdstation, line 135: Why is this deleted?

MichaelMarien: Perhaps a bit over-aggressive, but it's the exact same implementation as the method in the parent class (BaseMPS). Don't mind putting it back though.

Thenerdstation: That's what I assumed, just making sure. Let keep it deleted.

PR #576, opened 2020-05-04
Adding MPO layer

bpenchas:


tensornetwork/tn_keras/mpo.py
```diff
@@ -0,0 +1,168 @@
...
+        except: # pylint: disable=W0702
+          pass
+        try:
+          tn_nodes[i][1] ^ tn_nodes[i + 1][1]
+        except: # pylint: disable=W0702
+          pass
+        try:
+          tn_nodes[i][2] ^ tn_nodes[i + 1][1]
+        except: # pylint: disable=W0702
+          pass
```
Thenerdstation, line 142: The naked try except is a little sketchy, especially since you're disabling the linter on this. If there is a real error connecting the legs (perhaps the dimensions were mismatched?) this would eat the error.

I assume you do this to account the the end nodes. Can you refactor this to not use try/excepts?

Thenerdstation:
Can you fix the merge conflicts?

tensornetwork/tn_keras/mpo.py
```diff
@@ -0,0 +1,167 @@
...
+    if input_shape[-1] is None:
+      raise ValueError('The last dimension of the inputs to `Dense` '
+                       'should be defined. Found `None`.')
+
+    def is_perfect_root(n, n_nodes):
+      root = n**(1. / n_nodes)
+      return round(root)**n_nodes == n
+
+    # Ensure the MPO dimensions will work
+    assert is_perfect_root(input_shape[-1],
```
markomernick, line 86: Nit: It's generally good to print the values in the error message - if a user gets this, it's likely they will have to tear apart their code or add print statements to see what the values actually were.

tensornetwork/tn_keras/mpo.py
```diff
@@ -0,0 +1,167 @@
...
+from tensorflow.keras import activations
+from tensorflow.keras import initializers
+from typing import List, Optional, Text, Tuple
+import tensornetwork as tn
+from tensornetwork.network_components import Node
+import numpy as np
+import math
+
+
+class DenseMPO(Layer):
```
markomernick, line 12: This layer needs a get_config() method.

tensornetwork/tn_keras/test_layer.py
```diff
@@ -1,11 +1,14 @@
...
+import math
 from tensorflow.keras import backend as K
 from tensorflow.keras.models import Sequential
 from tensornetwork.tn_keras.dense import DenseDecomp
+from tensornetwork.tn_keras.mpo import DenseMPO
+from tensorflow.keras.layers import Dense  # type: ignore
 
 
-@pytest.fixture(params=[50, 100, 256])
+@pytest.fixture(params=[256, 1024])
```
markomernick, line 11: Some other tests to add:

-> cls.from_config(layer.get_config()) should return an identically configured layer. (The weights will be different, but you should be able to call set_weights(get_weights()) to get an identical layer.
-> model.save() should not die. (This is required for Tensorflow Serving, which I think will be a huge win with this layer.)

bpenchas: Agreed! Added `test_config` and `test_model_save` to address these.

tensornetwork/tn_keras/mpo.py
bpenchas: Sounds great - added `get_config` to both layers.

tensornetwork/tn_keras/mpo.py
bpenchas: Good call, updated the error message

tensornetwork/tn_keras/mpo.py
```diff
@@ -0,0 +1,167 @@
...
+               output_dim: int,
+               num_nodes: int,
+               bond_dim: int,
+               use_bias: Optional[bool] = True,
+               activation: Optional[Text] = None,
+               kernel_initializer: Optional[Text] = 'glorot_uniform',
+               bias_initializer: Optional[Text] = 'zeros',
+               **kwargs) -> None:
+    if 'input_shape' not in kwargs and 'input_dim' in kwargs:
+      kwargs['input_shape'] = (kwargs.pop('input_dim'),)
```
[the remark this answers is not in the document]
bpenchas: Allows the user to specify `input_dim` instead of `input_shape` - this can be done in other Keras layers like Dense and I wanted to make the layer creation as consistent as possible. Of course, this is redundant and we could just force the user to specify `input_shape=(input_dim,)` in cases where `input_dim` would otherwise be used. I'll keep it for now, and remove it if you feel strongly we should enforce a full `input_shape`.

tensornetwork/tn_keras/mpo.py
```diff
@@ -0,0 +1,167 @@
...
+class DenseMPO(Layer):
+  """Matrix Product Operator (MPO) TN layer.
+
+  Example:
+
+  ```python
+  # as first layer in a sequential model:
+  model = Sequential()
+  model.add(
+    DenseMPO(1024, num_nodes=4, bond_dim=8, activation='relu', input_dim=1024))
```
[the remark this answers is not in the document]
bpenchas: See below

tensornetwork/tn_keras/mpo.py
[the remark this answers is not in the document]
markomernick: It's not necessary to handle this here - the superclass will do this for you.

tensornetwork/tn_keras/dense.py
```diff
@@ -122,3 +125,31 @@ def f(x: tf.Tensor, a_var: tf.Tensor, b_var: tf.Tensor, use_bias: bool,
...
+    # Serialize the activation
+    config['activation'] = activations.serialize(getattr(self, 'activation'))
+
+    # Serialize the initializers
+    decomp_initializers = ['kernel_initializer', 'bias_initializer']
+    for initializer_arg in decomp_initializers:
+      config[initializer_arg] = initializers.serialize(
+          getattr(self, initializer_arg))
+
+    return config
```
markomernick, line 155: Because you're passing a **kwargs into super(), you need to be sure to get the superclass config dict as well. The idiomatic Keras for that is:
    config = {...}
    base_config = super(CLS, self).get_config()
    return dict(list(base_config.items()) + list(config.items()))

tensornetwork/tn_keras/dense.py
```diff
@@ -55,14 +55,16 @@ def __init__(self,
                **kwargs) -> None:
     if 'input_shape' not in kwargs and 'input_dim' in kwargs:
```
markomernick, line 56: This is handled in super.__init__(), you can elide it here.

tensornetwork/tn_keras/test_layer.py
```diff
@@ -104,3 +120,84 @@ def test_num_parameters(dummy_data):
...
+  model = make_model
+  model.compile(optimizer='adam',
+                loss='binary_crossentropy',
+                metrics=['accuracy'])
+
+  # Train the model for 5 epochs
+  model.fit(data, labels, epochs=5, batch_size=32)
+
+  # Save model to a h5 file, then load from file
+  model.save('test_model.h5')
```
markomernick, line 190: You might also want to test saving in format=tf (which is SavedModel).

tensornetwork/tn_keras/mpo.py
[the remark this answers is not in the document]
markomernick: Edit: I see now. Yes, if we want to make these drop-in replacements for Dense, we should probably support this (but we shouldn't use it in the docstring, since it's a legacy modality). Please also put a comment describing that for posterity :)

Thenerdstation:
@markomernick Once you approve I'll merge this in.

tensornetwork/tn_keras/mpo.py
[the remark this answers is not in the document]
bpenchas: Great - updated the docstring and added a note.

tensornetwork/tn_keras/dense.py
bpenchas: Agreed - at present I get the superclass config on line 139: `config = super(DenseDecomp, self).get_config()` and then add to it. Is there a particular reason to keep `base_config` and `config` separate until the last line? If yes, I'll switch to that approach.

markomernick: The only reason would be to match the expression in other Keras layers - I missed what your code was doing :) Up to you though.

tensornetwork/tn_keras/test_layer.py
bpenchas: Sounds good, added a test for the SavedModel format.

tensornetwork/tn_keras/dense.py
bpenchas: Sounds good! Updated to the idiomatic Keras for posterity :)

tensornetwork/tn_keras/dense.py
```diff
@@ -19,10 +19,12 @@ class DenseDecomp(Layer):
   # as first layer in a sequential model:
   model = Sequential()
   model.add(
-    DenseDecomp(512, decomp_size=128, activation='relu', input_dim=1024))
+    DenseDecomp(512, decomp_size=128, activation='relu', input_shape=(1024,)))
   # now the model will take as input arrays of shape (*, 1024)
```
markomernick, line 23: Minor notes on this comment:

-> input_shape() effectively declares an input to the model, not the layer - it's maybe easier to think about things in that way. You can alternatively avoid the whole question by using Input((1024,) as the first layer in your example :)
-> I wouldn't mention input_dim at all - we shouldn't be encouraging folks to use deprecated APIs.

I think it might be better to put the input_dim note as a comment in the __init__ function, so that we're making the API as simple as possible for new users.

tensornetwork/tn_keras/mpo.py
```diff
@@ -0,0 +1,203 @@
...
+from tensorflow.keras import activations
+from tensorflow.keras import initializers
+from typing import List, Optional, Text, Tuple
+import tensornetwork as tn
+from tensornetwork.network_components import Node
+import numpy as np
+import math
+
+
+class DenseMPO(Layer):
```
markomernick, line 12: You can register this with the Keras object serialization logic using register_keras_serializable(). Then you don't need to add it to the custom object dicts below.

bpenchas: Nice, this makes model loading much cleaner - updated.

tensornetwork/tn_keras/dense.py
bpenchas: Sounds good - I removed mentions of input_dim except for a comment in init.

tensornetwork/tn_keras/mpo.py
```diff
@@ -127,7 +127,7 @@ def is_perfect_root(n, n_nodes):
         trainable=True,
         initializer=self.bias_initializer) if self.use_bias else None
 
-  def call(self, inputs: tf.Tensor) -> tf.Tensor:
+  def call(self, inputs: tf.Tensor, **kwargs) -> tf.Tensor: # pylint: disable=unused-argument
```
markomernick, line 130: Is there a reason for the **kwargs here? The Keras engine should handle not calling this with the wrong argspec.

Thenerdstation: This was added for TF 2.2.0 compatibility as it broke our linter build saying this wasn't there.

markomernick: Hm, that's unfortunate. LGTM, I'll talk to Keras-team about it.

Thenerdstation:
Thanks for all of the help comments!

bpenchas:
Thanks for all the great discussion - really helpful!

tensornetwork/tn_keras/mpo.py
omalleyt12: That's unfortunate, I'd recommend suppressing the original linter warning for now though, and removing the `**kwargs` here. We'll look into why this changed in 2.2

PR #577, opened 2020-05-04
add `jit` to backend

mganahl:


mganahl:
This depends on #570. Once #570 is pulled in, this should only contain small changes

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -397,3 +397,7 @@ def expm(self, matrix: Tensor):
                            x=matrix.shape[0], y=matrix.shape[1]))
     # pylint: disable=no-member
     return jsp.linalg.expm(matrix)
+
+  def jit(self, fun: Callable, *args: List, **kwargs: dict) -> Callable:
+    return libjax.jit(fun, **kwargs)
```
Thenerdstation, line 402: Any reason you're ignoring `args`?

mganahl: my bad, thanks for catching that

Thenerdstation:
Let's add unit tests.

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -530,3 +531,27 @@ def test_matrix_ops_raises(dtype, method):
...
+  res = fun_jit(x, A, y)
+  res.block_until_ready()
+  t2 = time.time()
+
+  res = fun_jit(x, A, y)
+  res.block_until_ready()
+  t3 = time.time()
+  dt2 = t3 - t2
+  dt1 = t2 - t1
+  assert dt2 < dt1
```
Thenerdstation, line 557: Don't try to do timing tests here, its far too flaky. 

A simple test of "The original function and the jitted function return the same result" is good enough. Also, be sure to test with and without keyword arguments since we missed `*args` earlier.

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -5,6 +5,7 @@
 import pytest
 from tensornetwork.backends.jax import jax_backend
 import jax.config as config
+import time
```
Thenerdstation, line 8: Delete this import

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -530,3 +531,18 @@ def test_matrix_ops_raises(dtype, method):
...
+
+  def fun(x, A, y):
+    return jax.numpy.dot(x, jax.numpy.dot(A, y))
+
+  fun_jit = backend.jit(fun)
+  x = jax.numpy.array(np.random.rand(4))
+  y = jax.numpy.array(np.random.rand(4))
+  A = jax.numpy.array(np.random.rand(4, 4))
+  res1 = fun(x, A, y)
+  res2 = fun_jit(x, A, y)
```
Thenerdstation, line 547: Be sure to also add a test that uses key word args.

mganahl: was just about to push :)

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -546,3 +545,22 @@ def fun(x, A, y):
...
+
+
+def test_jit_args_args():
+  backend = jax_backend.JaxBackend()
+
+  def fun(x, A, y):
+    return jax.numpy.dot(x, jax.numpy.dot(A, y))
+
+  fun_jit = backend.jit(fun, static_argnums=(0,))
+  fun_jit_3 = backend.jit(fun, (0,))
```
Thenerdstation, line 557: Wrong function to use keyword args on haha. 

do something like this.

```python

fun_jit = backend.jit(fun, (0,))
res1 = fun_jit(x, A, y)
res2 = fun_jit(x, A=A, y=y)
```

mganahl: ohh, OK got it

PR #450, opened 2020-01-27
ncon should take a BaseBackend object (#387)

jin14:
This pull request addresses[ issue](https://github.com/google/TensorNetwork/issues/387)
- support passing of BaseBackend object in ncon and ncon_network

PR #580, opened 2020-05-05
Jin14 ncon take basebackend object

### LINKED ISSUE #387: ncon should take a BaseBackend object
Since we are now supporting passing `BaseBackend` to `Node` constructors, we should also
allow it in `ncon`

mganahl:
fix PR #450 
fixes #387


Thenerdstation:
Jin4 and martin have already both signed CLAs.

PR #585, opened 2020-05-07
add FiniteTFI, FiniteXXZ and FiniteDMRG to top level imports

mganahl:


mganahl:
Hmm, this is strange. Could it be that our master build is broken?

PR #587, opened 2020-05-07
Fix build

### LINKED ISSUE #586: tensorflow 2.2.0 breaks our build
It seems that tensorflow 2.2.0 breaks our build, at least it does so on my platform.
We should also bump versions of jax and jaxlib in requirements_travix.txt, there was a bug
in jax.numpy.qr on macOS that got only recently fixed.

mganahl:
silenced the linter ...
fixes #586

### DIFF
--- a/tensornetwork/tn_keras/dense.py
+++ b/tensornetwork/tn_keras/dense.py
@@ -1,5 +1,5 @@
 import tensorflow as tf
-from tensorflow.keras.layers import Layer # type: ignore
+from tensorflow.keras.layers import Layer  # type: ignore
 from tensorflow.keras import activations
 from tensorflow.keras import initializers
 from typing import List, Optional, Text, Tuple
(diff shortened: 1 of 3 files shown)


tensornetwork/tn_keras/dense.py
```diff
@@ -71,22 +71,24 @@ def build(self, input_shape: List[int]) -> None:
...
+        initializer=self.kernel_initializer)
     self.bias_var = self.add_weight(
         name='bias',
         shape=(self.output_dim,),
         trainable=True,
         initializer=self.bias_initializer) if self.use_bias else None
     super(DenseDecomp, self).build(input_shape)
 
-  def call(self, inputs: tf.Tensor) -> tf.Tensor:
+  def call(self, inputs: tf.Tensor, **kwargs) -> tf.Tensor:  # pylint: disable=unused-argument
```
Thenerdstation, line 91: Why is this unused?

mganahl: I added **kwargs because tensorflow-keras has it as well.

mganahl: And its also unsed there

PR #591, opened 2020-05-10
Remove duplicates in get_neighbors function

lingxz:


tensornetwork/network_operations.py
```diff
@@ -894,4 +894,4 @@ def get_neighbors(node: BaseNode) -> List[Node]:
         neighbors.append(edge.node2)
       else:
         neighbors.append(edge.node1)
-  return neighbors
+  return list(dict.fromkeys(neighbors))
```
Thenerdstation, line 897: I'm not sure this always gives a deterministic order. Perhaps just do `if edge.node not in neighbors` before appending to the lists.

lingxz: It does for python 3.6+, dict preserves insertion order. 
checking edge.node not in neighbors will be n^2 in number of neighbours, is that fine? or I can create a set for checking duplicates.

Thenerdstation: Looks like you are correct, but the python designers tell us not to rely on it.

http://gandenberger.org/2018/03/10/ordered-dicts-vs-ordereddict/

> Python dict objects now preserve insertion order in the reference CPython implementation — but relying on that property is risky.

I like your set idea. Let's go ahead with that.

lingxz: ahh apparently it's part of the language spec from 3.7 onwards (https://mail.python.org/pipermail/python-dev/2017-December/151283.html, https://docs.python.org/3.7/tutorial/datastructures.html#dictionaries). But I updated it to check duplicates using a set, probably clearer anyway, thanks!

PR #593, opened 2020-05-11
API change for eigs

mganahl:
conform to eigsh_lanczos API


tensornetwork/backends/base_backend.py
```diff
@@ -334,24 +334,31 @@ def eigh(self, matrix: Tensor):
 
   def eigs(self,
            A: Callable,
+           args: List,
```
Thenerdstation, line 337: We should have the default of `args` be `None` and then have the check

```
if args is None:
  args = []
```

That allows us to avoid the annoying `eigs(matvec, args=[])`

mganahl: that's true, I'll fix it

PR #602, opened 2020-05-13
changed gate edge order in base_mps, apply_two_site_gate

sr33dhar:
Currently, the input gate edges are said to be gate_node[2] and gate_node[3], while the output edges are gate_node[0], and gate_node[1].

This is giving me errors, as when I define my gate as a 2x2x2x2 numpy array, and then convert it to a tensor node, the input edges that works for me is gate_node[0], and gate_node[2], with gate_node[1], and gate_node[3] as output.

So hence for my code to work, I need to reorder the edges each time.
I say my code, but I think this applies generally as I just use the 

gate_node = numpy.tensordot(X,Y,axes=0) and 
gate_node = tn.Node(gate_node) 
commands to create my gate, and there is no edge specification from my end.

### DIFF
--- a/tensornetwork/matrixproductstates/base_mps.py
+++ b/tensornetwork/matrixproductstates/base_mps.py
@@ -483,10 +483,10 @@ class BaseMPS:
     node1 = Node(self.tensors[site1], backend=self.backend)
     node2 = Node(self.tensors[site2], backend=self.backend)
     node1[2] ^ node2[0]
-    gate_node[2] ^ node1[1]
-    gate_node[3] ^ node2[1]
-    left_edges = [node1[0], gate_node[0]]
-    right_edges = [gate_node[1], node2[2]]
+    gate_node[0] ^ node1[1]
+    gate_node[2] ^ node2[1]
+    left_edges = [node1[0], gate_node[1]]
+    right_edges = [gate_node[3], node2[2]]
     result = node1 @ node2 @ gate_node
     U, S, V, tw = split_node_full_svd(
         result,


sr33dhar:
@googlebot I signed it!

mganahl:
thanks for the PR! Could you give a brief code example of what gives you errors?
thanks!

Thenerdstation:
From my understanding this is just a usability change. However, as you can see this breaks our tests, so the change isn't going to be that simple.

I'm not sure that I agree that axis 0 and 2 should be the inputs instead of 0 and 1. There are many instances where I describe operations like CNOT like this.

```python
CNOT = np.array(
[[1, 0, 0, 0],
[0, 1, 0, 0],
[0, 0, 0, 1],
[0, 0, 1, 0]]).reshape((2, 2, 2, 2))
```
I understand it's more convenient for defining operations that are from tensor products, but any entangling gate would become much more burdensome.

mganahl:
This change would also violate certain conventions with MPOs that we have right now, so I'd rather keep it as it is.

Thenerdstation:
I'll close this PR in that case, though I think sr33dhar use case shouldn't be ignored. I've opened a new issue for this.

sr33dhar:
@Thenerdstation @mganahl 
Thank you so much for your prompt responses. I am very new to using tensor networks and hence my statements could be wrong!

I am at present implementing a quantum circuit simulator using TN and the for gates needed here, I am defining them as tensor products as stated before. Of course in this picture, entangling gates have to be defined as a sum of tensor products and that is how I am doing so.

I understand that in using 
CNOT = np.array(
[[1, 0, 0, 0],
[0, 1, 0, 0],
[0, 0, 0, 1],
[0, 0, 1, 0]]).reshape((2, 2, 2, 2))
And this requires the inputs to be 0 and 1 [even here, currently the inputs are 2 and 3. Does that make a difference?], would work, but is it guarenteed that the unitary transformation I get from a reshape operation corresponds to the actual CNOT operation?

Because I implimented CNOT as a sum of tensor products, and it does not correspond to the the reshaped 2x2x2x2 tensor.

Thank you again.

PR #606, opened 2020-05-14
Added operator kron

### LINKED ISSUE #603: Outer product for operators
#602 brought up an interesting point of defining operators based on the outerproduct of several smaller tensors. The issue is the normally an operator has its "input axes" defined as the first N axis, and the "output axes" are the last N axes. When dot producting tensors together, the input and output axes are weaved together, forcing a user to do a nasty `reshape` or `reorder_edges` before the tensor can be used for things like the `FiniteMPS` class.

I propose a new method `tn.operator_outer_product(list_of_nodes)` that would do the outer product but maintain the edge order thats expected for operator tensors.

Thenerdstation:
Fixes #603


gevenbly:
I did some quick test of the `operator_kron` using both the numpy and the symmetric backends, and the results were correct for me. However, I thought it strange that it always returned node objects, even when the inputs were tensors. Is this intended behavior?

mganahl:
I think the input and return type should always coincide.

tensornetwork/network_operations.py
```diff
@@ -899,3 +899,34 @@ def get_neighbors(node: BaseNode) -> List[Node]:
...
+    A `Node` that is the kronecker product of the given inputs. The first
+    half of the edges of this node would represent the "input" edges of the
+    operator and the last half of edges are the "output" edges of the
+    operator.
+  """
+  nodes = []
+  input_edges = []
+  output_edges = []
+  for tensor in tensors:
+    if isinstance(tensor, BaseNode):
```
mganahl, line 919: I think Node() also accepts `Node` as input, so the check here could probably be removed

Thenerdstation:
> I think the input and return type should always coincide.

That would break the pattern that all of existing ops already have, which is basically node in node out. The only non-backend method that takes in tensors and returns tensors is `ncon`, for a good reason as users using `ncon` may never actually want to deal with `Node`s.

I think instead I'll just have this method only accept `Node`s. Users can always wrap their tensors in `Node`s before calling this function. (If a user wants just tensors, they can always just do `np.kron` or equivalent.)

tensornetwork/network_operations.py
```diff
@@ -899,3 +899,34 @@ def get_neighbors(node: BaseNode) -> List[Node]:
...
+      node = tensor
+    else:
+      node = Node(tensor)
+    order = len(node.shape)
+    if order % 2 != 0:
```
mganahl, line 919: that seems weird to me. Can we rrelax this?

Thenerdstation: How would you differentiate then from input and output legs?

mganahl: Yeah, only by passing an argument, but then you're back to outer_product_final_nodes...
let's leave it

mganahl:
> > I think the input and return type should always coincide.
> 
> That would break the pattern that all of existing ops already have, which is basically node in node out. The only non-backend method that takes in tensors and returns tensors is `ncon`, for a good reason as users using `ncon` may never actually want to deal with `Node`s.
> 
> I think instead I'll just have this method only accept `Node`s. Users can always wrap their tensors in `Node`s before calling this function. (If a user wants just tensors, they can always just do `np.kron` or equivalent.)

That would be my choice as well

PR #611, opened 2020-05-18
Changed internal structure of Edge class to allow for hyper edges

Thenerdstation:
First of many PRs to support hyper edges.

Next step is to make all methods no longer rely on `edge.node1` and `edge.node2`. This will take several PRs.

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -1011,7 +1011,7 @@ class Edge:
     self._axes[0] = axis1
 
   @property
-  def axis2(self) -> int:
+  def axis2(self) -> Optional[int]:
     if self.is_disabled:
       raise ValueError(
           'Edge has been disabled, accessing axis2 is no longer possible')
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -964,11 +964,9 @@ class Edge:
     else:
       if not isinstance(name, str):
         raise TypeError("Edge name should be str type")
+    self._nodes = [node1, node2]
+    self._axes = [axis1, axis2]
     self._name = name
-    self.node1 = node1
-    self._axis1 = axis1
-    self.node2 = node2
-    self._axis2 = axis2
     self._is_dangling = node2 is None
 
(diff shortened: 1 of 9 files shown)


tensornetwork/network_components.py
```diff
@@ -1074,25 +1069,25 @@ def node2(self) -> Optional[BaseNode]:
...
   def node1(self, node: BaseNode) -> None:
     if self.is_disabled:
       raise ValueError(
           'Edge has been disabled, setting node1 is no longer possible')
     # pylint: disable=attribute-defined-outside-init
-    self._node1 = node
+    self._nodes[0] = node
 
   @node2.setter
   def node2(self, node: Optional[BaseNode]) -> None:
```
bpenchas, line 1085: To support arbitrary number of nodes, seems like it would be cleaner to have a single getter and a single setter that takes the node number as an argument.

tensornetwork/tests/network_components_free_test.py
```diff
@@ -1457,7 +1457,7 @@ def test_edge_name_setter_throws_type_error(single_node_edge, name):
 
 def test_edge_node1_throws_value_error(single_node_edge):
   edge = Edge(node1=single_node_edge.node, axis1=0, name="edge")
-  edge._node1 = None
+  edge.node1 = None
```
bpenchas, line 1460: Should be `edge._nodes[0]` to match new design and `test_edge_node2_throws_value_error` below.

tensornetwork/network_components.py
Thenerdstation: Good point,for now this is just to allow the existing code to work. Ultimately, I plan on completely removing the `axis1, axis2, node1, node2` entirely. This will take many PRs as I basically have to modify all of the existing `Edge` logic.

And that getter and setter that you mentioned will basically just be what the `update_axis` method currently is.

tensornetwork/tests/network_components_free_test.py
Thenerdstation: Good point. Added.

PR #612, opened 2020-05-20
added option for QR decomposition if there is no truncation in apply_two_qubit_gate

sr33dhar:
I am using the tensornetwork package for quantum circuit simulation, QAOA to be specific, and need to repeatedly apply two-qubit gates, most of the times without truncation.

Speed optimisation is my primary goal, and since doing a full SVD when there is no truncation is wasteful, I would prefer having an option for QR decomposition in this case.

I tested this for a couple of my simulations, and it seems to be working.
Please consider this edit.
Thank you.


tensornetwork/matrixproductstates/base_mps.py
```diff
@@ -471,37 +471,61 @@ def apply_two_site_gate(self,
...
+                max_truncation_err=max_truncation_err,
+                left_name=node1.name,
+                right_name=node2.name)
+            V.reorder_edges([S[1]] + right_edges)
+            left_edges = left_edges + [S[1]]
+            res = contract_between(U, S, name=U.name).reorder_edges(left_edges)
+            self.tensors[site1] = res.tensor
+            self.tensors[site2] = V.tensor
+            return tw
+
```
Thenerdstation, line 508: We use 2 space tabing.

tensornetwork/matrixproductstates/base_mps.py
```diff
@@ -471,37 +471,61 @@ def apply_two_site_gate(self,
...
+        
+        gate_node = Node(gate, backend=self.backend)
+        node1 = Node(self.tensors[site1], backend=self.backend)
+        node2 = Node(self.tensors[site2], backend=self.backend)
+        node1[2] ^ node2[0]
+        gate_node[2] ^ node1[1]
+        gate_node[3] ^ node2[1]
+        left_edges = [node1[0], gate_node[0]]
+        right_edges = [gate_node[1], node2[2]]
+        result = node1 @ node2 @ gate_node
```
Thenerdstation, line 519: This code is duplicated from above. Can you restructure the `if/else` statements to avoid the duplicate code?

tensornetwork/matrixproductstates/base_mps.py
```diff
@@ -471,37 +471,61 @@ def apply_two_site_gate(self,
...
+        Q, R = split_node_qr(
```
Thenerdstation, line 520: Based on the above code, the singular values are pulled into the `U` tensor, making the right side unitary. To keep that consistency, we should actually use `split_node_rq` instead.

tensornetwork/matrixproductstates/base_mps.py
```diff
@@ -471,37 +471,61 @@ def apply_two_site_gate(self,
...
+            result,
+            left_edges=left_edges,
+            right_edges=right_edges,
+            left_name=node1.name,
+            right_name=node2.name)
+        self.tensors[site1] = Q.tensor
+        self.tensors[site2] = R.tensor
+        return None
```
Thenerdstation, line 528: Can you return `backend.convert_to_tensor([])` instead?

Thenerdstation:
Gonna leave it up to @mganahl as to whether we want to do this or not.

tensornetwork/matrixproductstates/base_mps.py
```diff
@@ -471,37 +471,61 @@ def apply_two_site_gate(self,
...
+    if (max_singular_values or max_truncation_err): 
+    
+        if (self.center_position not in (site1, site2)):
+            raise ValueError(
+              'center_position = {}, but gate is applied at sites {}, {}. '
+              'Truncation should only be done if the gate '
+              'is applied at the center position of the MPS'.format(
+                  self.center_position, site1, site2))
+            
+        else:
```
mganahl, line 483: that else can go (the clause returns)

tensornetwork/matrixproductstates/base_mps.py
```diff
@@ -471,37 +471,61 @@ def apply_two_site_gate(self,
...
+        right_edges = [gate_node[1], node2[2]]
+        result = node1 @ node2 @ gate_node
+        Q, R = split_node_qr(
+            result,
+            left_edges=left_edges,
+            right_edges=right_edges,
+            left_name=node1.name,
+            right_name=node2.name)
+        self.tensors[site1] = Q.tensor
+        self.tensors[site2] = R.tensor
```
mganahl, line 527: you need to set center_position accordinlgy

tensornetwork/matrixproductstates/base_mps.py
sr33dhar: Ah I see. Thank you

tensornetwork/matrixproductstates/base_mps.py
sr33dhar: But the current code does not alter the center_position, does it?
Doesn't applying gates destroy any canonical form that the states may have?

Because even previously, the contracted tensor between U, and S is not guaranteed to be unitary, right?

sr33dhar:
@googlebot I fixed it

tensornetwork/matrixproductstates/base_mps.py
mganahl: that's actually correct, thanks for noticing. It would be great if you could fix this as well!

mganahl: Make sure that the center_position doesn't chagne

sr33dhar: Personally, I would like to keep this as is.
For circuit simulation where my end goal is calculating expectation values, canonicality is irrelevant. Preserving canonicality would mean extra steps for me, and when trying to simulate large circuits, this can be costly in my code.

If the user wishes to preserve center_position, they can always use the self.position() command, right?

Also, the description states  "Apply a two-site gate to an MPS. This routine will in-general destroy any canonical form of the state. If a canonical form is needed, the user can restore it using `FiniteMPS.position`."

mganahl: well, it doesn't cost you anything to reset center_position, does it?

mganahl: i think that would be acceptable

sr33dhar: Or would you like me to accept if the user would like to preserve canonicality as an argument and then code accordingly?

Something like,
  def apply_two_site_gate(self,
                          gate: Tensor,
                          site1: int,
                          site2: int,
                          max_singular_values: Optional[int] = None,
                          max_truncation_err: Optional[float] = None
                          canonical[bool] = False) -> Tensor:

sr33dhar: I am trying to simulate a truncated 50 qubit QAOA circuit about 10000 times.
Resetting center_position would involve a whole bunch of QR decompositions*10000, right?
I am not sure if that would be so inexpensive.

And could you please clarify?
By "that would be acceptable", did you mean my suggestion of accepting canonical as a function argument?
Thank you.

tensornetwork/matrixproductstates/base_mps.py
sr33dhar: done

tensornetwork/matrixproductstates/base_mps.py
sr33dhar: done

tensornetwork/matrixproductstates/base_mps.py
sr33dhar: done

tensornetwork/matrixproductstates/base_mps.py
mganahl: sorry for the ambiguous comment above. I was referring to my suggestion to reset `center_position`. 

For clarification: I assume you are simulating a low-depth quantum circuit for 50 qubits? The bond-dimension grows exponentially with the number of gate-applications, so without truncation this becomes quickly infeasible.  I also assume you are simulating nearest-neighbor gates, otherwise things will become substantially more expensive.

If possible I think we should keep the center-position consistent with expectations, namely that all tensors to the left of `center_position` are left orthogonal and all tensors to the right are right-orthogonal. 

Here is a more detailed description of my suggested changes:
1. We allow `center_position` to be `None`, indicating that the state is not in a canonical form. 
2. In `apply_two_site_gate` we check if `center_position` is `None`, and if so we merely apply the gate, doing either SVD or QR. In the SVD case with truncation we should at least raise a warning (preferably an error) because truncation is suboptimal if the state is not in a canonical form. 

3. For the case that `center_position` has a value different from `None` we do the following: 
 - if the gate overlaps with the center position, we adjust the value of `self.center_position` such that after the gate application the state is consistent as in the above-described sense. We can do this by letting the user pass a variable that lets him decide where to put the orthogonality center. If he doesn't we do the SVD or QR decomp such that the value of `center_position` is unchanged. 
 - if the gate does not overlap with the center position, we apply the gate (again raising a warning or an error if truncation is performed) and set `center_position=None`. 
4. We clearly document this behaviour in the docstring.

would this solve your problem?

sr33dhar: Clarification:

1.) Yes, I am trying to simulate a shallow depth 50 qubit QAOA circuit with truncated bond dimensions.
2.) No, I am not simulating nearest neighbour gates and am using SWAP gates to implement non-nearest neighbour gates

This suggestion would solve my problem.
@mganahl Could you please go through the image in the link below to confirm if I have got this right?
[https://drive.google.com/open?id=1D_4L7Ryd8zJ4MZT6NXN01yUGsupM_Xfe](url)

But for that, we will also need to add some changes to apply_one_site_gate function as currently, it destroys canonicality, but still, self.center_position is shown to be an integer and not None.

I will implement this over the weekend and get back to you.

mganahl: it seems the link you sent is broken.
Yes, this would require changes in more places (i.e. constructors of the mps classes as well). I think the best option here is that I will implement this, if that's OK for you. I'll try to come back to it as quickly as possible.

mganahl: actually it works now

mganahl: yes, the image shows the intended behaviour

mganahl: You can make these changes to your PR and I will take care of the rest

sr33dhar: Great, 
Maybe then I could make these changes into the apply_two_site_gate and apply_one_site_gate functions, and you can take it from there?

I already started with it, and it shouldn't take long.

sr33dhar: Hey @mganahl
Does this work?
I made the changes specified in the figure

sr33dhar: Just saw that there are a few errors.
Will fix that over the weekend.

sr33dhar: Hey Martin,
Any updates on this?
Thank you.
@mganahl

mganahl: Hi @sr33dhar sorry for the long silence. I have updated the MPS classes as we discussed. Once you have fixed the errors on this PR we can pull it in

mganahl:
addressed in #913

PR #617, opened 2020-05-21
fixing transposition in add and sub

### LINKED ISSUE #614: Unintended lazy transposition in blocksparse
Hi @mganahl , I have encountered some weird behavior with the transposition in blocksparse. Consider the following code:
```
ham = BlockSparseTensor.random([i0,i0,i0,i1,i1,i1]) 
hamT = ham.conj().transpose([3,4,5,0,1,2]) 
```
If I check the orders then the following is returned:
```
# ham._order == [[0],[1],[2],[3],[4],[5]]
# hamT._order == [[3],[4],[5],[0],[1],[2]]
```
As expected, this indicates that `hamT` is subject to a lazy transposition (and `ham` is not). However if I now do:
```
matrix = ham + hamT
```
Then when I recheck the order I now get:
```
# ham._order == [[3],[4],[5],[0],[1],[2]]
```
So, in when adding the two tensors together, blocksparse has done a transposition of `ham` in memory, even though `ham` was never a lazily transposed tensor. I think I understand why this is the case, but do not think that this should be allowed behavior from blocksparse. One reason is that it could result in the code transposing tensors in memory much more than is necessary. And on a conceptual level, I do not think blocksparse should be doing transpositions (in memory) on tensors where no transposition call was ever issued (i.e. the lazy transpose should delay evaluation of a `.transpose` operation until necessary, but should not decide on its own to start shifting in memory tensors where `.transpose` was never called to begin with).

I think that binary operations on tensors, e.g. `ham + hamT`, should instead ask: has any tensor been subjected to a lazy transposition that has not yet been evaluated? If yes, then do the transposition explicitly before performing the binary operation.

mganahl:
fixes #614

### DIFF
--- a/tensornetwork/block_sparse/blocksparsetensor.py
+++ b/tensornetwork/block_sparse/blocksparsetensor.py
@@ -671,17 +671,6 @@ class BlockSparseTensor(ChargeArray):
         order=self._order,
         check_consistency=False)
 
-    # perm = np.empty(len(self.flat_order), dtype=np.int32)
-    # perm[other.flat_order] = self.flat_order
-    # self.transpose_data(perm, inplace=True)
-    # #now addition is save
-    # return BlockSparseTensor(
-    #     data=self.data + other.data,
-    #     charges=self._charges,
-    #     flows=self._flows,
-    #     order=self._order,
-    #     check_consistency=False)
-
   def __mul__(self, number: np.number) -> "BlockSparseTensor":
     if not np.isscalar(number):
       raise TypeError(
--- a/tensornetwork/block_sparse/blocksparsetensor.py
+++ b/tensornetwork/block_sparse/blocksparsetensor.py
@@ -626,13 +626,33 @@ class BlockSparseTensor(ChargeArray):
       raise ValueError(
           "cannot add or subtract tensors non-matching sparse shapes")
 
+  def _align_storage_layout(self, other):
+    """
+    Align storage layouts of self and other.
+    """
+    self_is_ordered = np.array_equal(self.flat_order,
+                                     np.arange(len(self.flat_order)))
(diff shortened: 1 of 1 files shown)


tensornetwork/block_sparse/blocksparsetensor.py
```diff
@@ -626,13 +626,33 @@ def _sub_add_protection(self, other):
...
+    self_is_ordered = np.array_equal(self.flat_order,
+                                     np.arange(len(self.flat_order)))
+    other_is_ordered = np.array_equal(other.flat_order,
+                                      np.arange(len(other.flat_order)))
+    if self_is_ordered and (not other_is_ordered):
+      #bring other into the same storage layout as other
+      perm = np.empty(len(other.flat_order), dtype=np.int32)
+      perm[self.flat_order] = other.flat_order
+      other.transpose_data(perm, inplace=True)
+    elif (not self_is_ordered) and other_is_ordered:
```
Thenerdstation, line 642: This and the `else` below appear to be the same code. Can you just make it a single `else`?

mganahl: They are, I thought it might be easier to parse for the reader this way, this is why I split it up

mganahl: I'll fix it!

mganahl:
> Is there a reason the transposition is happening in place? Will this have a negative effect of this BlockSparseTensor is reused later?

The transpose operation of BlockSparseTensor is lazy, meaning that only the meta data is changed but the data in memory is actually not reshuffled. `transpose_data` is only shuffling the data in the memory without changing the appearance of the object to the users (that is, unless he decides to poke around in the private attributes).

Thenerdstation:
Great! That is what I wanted to hear.

PR #619, opened 2020-05-21
Fix flat flows

### LINKED ISSUE #613: Flat flows and flat charges in blocksparse
Hi @mganahl, I believe there are some issues with `.flows` `._flows` and `.flat_flows` in blocksparse. Consider the following example:
 
```
# define tensor
i1 = Index(U1Charge([-1,0,1]), flow=False)
i2 = Index(U1Charge([-1,0,0,1]), flow=False)
i3 = Index(U1Charge([-2,2]), flow=True)
i4 = Index(U1Charge([-2,0,2]), flow=True)
i5 = Index(U1Charge([-2,0,0,2]), flow=True)
T0 = BlockSparseTensor.random([i0,i1,i2,i3,i4,i5]) 
# do transposition
T1 = T0.transpose([2,3,0,1,5,4]).reshape([8,2,3,12])
```
Then when I examine `T1` the following is returned:
```
# T1._order == [[2, 3], [0], [1], [5, 4]]
# T1.flat_order == [2,3,0,1,5,4]

# T1.flows == [[False, True], [False], [False], [True, True]]
# T1._flows == [False, False, False,  True,  True,  True]
# T1.flat_flows == [False, False, False, True, True, True]

# T1.charges == [[i2,i3],i0,i1,[i5,i4]]
# T1._charges == [i0,i1,i2,i3,i4,i5]
# T1.flat_charges == [i0,i1,i2,i3,i4,i5]
```
So `T1._flows` and `T1.flat_flows` are exactly the same (similarly for the charges); they both give the flows before the lazy transposition and reshape. Is this intended behavior? It seems that it should instead be:
```
# T1.flat_flows == [False, True, False, False, True, True]
# T1.flat_charges == [i2,i3,i0,i1,i5,i4]
```
such that  `T1.flat_flows` is a flattening of  `T1.flows` rather than of `T1._flows`, which would be a useful attribute to have defined.

mganahl:
flat_flows and flat_charges now return flows and charges in the current index-order of the tensor instead of the actual index order corresponding to the storage layout.
fixes #613

### DIFF
--- a/tensornetwork/block_sparse/blocksparsetensor.py
+++ b/tensornetwork/block_sparse/blocksparsetensor.py
@@ -671,17 +671,6 @@ class BlockSparseTensor(ChargeArray):
         order=self._order,
         check_consistency=False)
 
-    # perm = np.empty(len(self.flat_order), dtype=np.int32)
-    # perm[other.flat_order] = self.flat_order
-    # self.transpose_data(perm, inplace=True)
-    # #now addition is save
-    # return BlockSparseTensor(
-    #     data=self.data + other.data,
-    #     charges=self._charges,
-    #     flows=self._flows,
-    #     order=self._order,
-    #     check_consistency=False)
-
   def __mul__(self, number: np.number) -> "BlockSparseTensor":
     if not np.isscalar(number):
       raise TypeError(
--- a/tensornetwork/block_sparse/blocksparsetensor.py
+++ b/tensornetwork/block_sparse/blocksparsetensor.py
@@ -160,7 +160,7 @@ class ChargeArray:
 
   @property
   def flat_flows(self) -> List:
-    return list(self._flows)
+    return list([self._flows[o] for o in self.flat_order])
 
   @property
   def flat_order(self) -> np.ndarray:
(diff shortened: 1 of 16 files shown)


Thenerdstation:
Fix conflicts

tensornetwork/block_sparse/blocksparsetensor.py
```diff
@@ -156,11 +156,11 @@ def flows(self) -> List[List]:
 
   @property
   def flat_charges(self) -> List[BaseCharge]:
-    return self._charges
+    return list([self._charges[o] for o in self.flat_order])
```
Thenerdstation, line 159: You don't need to wrap this in a list since [...] Already makes it a list.

tensornetwork/block_sparse/blocksparsetensor.py
```diff
@@ -156,11 +156,11 @@ def flows(self) -> List[List]:
...
 
   @property
   def flat_flows(self) -> List:
-    return list(self._flows)
+    return list([self._flows[o] for o in self.flat_order])
```
Thenerdstation, line 163: Same here. Remove the list keyword.

PR #623, opened 2020-05-26
Added jitted ncon.

Thenerdstation:
Also removed the jit of the tensorflow backend because tf.function is bad. It does far too much hidden magic and we shouldn't continue to support it.


tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -249,4 +249,5 @@ def expm(self, matrix: Tensor) -> Tensor:
     return tf.linalg.expm(matrix)
 
   def jit(self, fun: Callable, *args: List, **kwargs: dict) -> Callable:
-    return tf.function(fun, *args, **kwargs)
+    # tf.function is slow and bad.
```
mganahl, line 252: maybe add a comment to the dosctring of BaseCharge explaining that jit only works for jax

Thenerdstation: Good idea.

PR #631, opened 2020-05-29
Isscalar

mganahl:
Fixes a bug with block-sparse backend and ncon
fixes 615 partially


Thenerdstation:
Seems ok, fix the build

PR #630, opened 2020-05-28
Create the linalg directory, implement and test eye.

alewis:

PR #632, opened 2020-05-29
Initialization

alewis:


mganahl:
Looks like it has the same issues as #630. I'll have a look once the tests pass

PR #634, opened 2020-06-01
Blocksparse trace

### LINKED ISSUE #615: Blocksparse not evaluating scalar products properly
Hi @mganahl , I am having issues when trying to contract networks that evaluate to a scalar. For example defining:
```
i1 = Index(U1Charge([-1,1]), flow=True)
T0 = BlockSparseTensor.random([i0,i1]) 
```
then evaluating `ncon([T0,T0],[[1,2],[2,1]])` produces an error. It this specific instance, I can instead evaluate the network as `c = trace(T0 @ T0)`, however `c` is returned as some kind of weird rank-0 blocksparse tensor rather than a scalar as expected. Evaluating `c.todense()` produces an empty array, but taking `c.data` gives the correct scalar (although the .data field is not something the user should ever have to touch). Can blocksparse `trace` be fixed to return a scalar? And also blocksparse `ncon` fixed to work with closed tensor networks? (i.e networks that contain no dangling edges)

On a related note, I see that the numpy version of `ncon` returns a length-1 numpy array when contracting closed tensor networks. I would strongly advocate that the return from evaluating a closed tensor network (no dangling edges) should always be a plain scalar, regardless of the types of the tensors involved.

mganahl:
fix block-sparse trace to return a number for trace of matrices
fixes #615

### DIFF
--- a/tensornetwork/block_sparse/blocksparsetensor.py
+++ b/tensornetwork/block_sparse/blocksparsetensor.py
@@ -395,7 +395,6 @@ class ChargeArray:
     if self.ndim == 0:
       if len(self.data) == 1:
         return self.data[0]
-      return self.dtype.type(0.0)  #the default value is currently 0.0
     raise ValueError("can only convert an array of size 1 to a Python scalar")
 
 
--- a/tensornetwork/block_sparse/blocksparsetensor.py
+++ b/tensornetwork/block_sparse/blocksparsetensor.py
@@ -391,6 +391,13 @@ class ChargeArray:
   def __truediv__(self, number: np.number) -> "ChargeArray":
     raise NotImplementedError("__truediv__ not implemented for ChargeArray")
 
+  def item(self):
+    if self.ndim == 0:
+      if len(self.data) == 1:
+        return self.data[0]
+      return self.dtype.type(0.0)  #the default value is currently 0.0
+    raise ValueError("can only convert an array of size 1 to a Python scalar")
+
 
 class BlockSparseTensor(ChargeArray):
   """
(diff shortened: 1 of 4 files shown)


tensornetwork/block_sparse/blocksparsetensor.py
```diff
@@ -391,6 +391,13 @@ def __rmul__(self, number: np.number) -> "ChargeArray":
   def __truediv__(self, number: np.number) -> "ChargeArray":
     raise NotImplementedError("__truediv__ not implemented for ChargeArray")
 
+  def item(self):
+    if self.ndim == 0:
+      if len(self.data) == 1:
+        return self.data[0]
+      return self.dtype.type(0.0)  #the default value is currently 0.0
```
Thenerdstation, line 398: Why do we have a default here?

tensornetwork/block_sparse/blocksparsetensor_test.py
```diff
@@ -737,3 +737,20 @@ def test_flat_charges():
...
+      flows=[],
+      order=[],
+      check_consistency=False)
+  assert t1.item() == 1
+  Ds = [10, 11, 12, 13]
+  charges = [U1Charge.random(Ds[n], -5, 5) for n in range(4)]
+  flows = [True, False, True, False]
+  inds = [Index(c, flows[n]) for n, c in enumerate(charges)]
+  t2 = BlockSparseTensor.random(inds, dtype=np.float64)
+  with pytest.raises(ValueError):
```
Thenerdstation, line 755: be sure to include `match=...` on all raises to make sure you hit the right one.

tensornetwork/block_sparse/blocksparsetensor.py
mganahl: that's the default value for tensor elements that do not conserve charge. In fact, I think that this line should never be hit, so I'll remove it

PR #636, opened 2020-06-01
Add linalg directory and initialization code.

alewis:


tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+    fname:  Name of the method of backend to call (a string).
+    *fargs: Positional arguments to the  initialization method.
+    name: Optional name of the Node.
+    axis_names: Optional names of the Node's dangling edges.
+    backend: The backend or its name.
+    **fkwargs: Keyword arguments to the initialization method.
+
+  Returns:
+    node: A Node wrapping data generated by
+          (the_backend).fname(*fargs, **kwargs), with one dangling edge per
```
mganahl, line 54: typo

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+                             dtype=dtype, M=M)
+  return the_node
+
+
+def zeros(shape: Sequence[int],
+          dtype: Optional[Type[np.number]],
+          name: Optional[Text] = None,
+          axis_names: Optional[List[Text]] = None,
+          backend: Optional[Union[Text, BaseBackend]] = None) -> Tensor:
+  """Return a Node of shape 'shape' of all zeros.
```
mganahl, line 100: use `shape` instead of 'shape'

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+        ones.
+  """
+  the_node = initialize_node("eye", N,
+                             name=name, axis_names=axis_names, backend=backend,
+                             dtype=dtype, M=M)
+  return the_node
+
+
+def zeros(shape: Sequence[int],
+          dtype: Optional[Type[np.number]],
```
mganahl, line 96: should this have a default value?

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+    backend_obj = backends.backend_factory.get_backend(backend)
+  func = getattr(backend_obj, fname)
+  data = func(*fargs, **fkwargs)
+  node = network_components.Node(data, name=name, axis_names=axis_names,
+                                 backend=backend)
+  return node
+
+
+def eye(N: int,
+        dtype: Optional[Type[np.number]],
```
mganahl, line 69: default value?

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+        Represents an array of all zeros.
+  """
+  the_node = initialize_node("zeros", shape,
+                             name=name, axis_names=axis_names, backend=backend,
+                             dtype=dtype)
+  return the_node
+
+
+def ones(shape: Sequence[int],
+         dtype: Optional[Type[np.number]],
```
mganahl, line 119: default

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+         name: Optional[Text] = None,
+         axis_names: Optional[List[Text]] = None,
+         backend: Optional[Union[Text, BaseBackend]] = None) -> Tensor:
+  """Return a Node of shape 'shape' of all ones.
```
mganahl, line 123: quotes

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+        Represents an array of all ones.
+  """
+  the_node = initialize_node("ones", shape,
+                             name=name, axis_names=axis_names, backend=backend,
+                             dtype=dtype)
+  return the_node
+
+
+def randn(shape: Sequence[int],
+          dtype: Optional[Type[np.number]],
```
mganahl, line 142: default

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+          seed: Optional[int] = None,
+          name: Optional[Text] = None,
+          axis_names: Optional[List[Text]] = None,
+          backend: Optional[Union[Text, BaseBackend]] = None) -> Tensor:
+  """Return a Node of shape 'shape' of Gaussian random floats.
```
mganahl, line 147: quotes

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+        Represents an array of all ones.
+  """
+  the_node = initialize_node("randn", shape,
+                             name=name, axis_names=axis_names, backend=backend,
+                             seed=seed, dtype=dtype)
+  return the_node
+
+
+def random_uniform(shape: Sequence[int],
+                   dtype: Optional[Type[np.number]],
```
mganahl, line 167: dtype

tensornetwork/linalg/linalg.py
alewis: That seems reasonable, yes. What should it be?

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+                   seed: Optional[int] = None,
+                   boundaries: Optional[Tuple[float, float]] = (0.0, 1.0),
```
mganahl, line 169: apparently `Optional` means that it can be `None` as well. If that is not possible then  remove `Optional`

tensornetwork/linalg/linalg.py
mganahl: I would say double precision floats

tensornetwork/linalg/linalg_test.py
```diff
@@ -0,0 +1,105 @@
...
+  name = "Jeffrey"
+  axis_names = ["Sam", "Blinkey"]
+  backend_list = ["jax", "numpy"]
+  for backend in backend_list:
+    for dtype in np_dtypes + np_int_dtypes + np_uint_dtypes:
+      tnI = linalg.eye(N, dtype=dtype, M=M, name=name, axis_names=axis_names,
+                       backend=backend)
+      npI = np.eye(N, M=M, dtype=dtype)
+      np.testing.assert_allclose(tnI.tensor, npI)
+
```
mganahl, line 33: also check if axis names are as expected

mganahl: and `name`

mganahl: and backend

mganahl: also in all other tests

mganahl: to test all backends you can use a @pytest.mark.parametrize(backend=['numpy','jax','tensorflow','pytorch']), or just use the backend fixture `backend`, i.e.

```python 
test_eye(backend): #the variable `backend` is defined by the pytest fixture
  ...

```

mganahl: In general we always test all backends (except `symmetric`)

tensornetwork/linalg/linalg.py
```diff
@@ -0,0 +1,191 @@
...
+from tensornetwork import network_components
+
+
+Tensor = Any
+BaseBackend = base_backend.BaseBackend
+
+
+# INITIALIZATION
+def initialize_node(fname: Text,
+                    *fargs: Any,
```
Thenerdstation, line 33: Looks like you never actually use these positional args. Can you remove it in favor of just always using `fkwargs`?

alewis: I use them to feed in the shape in a couple of places, but it's easy enough to make that a keyword argument if you like that more

Thenerdstation: Yeah that sounds good. More readable that way too.

alewis: Ah, I remember why I did this now; it's because in certain cases the backend function expects certain arguments to be positional. If only `fkwargs` is used I don't know how to determine which are which.

Thenerdstation: Aw, ok no worries then. Keep it as-is.

tensornetwork/linalg/linalg.py
alewis: Do you mean the word shape or the Markdown-formatted `shape`

tensornetwork/linalg/linalg.py
alewis: The default is now None, under the (tested) assumption that all backends provide a default value in this case.

mganahl: why not np.float64?

mganahl: then it would be clearer for the user

alewis: Will this be valid for all backends (notably pytorch)?

alewis: It also strikes me that the "best default" ought to be located at the backend; one could imagine that different backends might prefer different dtypes.

mganahl: I see, passing numpy dtypes using the backend initialization is problematic, yes. The default value though should be set within the top level function, and the user should be able to know what it is without looking into the backend documentation. Otherwise this is could be a source of really obscure bugs.

alewis: Well, I guess I could just have a big if statement or something. For that to work, Pytorch would need to be imported at the top level, which would make it a dependency of the interface. Is that something we want?

mganahl: No we don't want that. My suggestion is to use numpy dtypes and convert them to the backend dtypes within the backend. Maybe I'm too picky, I just think that hidden choices about dtypes are really annoying.

alewis: I'm fine with that, but then wouldn't the user need to consult the backend to figure out that the conversion had occurred?

mganahl: The idea is to use numpy to configure the dtypes (i.e. half vs single vs double), and convert this to the corresponding backend dtypes, (ie.e. np.floa64 will be converted to jax.numpy.float64, or torch.float64)

alewis: We could give each backend a member "backend.default_dtype". Then, it would be obvious that each backend might do something different (rather than users perhaps assuming that their Nodes literally store np.float64), and while one would still need to look up the backend documentation to figure out what the default was, at least there would be only one clear place to look.

alewis: @Thenerdstation

alewis: > The idea is to use numpy to configure the dtypes (i.e. half vs single vs double), and convert this to the corresponding backend dtypes, (ie.e. np.floa64 will be converted to jax.numpy.float64, or torch.float64)

This also seems like a reasonable solution, provided it be documented appropriately.

alewis: Jax, NumPy, and TensorFlow already behave in this way, so it seems only Pytorch would need to be modified.

Thenerdstation: My opinion is to leave it as `None` and just stick to the default of the given backend, as that default option will usually be the most preformant choice for that backend.

I doubt there would be many "obsure bugs" related to a bad dtype. Infact, its actually more likely that the obsure bug would be a slow down do to defaulting to float64 instead of float32 for pytorch since float32 runs significantly faster on GPU hardware.

My preference is to leave it as `None` instead of adding a default dtype to the backends that the user can't control.

alewis: In this case, pending Martin's agreement, all issues with this PR appear to be resolved

mganahl: I consider myself to be overuled :)

PR #637, opened 2020-06-01
improve Blocksparse __repr__ 

### LINKED ISSUE #616: Blocksparse REPL problems
Hi @mganahl , there appears to be some issues with the REPL w.r.t the Index class. For instance if I define:
```
```
Then if I either evaluate `i0` in the REPL or try to `print(i0)` the return is simply `2`. So it appears that the REPL is set to return `i0.dim` when called, but I found this to be very confusing (I thought that I had made a mistake somewhere, and that `i0` was assigned to the integer value 2). Maybe this is not such a big deal if the Index class is going to be removed or hidden from the user, as we discussed could be an option.

Also, on a related note, it would be a great help if the REPL and print() for blocksparse tensors was more informative. At the moment only the class name is returned, but it would be useful to see the rank of the tensor, the dimension of its indices, its flows, its symmetry group(s) etc.

mganahl:
fixes #616


tensornetwork/block_sparse/blocksparsetensor.py
```diff
@@ -391,6 +391,27 @@ def __rmul__(self, number: np.number) -> "ChargeArray":
...
+    if len(self._charges) > 0:
+      charge_types = self._charges[0].names
+    else:
+      charge_types = 'no charge types (scalar)'
+    output = 'BlockSparseTensor\n   shape: ' + repr(
+        self.shape
+    ) + '\n   charge types: ' + charge_types + '\n   dtype: ' + repr(
+        self.dtype.name) + '\n   flat flows: ' + repr(
+            self.flat_flows) + '\n   order: ' + repr(
+                self._order) + '\n   data:' + repr(self.data)
```
Thenerdstation, line 404: Do you have an example of what this prints out?

mganahl: yes, looks like this:
```
BlockSparseTensor
   shape: (10, 10)
   charge types: ['U1Charge', 'Z2Charge']
   dtype: 'float64'
   flat flows: [True, False]
   order: [[0], [1]]
   data:array([0.36553923, 0.14096483, 0.74018218, 0.42214647, 0.0460793 ,
       0.03905852, 0.57348562, 0.47659091, 0.12239714, 0.01138399,
       0.47512114, 0.07621713, 0.02356647, 0.60349018])

```

Thenerdstation: beautiful!

mganahl: thanks, I  put some effort in it ;)

tensornetwork/block_sparse/charge.py
```diff
@@ -528,10 +528,26 @@ def isin(self, target_charges: Union[np.ndarray, "BaseCharge"]) -> np.ndarray:
...
+               charge_labels: Optional[np.ndarray] = None,
+               charge_types: Optional[List[Type["BaseCharge"]]] = None,
+               charge_dtype: Optional[Type[np.number]] = np.int16) -> None:
+
+    super().__init__(
+        charges,
+        charge_labels,
+        charge_types=[type(self)],
+        charge_dtype=charge_dtype)
+
```
Thenerdstation, line 550: Why did you add this back?

PR #638, opened 2020-06-01
Movedir

alewis:


Thenerdstation:
We have users who depend on the MPS classes, so we can't just remove them. Everything inside examples isn't included in the pip release.

PR #641, opened 2020-06-02
k-step arnoldi factorization

mganahl:


tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -149,3 +150,146 @@ def body_vector(i, vals):
...
+        return jax.lax.cond(val < thresh, False, lambda x: x, True, lambda x: x)
+
+      return jax.lax.cond(iteration < maxiter, [norm, threshold], check_thresh,
+                          False, lambda x: x)
+
+    norms_dtype = np.real(v0.dtype).dtype
+    kvfinal, Hfinal, _, _, norm, _, it, _ = jax.lax.while_loop(
+        cond_fun, body, [
+            krylov_vectors, H, matvec, v,
+            norms_dtype.type(100000000000.0), eps, start, num_krylov_vecs
```
Thenerdstation, line 291: any reason for the magic bajillion here?

mganahl: just any large positive number should dp

mganahl: its the initial  value of a convergence flag

mganahl: I added better formatting

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -59,9 +59,8 @@ def jax_lanczos(matvec, arguments, init, ncv, neig, landelta, reortho):
     def body_modified_gram_schmidt(i, vals):
       vector, krylov_vectors = vals
       v = krylov_vectors[i, :]
-      vector -= jax.numpy.dot(jax.numpy.conj(v),
-                              jax.numpy.ravel(vector)) * jax.numpy.reshape(
-                                  v, vector.shape)
+      vector -= jax.numpy.vdot(v, jax.numpy.ravel(vector)) * jax.numpy.reshape(
+          v, vector.shape)
```
alewis, line 63: Weren't we using jnp? `ravel` is also redundant inside `vdot`. What about

`vector -= jnp.vdot(v, vector).reshape(vector.shape)`

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -149,3 +150,151 @@ def body_vector(i, vals):
...
+    m = min(`it`,`num_krylov_vecs`). The factorization will 
+    do at most `num_krylov_vecs` steps. The returned arrays 
+    `kv` and `H` will satisfy the Arnoldi recurrence relation
+    ```
+    matrix @ Vm - Vm @ Hm - fm * em = 0
+    ```
+    with `matrix` the matrix representation of `matvec` and 
+    `Vm =  jax.numpy.transpose(kv[:it, :])`, 
+    `Hm = H[:it, :it]`, `fm = np.expand_dims(kv[it, :] * H[it, it - 1]`,1) 
+    and `em` a kartesian basis vector of shape `(1, kv.shape[1])`
```
alewis, line 239: typo

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -149,3 +150,151 @@ def body_vector(i, vals):
...
+        `krylov_vectors`: jax.array of collected krylov vectors
+        `n`: integer denoting the column-position of the overlap 
+          <`krylov_vector`|`vector`> within `H`.
+    Returns:
+      updated vals.
+             
+    """
+    vector, krylov_vectors, n, H = vals
+    v = krylov_vectors[j, :]
+    h = jax.numpy.vdot(v, jax.numpy.ravel(vector))
```
alewis, line 220: Redundant `ravel`

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -149,3 +150,151 @@ def body_vector(i, vals):
...
+    H = jax.ops.index_update(H, jax.ops.index[j, n], h)
+    vector = vector - h * jax.numpy.reshape(v, vector.shape)
+    return [vector, krylov_vectors, n, H]
+
+  @partial(jax.jit, static_argnums=(5, 6, 7))
+  def _arnoldi_fact(matvec, args, v0, krylov_vectors, H, start, num_krylov_vecs,
```
alewis, line 226: We should probably settle on what we are going to call the number of Krylov vectors in these various algorithms; in Lanczsos it looks like you call the same thing `ncv`. I personally like either that or `n_kry`; `num_krylov_vecs` feels really long to me but it's a matter of taste.

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -149,3 +150,151 @@ def body_vector(i, vals):
...
+                                            jax.ops.index[i + 1, :],
+                                            jax.numpy.ravel(Av))
+      return [krylov_vectors, H, matvec, Av, norm, threshold, i + 1, maxiter]
+
+    def cond_fun(vals):
+      _, _, _, _, norm, threshold, iteration, maxiter = vals
+
+      def check_thresh(check_vals):
+        val, thresh = check_vals
+        return jax.lax.cond(val < thresh, False, lambda x: x, True, lambda x: x)
```
alewis, line 291: These might deserve a block comment spelling out what they do? Not critical though.

tensornetwork/backends/jax/jitted_functions_test.py
```diff
@@ -0,0 +1,33 @@
...
+from tensornetwork.backends.jax.jitted_functions import _generate_arnoldi_factorization
+jax.config.update('jax_enable_x64', True)
+
+
+def test_arnoldi_factorization():
+  np.random.seed(10)
+  D = 20
+  mat = np.random.rand(D, D)
+  x = np.random.rand(D)
+  dtype = np.float64
```
alewis, line 14: Is it sufficient to just test the one dtype?

tensornetwork/backends/jax/jitted_functions.py
mganahl: jnp is only used within the JaxBackend

tensornetwork/backends/jax/jitted_functions.py
mganahl: You're right, we should be consistent here. 

I think long argument names are generally the preferable choice, in particular in backend modules that don't interface with the users.

mganahl: or let me rephrase: expressive argument names are preferable

tensornetwork/backends/jax/jitted_functions_test.py
mganahl: no, good catch!

tensornetwork/backends/jax/jitted_functions.py
mganahl: you mean kartesian?

tensornetwork/backends/jax/jitted_functions.py
alewis: Fair enough. `vdot` still flattens by definition, though.

tensornetwork/backends/jax/jitted_functions.py
alewis: That is what I was referring to, but if you were writing that on purpose for some reason go for it

tensornetwork/backends/jax/jitted_functions.py
alewis: I suppose `num_krylov_vecs` is fine, then

tensornetwork/backends/jax/jitted_functions.py
mganahl: removed `ravel`

tensornetwork/backends/jax/jitted_functions.py
mganahl: fixedx

tensornetwork/backends/jax/jitted_functions.py
mganahl: i usually use `ncv` for user-facing functions to keep it consistent with scipy. Is taht OK?

alewis:
Yeah, that makes sense to me. The difficulty is that SciPy itself is not
consistent: this parameter gets called ncv in eigh, for example, and
"restart" (which I find very confusing) in gmres, for example. When I went
through this thought process writing vumps, I eventually called it n_krylov
throughout on the grounds that consistency wasn't possible and neither of
the original names are very expressive. ncv consistently is ok too with me

Thenerdstation:
Yeah, I vote `ncv` to be consistent with scipy.

mganahl:
I left it `num_krylov_vecs` for now because I'll have to change this for all backends and in all tests. This is better to submit in a separate PR

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -115,7 +111,9 @@ def check_thresh(check_vals):
     diag_elems = jax.numpy.zeros(ncv, dtype=init.dtype)
 
     norms = jax.ops.index_update(norms, jax.ops.index[0], 1.0)
-    norms_dtype = np.real(init.dtype).dtype
+
+    norms_dtype = jax.numpy.real(jax.numpy.empty((0, 0),
+                                                 dtype=init.dtype)).dtype
```
Thenerdstation, line 116: revert this.

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -149,3 +147,163 @@ def body_vector(i, vals):
...
+  with `matrix` the matrix representation of `matvec` and 
+  `Vm =  jax.numpy.transpose(kv[:it, :])`, 
+  `Hm = H[:it, :it]`, `fm = np.expand_dims(kv[it, :] * H[it, it - 1]`,1) 
+  and `em` a kartesian basis vector of shape `(1, kv.shape[1])` 
+  with `em[0, -1] == 1` and 0 elsewhere.
+
+  Note that the caller is responsible for dtype consistency between 
+  the inputs, i.e. dtypes between all input arrays have to match.
+
+  The arguments to `_arnoldi_fact` are:
```
Thenerdstation, line 175: Remove this line.

tensornetwork/backends/jax/jitted_functions.py
mganahl: the old code broke on my system

Thenerdstation: Why? What was the error?

Doing it this way is very hacky. There must be a better way.

mganahl: it works now so I reverted it.

PR #645, opened 2020-06-03
Adding jupyter notebook with Keras layer examples

bpenchas:


Thenerdstation:
Can you move this to `tn_keras/colabs`?

PR #649, opened 2020-06-04
Modify DenseDecomp Layer to handle ndimensional input

jacksonwb:
* Modify DenseDecomp Layer to handle ndimensional input tensors
* Add tests

### DIFF
--- a/tensornetwork/tn_keras/dense.py
+++ b/tensornetwork/tn_keras/dense.py
@@ -15,6 +15,10 @@ class DenseDecomp(Layer):
   This layer is similar to performing a SVD on the weight matrix and dropping
   the lowest singular values.
 
+  This layer can take an input shape of arbitrary dimension, with the first
+  dimension expected to be a batch dimension. The weight matrix will be
+  constructed from and applied to the last input dimension.
+
   Example:
 
   ```python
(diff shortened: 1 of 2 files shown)


tensornetwork/tn_keras/dense.py
```diff
@@ -41,10 +45,10 @@ class DenseDecomp(Layer):
     bias_initializer: Initializer for the bias vector.
 
   Input shape:
-    2D tensor with shape: `(batch_size, input_shape[-1])`.
+    nD tensor with shape: `(batch_size, input.shape[1:])`.
```
bpenchas, line 48: Should be `input_shape` instead of `input.shape`, to match the usual Keras argument.
Also, tiny nit, but I think `N-D tensor` is more clear than `nD tensor`.

jacksonwb: Right. I changed it from input_shape to input.shape because in the included example, input_shape is a parameter passed to Keras specifying the size of the last dimension and thought this was confusing as an example. But I can change it back if you feel it is better.

bpenchas: `input_shape` specifies the full input shape, not just the last dimension i.e. `input_shape[-1]` is the last dimension. You can also re-write to use an ellipsis. Check out this input specification example from Keras:

 Input shape:
    N-D tensor with shape: `(batch_size, ..., input_dim)`.
    The most common situation would be
    a 2D input with shape `(batch_size, input_dim)`.

jacksonwb:
is travis broken?

Thenerdstation:
Yeah I don't know what the issue is.

Thenerdstation:
The build did in fact pass so we're good.
https://travis-ci.org/github/google/TensorNetwork/builds/695178747

PR #657, opened 2020-06-09
Experimental ncon

alewis:
Adds tensor.py and some basic methods.

- A new array class tn.Tensor intended to be the basic data structure of the library.
- tn.Tensor has, in this commit, a few basic methods (reshape, transpose, etc).
- Modify all backends so that transpose reverses axes when no perm is specified.
- Modify the TensorFlow backend so that conj does not throw an error when dtype=tf.bool.


tensornetwork/tensor.py
```diff
@@ -0,0 +1,210 @@
...
+  #    array_i = self.backend.imag(self.array)
+  #    return Tensor(array_i, backend=self.backend)
+
+  # def all(self):
+
+  # def any(self):
+
+  # def argmax():
+
+  # def argmin():
```
Thenerdstation, line 84: In general, always delete commented out code.

If you're adding these now to be made later, mark them with `TODO(alewis): Implement the following methods`.

Thenerdstation:
I'm still having a hard time seeing the value in this. In what situations does having this `Tensor` class makes things easier than just operating on say `np.array`s or `tf.Tensor`s?

alewis:
> I'm still having a hard time seeing the value in this. In what situations does having this `Tensor` class makes things easier than just operating on say `np.array`s or `tf.Tensor`s?

It shields the user from the backend, essentially. One can write code from a unified interface and switch easily between numpy, Jax, etc (and thus between easily mutated, compiled, GPU-accelerated, distributed... implementations). 

Having provided this abstraction, this interface can be tweaked in ways that make the library more convenient for traditional tensor network code without needing to supply a different function for each backend at the interface level. For example, we can provide a single SVD function that fuses and unfused indices into and back from a matricization.

Currently this feature of the library, which is important for simplifying, accelerating, and distributing traditional tensor network algorithms, is closely coupled to the Node/Edge framework for specifying a particular contraction. While this is potentially a good thing in e.g. circuit simulators where a tensor is always going to be viewed as part of a single network, this is not usually the case in more traditional applications. In those cases, where a tensor is often considered on its own rather than as part of such a contraction, this extra overhead can complicate development and maintenance. 

The idea is thus to decouple the backend-independence and quality-of-life features from the specific contraction representation.

Thenerdstation:
Right, I understand we want to decouple some non-contracting operations from the `Node` and `Edge` classes, but why specifically add a new `Tensor` class? I don't see the benefit of it over just operating on say `np.array`, `jax.DeviceArray`, `tf.Tensor`s directly. As a user, it seems even more annoying to have to deal with this extra class that seems to basically just do what my "backend Tensors" already do, but now I lose functionality of my specific backend that I wanted. (For example, I can't pass this `tn.Tensor` object into a jax.jit)

Thenerdstation:
The argument in favor of doing this is that users can switch between backends more easily. But lets really think about this, who is that useful to? The only situation I can think of is when a user wants to switch from numpy to JAX to run on GPUs, but there are so many edge cases for that transition that TN doesn't have control over (the big one being inplace updating of an array) that users will just not do it. Multiple residents had issues like that, and I expected the wider community to have the same problems.

Thenerdstation:
The main benefit I see with this is that the `tn.Tensor` object holds its own backend, but `tn.set_default_backend` and `tn.DefaultBackend(...)` largely solves that issue too.

I don't know, @mganahl, @alewis which do you think is nicer?

with `tn.Tensor`
```
tensor = tn.Tensor(jnp.ones(...))
u, s, v = tn.svd(tensor)
print(jnp.sum(jnp.diag(s.array)))
```

With `tn.DefaultBackend`

```
with tn.DefaultBackend("jax"):
  u, s, v = tn.svd(jnp.ones(...))
  print(jnp.sum(jnp.diag(s)))
```

PR #664, opened 2020-06-11
Convert input shape dimension to an integer.

kahnvex:
This fixes a case where a user is using Keras and tensorflow.compat.v1, which will send a list of Dimension objects, which do not have a __pow__ function, but do have an __int__ function.

### DIFF
--- a/tensornetwork/tn_keras/mpo.py
+++ b/tensornetwork/tn_keras/mpo.py
@@ -86,15 +86,19 @@ class DenseMPO(Layer):
     if input_shape[-1] is None:
       raise ValueError('The last dimension of the inputs to `Dense` '
                        'should be defined. Found `None`.')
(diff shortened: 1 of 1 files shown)


Thenerdstation:
Great! I'll pull this into google3 later tonight.

kahnvex:
Thanks Chase!

tensornetwork/tn_keras/mpo.py
```diff
@@ -210,3 +214,4 @@ def get_config(self) -> dict:
     # Get base config
     base_config = super(DenseMPO, self).get_config()
     return dict(list(base_config.items()) + list(config.items()))
+
```
Thenerdstation, line 217: Linter is complaining about this extra new line. I'll remove it

PR #668, opened 2020-06-12
use einsum when calculating inner products with tensordot

mganahl:


mganahl:
Hey @Thenerdstation, can you let me know what you think of this? It turns out that 
einsum is a lot faster for calculating inner product like contractions (with possible permutations)
than np.tensordot or np.dot. This is a hotfix that would improve e.g. speed of ncon with numpy backend in certain cases.

Thenerdstation:
Add a test

Thenerdstation:
That that is why you add a test. :)

mganahl:
that bit of code broke quite a few tests :)

PR #669, opened 2020-06-13
Rename base to abstract

### LINKED ISSUE #599: Rename BaseBackend and BaseNode to AbstractBackend and AbstractNode
See title

MichaelMarien:
closes #599 

* renamed BaseNode the AbstractNode
* renamed BaseBackend to AbstractBackend

Question: I did not make AbstractBackend a true abstract class, do we want that? Any methods we want to make obligatory?

Remark: when I yapf-ed my code I noticed a lot of (new) code is not yapf compliant but I did not include this not to make the PR to big. Should l?


Thenerdstation:
Thanks Michael!

Issue #670, opened 2020-06-13
Quantum circuit simulation tutorial not working


cqc-alec:
I am trying to follow the "Multiple Qubits" section of the tutorial, but cannot get the code to run. There are a few obvious errors. I inserted the missing definitions of `H` and `CNOT`, fixed the syntax error in the definition of `state_nodes` (missing closing parenthesis), and changed the definition of `qubits` (which refers to an undefined variable `nodes`) to `qubits = [node[0] for node in state_nodes]`; but then when I run the code I get the error:
```
ValueError: Cannot connect edges of unequal dimension. Dimension of edge '__unnamed_edge__': 2, Dimension of edge '__unnamed_edge__': 4.
```

cqc-alec:
I realized that I was using an incorrect definition of `CNOT`, naively thinking it should be the usual 4x4 unitary representation when in fact it should be a 2x2x2x2 array. So I did get it working in the end!

Thenerdstation:
Thanks!

Would you like to make a PR of the typos?

cqc-alec:
Yes, will do.

PR #671, opened 2020-06-13
Fix code in multi-qubit example. #670

cqc-alec:


cqc-alec:
@googlebot I signed it!

PR #681, opened 2020-06-24
adding sum and matmul to backends

mganahl:


tensornetwork/backends/pytorch/pytorch_backend.py
```diff
@@ -343,3 +343,15 @@ def broadcast_left_multiplication(self, tensor1: Tensor, tensor2: Tensor):
 
   def jit(self, fun: Callable, *args: List, **kwargs: dict) -> Callable:
     return fun
+
+  def sum(self,
+          tensor: Tensor,
+          axis: Optional[Sequence[int]] = None,
```
Thenerdstation, line 349: Should be axes

mganahl: I'm using `numpy` convention here which is `axis`

Thenerdstation: sgtm.

PR #682, opened 2020-06-24
ncon with batching support

mganahl:
adds batching support to `ncon`. Depends on #681

### DIFF
--- a/tensornetwork/backends/jax/jax_backend.py
+++ b/tensornetwork/backends/jax/jax_backend.py
@@ -410,7 +410,7 @@ class JaxBackend(abstract_backend.AbstractBackend):
         explicit orthogonalization (more costly than `reorthogonalize=False`)
     Returns:
       (eigvals, eigvecs)
-       eigvals: An jax-array containing `numeig` lowest eigenvalues
+       eigvals: A jax-array containing `numeig` lowest eigenvalues
        eigvecs: A list of `numeig` lowest eigenvectors
     """
     if args is None:
--- a/tensornetwork/backends/jax/jax_backend.py
+++ b/tensornetwork/backends/jax/jax_backend.py
@@ -29,7 +29,7 @@ _CACHED_MATVECS = {}
 class JaxBackend(abstract_backend.AbstractBackend):
   """See abstract_backend.AbstractBackend for documentation."""
 
-  def __init__(self, dtype: Optional[np.dtype] = None):
+  def __init__(self, dtype: Optional[np.dtype] = None) -> None:
     # pylint: disable=global-variable-undefined
     global libjax  # Jax module
     global jnp  # jax.numpy module
@@ -47,13 +47,14 @@ class JaxBackend(abstract_backend.AbstractBackend):
     self.name = "jax"
     self._dtype = np.dtype(dtype) if dtype is not None else None
 
-  def tensordot(self, a: Tensor, b: Tensor, axes: Sequence[Sequence[int]]):
+  def tensordot(self, a: Tensor, b: Tensor,
+                axes: Sequence[Sequence[int]]) -> Tensor:
     return jnp.tensordot(a, b, axes)
 
-  def reshape(self, tensor: Tensor, shape: Tensor):
+  def reshape(self, tensor: Tensor, shape: Tensor) -> Tensor:
     return jnp.reshape(tensor, np.asarray(shape).astype(np.int32))
 
-  def transpose(self, tensor, perm):
+  def transpose(self, tensor, perm) -> Tensor:
     return jnp.transpose(tensor, perm)
 
   def shape_concat(self, values: Tensor, axis: int) -> Tensor:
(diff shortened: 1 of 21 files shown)


mganahl:
Sorry for this rather lengthy PR. The main thing happens in ncon_interface.py, which is about 600 lines of new code. It's kind of hard to break it apart it a meaningful way

tensornetwork/ncon_interface.py
```diff
@@ -224,9 +428,50 @@ def _jittable_ncon(tensors: Sequence[Tensor], flat_labels: Tuple[int],
...
+
+  for loc in locs:
+    labels = network_structure[loc]
+    contractable_inds = np.nonzero(np.isin(labels, contractable_labels))[0]
+    network_structure[loc] = np.delete(labels, contractable_inds)
+    tensors[loc] = backend_obj.sum(tensors[loc], tuple(contractable_inds))
+
+  # update con_order after collapsing single labels
+  con_order = np.delete(con_order,
+                        np.nonzero(np.isin(con_order, contractable_labels))[0])
```
Thenerdstation, line 450: I fear this isn't going to be nearly as efficient as using batch matrix multiplication on accelerated backends like JAX, especially since we're adding axes to the output tensor that just get summed over later.

mganahl: can you tell me which part you mean here?

mganahl: The backend functions I use to do batched matrix multiplications are e.g. `jax.numpy.matmul` for JAX, which supports broadcasting over a batch index.

the code above does simply perform a regular `sum` operation over positive labels that appear only once in the `ncon` call.

Thenerdstation: Got it. this should be fine then.

tensornetwork/ncon_interface.py
```diff
@@ -23,143 +23,256 @@
...
-    for l in out_order:
-      if out_order.count(l) != 1:
-        raise ValueError(f"label {l} appears more than once in `out_order`.")
 
-  return cont_labels, out_labels
+  # check if contracted dimensions are matching
+  locations = {
+      l: np.nonzero([[l1 == l for l1 in labels] for labels in network_structure
+                    ])[0] for l in cont_labels
+  }
```
Thenerdstation, line 244: Too hard to read.

tensornetwork/tests/ncon_interface_test.py
```diff
@@ -418,3 +455,300 @@ def test_partial_traces(backend):
...
+  with pytest.raises(
+      ValueError,
+      match=r"ncon seems stuck in an infinite loop. \n"
+      r"Please check if `con_order` = \[3\] is a valid "
+      r"contraction order for \n"
+      r"`network_structure` = \[\[3, 1, 2\], \[3, 1\], \[3, 2\]\]"):
+    ncon_interface.ncon([a, b, b], [[3, 1, 2], [3, 1], [3, 2]],
+                        con_order=[3],
+                        check_network=False,
+                        backend=backend)
```
Thenerdstation, line 754: Very appreciative of the extensive testing!

Thenerdstation:
But otherwise LGTM

mganahl:
> Small nit. I would also like adam to approve this for logic.

excellent idea

Thenerdstation:
@alewis Can we get a review from you on the logic of this? LGTM from engineering design side.

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -507,3 +507,14 @@ def expm(self, matrix: Tensor):
...
+
+  def sum(self,
+          tensor: Tensor,
+          axis: Optional[Sequence[int]] = None,
+          keepdims: bool = False) -> Tensor:
+    return np.sum(tensor, axis=axis, keepdims=keepdims)
+
+  def matmul(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
+    if (tensor1.ndim <= 1) or (tensor2.ndim <= 1):
+      raise ValueError("inputs to `matmul` have to be a tensors of order > 1,")
```
alewis: Typo

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -409,7 +409,7 @@ def A(H,x):
         explicit orthogonalization (more costly than `reorthogonalize=False`)
     Returns:
       (eigvals, eigvecs)
-       eigvals: A list of `numeig` lowest eigenvalues
+       eigvals: An jax-array containing `numeig` lowest eigenvalues
```
alewis: Typo (An -> A)

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -53,8 +53,8 @@ def jax_lanczos(matvec, arguments, init, ncv, neig, landelta, reortho):
       reortho: If `True`, reorthogonalize all krylov vectors at each step. 
         This should be used if `neig>1`.
     Returns:
-      list: Eigen values
-      list: Eigen values
+      jax.numpy.ndarray: Eigenvalues
+      list: Eigenvectors
```
alewis: Should this really be a list? In NumPy it would be an array whose first axis indexes the eigenvectors. That seems more convenient to me in cases where you want to project data onto an eigenspace etc.

tensornetwork/ncon_interface.py
```diff
@@ -23,143 +23,256 @@
...
-  free labels are labels appearing exactly once.
+  Compute the contracted, free and batch labels of `network_structure`, 
+  using the following rules:
+  * Any negative number-type element and any hyphen-prepended str-type 
+    element are considered output labels.
+  * Any positive number-type element and any non-hyphen-prepended 
+    str-type element are considered contracted labels.
+  * Any negative number-type element appearing more than once, any 
+    hyphen-prepended str-type element appearing more than once, 
+    any positive element appearing exactly once and
```
alewis, line 37: My personal preference would be to throw an error for string labels containing anything but alphanumeric characters, on the grounds that an index like "JE!@//$" was probably a mistake. Even then, it's going to be very confusing trying to decide whether "Jeffrey0" or "jeff0rey" go first. But I remember some disagreement on this point.

tensornetwork/backends/jax/jitted_functions.py
mganahl: that's for consistency with block-sparse backends. There one can't easily concatentate the vectors

tensornetwork/ncon_interface.py
mganahl: The first that is easy enough to fix and seems reasonable. I think using ascii ordering is fine if we tell the users

PR #691, opened 2020-06-26
fix bug in print_msg in dmrg

mganahl:
and fix some other things once I'm at it
depends on #697


pedersor:
Apologies, as I am still trying to understand the code, but on the second sweep (moving to the right) the print still seems to miss the 2nd to last rightmost site. This seems to be due to initial_site = 1 being fixed for sweeps > 1.

mganahl:
> Apologies, as I am still trying to understand the code, but on the second sweep (moving to the right) the print still seems to miss the 2nd to last rightmost site. This seems to be due to initial_site = 1 being fixed for sweeps > 1.

Hi Ryan, yes you're right! Thanks for pointing it out again. I've fixed it

mganahl:
@Thenerdstation should be good to do

mganahl:
pls pull in #697 first!

pedersor:
Looks good. I will start looking into the two-site implementation.

Thenerdstation:
Fix the import and this can be merged.

PR #694, opened 2020-06-26
fix abstract_backend

mganahl:


Thenerdstation:
Yeah how did this not break the build?

Thenerdstation:
Aww the doc string made the function non-empty. Well, good news that it didn't really break anything before.

PR #695, opened 2020-06-26
More linalg functions

alewis:
- tensornetwork/linalg/operations.py added with some more array manipulations
- tests
- slight changes to PyTorch backend to make this work


tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,259 @@
...
+    b: Another Tensor.
+    axes: Two lists of integers. These values are the contraction
+          axes.
+  Raises:
+    ValueError, if a and b have different backends.
+  Returns:
+    The result of the tensordot, a Tensor.
+  """
+  if a.backend.name != b.backend.name:
+    errstr = "Tried to Tensordot Tensors with differing backends \n"
```
mganahl, line 55: `tensordot` instead fo Tensordot

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,259 @@
...
+    The concatenated Tensor.
+  """
+  if len(tensors) <= 1:
+    raise ValueError("Must supply at least two Tensors to concatenate.")
+  all_backends_same, errstr = _check_backends(tensors, "concatenate")
+  if not all_backends_same:
+    raise ValueError(errstr)
+  arrays = [tensor.array for tensor in tensors]
+  backend = tensors[0].backend
+  concat_array = backend.shape_concat(arrays, axis)
```
mganahl, line 118: shape_concat doesn't do what you think it does. `shape_concat` should only be used to concatenate shapes of tensors, not tensors

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,259 @@
...
+  Returns:
+    The shape of the input Tensor.
+  """
+  return tensor.shape
+
+
+def prod(values: Tensor) -> Tensor:
+  """Take the product of all of the elements in values"""
+  out_array = values.backend.shape_prod(values.array)
+  return Tensor(out_array, backend=values.backend)
```
mganahl, line 136: `shape_prod` is similar to `shape_concat` (see pytorch_backend.py)

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+def diagonal(tensor: Tensor, offset: Optional[int] = 0,
+             axis1=0, axis2=1) -> Tensor:
+  """
+  Extract diagonals from a Tensor.
+
+  Args:
+    tensor: A Tensor.
+    offset     : Offset of the diagonal from the main diagonal. If tensor is 1D,
```
mganahl, line 244: typo

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,259 @@
...
+  backend_names = [tensor.backend.name for tensor in tensors]
+  all_backends_same = True
+  for name in backend_names[1:]:
+    all_backends_same = all_backends_same and name == backend.name
+  errstr = ""
+  if not all_backends_same:
+    errstr = "All Tensors fed to " + fname + "must have the same backend."
+    errstr += "Backends were: \n"
+    errstr += str([name + "\n" for name in backend_names])
+  return (all_backends_same, errstr)
```
mganahl, line 37: you don't need to pack this explicitly into a tuple

tensornetwork/linalg/operations.py
alewis: Do you mean the spacing?

tensornetwork/linalg/operations.py
mganahl: yes

tensornetwork/linalg/tests/operations_test.py
```diff
@@ -0,0 +1,379 @@
...
+np_complex = [np.complex64, np.complex128]
+np_int = [np.int8, np.int16, np.int32, np.int64]
+np_uint = [np.uint8, np.uint16, np.uint32, np.uint64]
+np_float_dtypes = np_real + np_complex
+np_not_bool = np_float_dtypes + np_int + np_uint + [None,]
+np_all_dtypes = np_not_bool + [np.bool,]
+
+torch_supported_dtypes = np_real + np_int + [np.uint8, np.bool, None]
+
+
```
mganahl, line 40: It might be worth adding type -tests, i.e. if the array-type of the Tensor and its backend are matching. For example, concatenate seems to violate this (unless Tensor checks this or fixes it)

tensornetwork/linalg/operations.py
alewis: Oh I see, never mind.

mganahl: yes, sorry replied to the wrong comment :)

tensornetwork/linalg/tests/operations_test.py
mganahl: Just saw that `Tensor` converts to tensor, so this will be fine

tensornetwork/linalg/operations.py
alewis: Fair enough, but in this case the relevant docstrings in abstract_backend.py are perhaps misleading

mganahl: for example, jax_backend uses numpy to concatenate. If you're running on GPU and try to concatenate you'll pull data off and to the GPU unneccessarily

mganahl: yeah, you're right. This should be fixed

tensornetwork/linalg/operations.py
alewis: In this case I would propose renaming `shape_prod` to `size`

mganahl: SGTM

alewis: Actually though, the pytorch backend code I originally modified did `np.prod(np.array(values))`, which I guess is supposed to calculate the product of a size tuple that you feed it. `Tensors` in the new implementation however store a `size` attribute which renders this redundant. It would perhaps make more sense for the ABC to just do 

`def size(tensor: Tensor) -> int:
     """Returns the total number of elements in tensor"""
     return tensor.size
`

alewis: On reflection my previous comment did not make sense. Given the way the function operates it should probably still be called `shape_prod`, since it indeed takes products of shapes. Do you know when this is useful? It seems odd to me that there would be a special function to convert shapes into sizes, rather than just having a means to get sizes directly from backend arrays.

mganahl: I forgot why we have this. I think it is fixing some issues arising from different backends treat shapes of tensors. tensorflow for example used to have some terrible conventions there

PR #696, opened 2020-06-29
issues-686 repr in index.py to return more info

shreeju:
modified repr in index.py to return dense shape, charges and flow inf…ormation

### DIFF
--- a/tensornetwork/block_sparse/index.py
+++ b/tensornetwork/block_sparse/index.py
@@ -44,7 +44,10 @@ class Index:
     return self.dim
 
   def __repr__(self) -> str:
-    return str(self.dim)
+    dense_shape = str(self.dim)
+    charges = str(self._charges) 
+    flow_info = str(self.flow)
+    return f"{dense_shape} {charges} {flow_info}"
 
   @property
   def dim(self) -> int:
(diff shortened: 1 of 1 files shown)


shreeju:
@googlebot I signed it!

Thenerdstation:
First pass makes me think we'll want a slightly more formatting, but I'll leave that up to martin.

mganahl:
I agree with Chase, the formatting right now is not great. In general, please add a meaningful descriptor to each output, instead of just the value, something like 
```
dimension: 10
```
instead of  just
```
10
```

tensornetwork/block_sparse/index.py
```diff
@@ -44,7 +44,10 @@ def __len__(self) -> int:
     return self.dim
 
   def __repr__(self) -> str:
-    return str(self.dim)
+    dense_shape = str(self.dim)
```
mganahl, line 47: add a heading, something like "Index:"
and then with some inset add the actual data
als add descriptors: "dimension", "charge", "flows"

shreeju: Thank you. I have added it now.

mganahl: Thanks @shreeju! I submitted an PR to your PR with some minor modifications. Should be good once you pull it in

shreeju:
@googlebot I consent

mganahl:
@googlebot I consent

PR #704, opened 2020-07-02
Numpy gmres

alewis:
Adds GMRES to numpy_backend.py and tests of it to numpy_backend_tests.py. Unlike the earlier one, this PR concerns a properly merged checkout.


tensornetwork/backends/numpy/numpy_backend.py
```diff
@@ -283,7 +283,146 @@ def matvec(vector):
     if dtype:
       eta = eta.astype(dtype)
       U = U.astype(dtype)
-    return eta, [np.reshape(U[:, n], shape) for n in range(numeig)]
+    evs = list(eta)
+    eVs = [np.reshape(U[:, n], shape) for n in range(numeig)]
+    return evs, eVs
+
+  def gmres(self,
```
mganahl, line 290: should we also add lgmres? I found lgmres to be more stable and also faster than gmres in my applications, though it's been a few years since I have last tested this.

tensornetwork/backends/numpy/numpy_backend.py
```diff
@@ -283,7 +283,146 @@ def matvec(vector):
...
+      errstring += "match that of b, " + str(b.dtype) + "."
+      raise ValueError(errstring)
+    if num_krylov_vectors is None:
+      num_krylov_vectors = b.size
+    if num_krylov_vectors <= 0 or num_krylov_vectors > b.size:
+      errstring = "num_krylov_vectors must be in "
+      errstring += "0 < num_krylov_vectors <= b.size.\n"
+      errstring += "num_krylov_vectors = " + str(num_krylov_vectors) + ".\n"
+      errstring += "b.size = " + str(b.size) + "."
+      raise ValueError(errstring)
```
mganahl, line 398: small hint: you can break lines with strings wihthout having to do +=, and 
with f-strings (supported in python 3.4 and later I think) you can use {} to 
insert expressions into the string
```python 
errstring = f"num_krylov_vectors must be in "
            f"0 < num_krylov_vectors <= b.size.\n"
            f"num_krylov_vectors = {num_krylov_vectors}.\n"
            f"b.size = {b.size}."
```

tensornetwork/backends/numpy/numpy_backend.py
```diff
@@ -283,7 +283,146 @@ def matvec(vector):
...
+      Avec = Av.ravel()
+      return Avec
+
+    A_shape = (b.size, b.size)
+    A_op = sp.sparse.linalg.LinearOperator(matvec=matvec, shape=A_shape)
+    x, info = sp.sparse.linalg.gmres(A_op, b, x0=x0, tol=tol, atol=atol,
+                                     restart=num_krylov_vectors,
+                                     maxiter=maxiter, M=M)
+    if info < 0:
+      raise ValueError("ARPACK gmres received illegal input or broke down.")
```
mganahl, line 423: it would make sense to report `info` since its value usually conveys some information of what went wrong

mganahl: I think. If not, then its fine

alewis: info is meaningful only if greater than zero, in which case it is returned by the function.

tensornetwork/backends/numpy/numpy_backend.py
alewis: Oh, neat!

tensornetwork/backends/numpy/numpy_backend.py
alewis: We should (it is, indeed, just better) but perhaps not yet, since I haven't written it in Jax.

alewis: We also should add an iterative symmetric solver like CGS.

mganahl: let's submit two issues then

tensornetwork/backends/numpy/numpy_backend_test.py
```diff
@@ -419,6 +419,45 @@ def test_eigsh_lanczos_raises():
     backend.eigsh_lanczos(lambda x: x, initial_state=[1, 2, 3])
 
 
+def test_gmres_raises():
+  backend = numpy_backend.NumPyBackend()
+  dummy_mv = lambda x: x
+  N = 10
+  with pytest.raises(ValueError): # x0, b have different sizes
```
mganahl, line 426: add pytest.raise(ValueError, match="the error string") to all checks so we know the right error is raised

tensornetwork/backends/numpy/numpy_backend_test.py
```diff
@@ -419,6 +419,45 @@ def test_eigsh_lanczos_raises():
...
+    backend.gmres(dummy_mv, np.zeros((N,)), num_krylov_vectors=0)
+  with pytest.raises(ValueError): # num_krylov_vectors > b.size
+    backend.gmres(dummy_mv, np.zeros((N,)), num_krylov_vectors=N+1)
+  with pytest.raises(ValueError): # tol < 0
+    backend.gmres(dummy_mv, np.zeros((N,)), tol=-0.3)
+  with pytest.raises(ValueError): # atol < 0
+    backend.gmres(dummy_mv, np.zeros((N,)), atol=-0.3)
+
+
+@pytest.mark.parametrize("dtype", np_dtypes)
```
mganahl, line 447: add a test that checks the accuracy of ||Ax-b|| for the found solution x for some non-trivial matrix dimension (maybe 100 or so)

alewis:
Comments should now be fixed

PR #705, opened 2020-07-02
Abstract gmres

alewis:
Adds the GMRES interface to abstract_backend. Tests that it is not implemented there, in PyTorch, or in TensorFlow.


mganahl:
I think this should probably go before #704, since the former shouldn't build if `gmres` is not defined in the `AbstractBackend` (eihter pytype or pylint should complain).

tensornetwork/backends/pytorch/pytorch_backend_test.py
```diff
@@ -497,6 +497,12 @@ def test_eigs_not_implemented():
     backend.eigs(np.ones((2, 2)))
 
 
+def test_gmres_not_implemented():
```
mganahl, line 500: thanks for writing those tests as well!

alewis:
Normally, unless I'm really out of it, a subclass can define extra stuff besides that in the base class. So unless you've specifically enforced this somehow it should not be true. Moreover I have checked in the other branch.

mganahl:
if #704 builds then its fine

tensornetwork/backends/pytorch/pytorch_backend_test.py
alewis: No problem. I couldn't decide whether the symmetric and shell backend (what is the shell backend?) needed these too, since they do not seem to inherit from AbstractBackend and do not seem to consistently check for NotImplemented.

mganahl: I think they both inherit from `AbstractBackend`, if not then that's a bug

alewis: Ok, I'll add it then

tensornetwork/backends/abstract_backend.py
```diff
@@ -429,6 +429,104 @@ def eigsh_lanczos(self,
...
+    balancing expense and accuracy are difficult to predict in advance, so
+    applying this function requires a degree of experimentation.
+
+    In a tensor network code one is typically interested in A_mv implementing
+    some tensor contraction. This implementation thus allows `b` and `x0` to be
+    of whatever arbitrary, though identical, shape `b = A_mv(x0, ...)` expects.
+    Reshaping to and from a matrix problem is handled internally.
+
+    This function is supported only with the NumPy and Jax backends.
+    The numpy backend version is simply an interface to
```
mganahl, line 473: The last part can go, it's enough if it is described in the appropriate backend I think

alewis:
The build error is caused by my  having failed to override gmres in the numpy_backend, so I think it makes more sense to just merge this PR with the numpy one.

mganahl:
sounds good

PR #707, opened 2020-07-02
Jax gmres

alewis:
Adds a Jax implementation of GMRES.


tensornetwork/backends/jax/gmres.py
```diff
@@ -0,0 +1,225 @@
...
+  Q = Q.T.conj()
+  R = Rtilde[:-1, :]
+  g = beta*jnp.ravel(Q[:-1, 0])
+  y = jax.scipy.linalg.solve_triangular(R, g)
+  update = Vk_1[:, :-1] @ y
+  x = x0 + update
+  return x
+
+
+@partial(jax.jit, static_argnums=(2,))
```
mganahl, line 91: Is this returning an arnoldi factorization?

mganahl: If so, do you think you can use the one that is implemented in the backend?

alewis: Well, the one in the backend returns the eigendecomposition; I need the V and H from the Krylov decomposition (actually I need V and the full QR decomposition of H). If Arnoldi were refactored to produce these it could be used.

However, a more efficient implementation, which I will add later, would also update the GMRES residual by a Givens rotation once per new Krylov vector. This becomes specialized enough that I think it makes more sense to keep things together with GMRES until proved otherwise, which is reflected in the code as written.

mganahl: sorry I was not precise in my wording. I meant the arnoldi factorization in jitted_functions.py. This one priduces exactly that

mganahl: would it make sense to move gmres into the jitted_functions.py file?

alewis: My recent push should address the above

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -690,3 +690,102 @@ def test_matmul():
...
+
+  backend = jax_backend.JaxBackend()
+  A = jax.numpy.array(([[1, 1], [3, -4]]), dtype=dtype)
+  b = jax.numpy.array([3, 2], dtype=dtype)
+  x0 = jax.numpy.ones(2, dtype=dtype)
+  n_kry = 2
+
+  def A_mv(x):
+    return A @ x
+  tol = 100*jax.numpy.finfo(dtype).eps
```
Thenerdstation, line 766: Why the magic 100 here?

alewis:
The build works now

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -326,11 +326,9 @@ def A(H,x):
           type(initial_state)))
     if A not in _CACHED_MATVECS:
       _CACHED_MATVECS[A] = libjax.tree_util.Partial(libjax.jit(A))
-    if not hasattr(self, '_iram'):
-      # pylint: disable=attribute-defined-outside-init
-      self._iram = jitted_functions._implicitly_restarted_arnoldi(libjax)
-    return self._iram(_CACHED_MATVECS[A], args, initial_state, num_krylov_vecs,
-                      numeig, which, tol, maxiter)
+    imp_arnoldi = jitted_functions._implicitly_restarted_arnoldi(libjax)
```
mganahl, line 329: love that name!

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -1,56 +1,56 @@
...
-from typing import List, Any, Tuple
+from typing import List, Any, Tuple, Callable, Sequence
 import numpy as np
 Tensor = Any
 
 def _generate_jitted_eigsh_lanczos(jax):
   """
-  Helper function to generate jitted lanczos function used 
-  in JaxBackend.eigsh_lanczos. The function `jax_lanczos` 
+  Helper function to generate jitted lanczos function used
```
mganahl, line 8: Did you remove trailing white space on purpose, or is this from yapfing?

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -287,22 +288,22 @@ def body(vals):
...
+      # Continue loop while iteration < num_krylov_vecs and norm > eps
+      _, _, _, _, norm, _, iteration, _ = vals
+      counter_done = (iteration >= num_krylov_vecs)
+      norm_not_too_small = norm > eps
+      continue_iteration = jax.lax.cond(counter_done,
+                                        _, lambda x: False,
+                                        _, lambda x: norm_not_too_small)
+
+      return continue_iteration
+    initial_norm_typecaster = np.zeros((1,)) + eps + 1.
```
mganahl, line 300: you can do `initial_norm = v.real.dtype.type(1.0+eps)` instead

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -517,3 +518,105 @@ def implicitly_restarted_arnoldi_method(
...
+              num_krylov_vectors: int,
+              maxiter: int) -> Tuple[jax.ShapedArray, float, int, bool]:
+    """
+    Solve A x = b for x using the m-restarted GMRES method. This is
+    intended to be called via jax_backend.gmres.
+
+    Given a linear mapping with (n x n) matrix representation
+        A = A_mv(*A_args) gmres_m solves
+        Ax = b          (1)
+    where x and b are length-b vectors, using the method of
```
mganahl, line 545: length-n vectors?

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -517,3 +518,105 @@ def implicitly_restarted_arnoldi_method(
...
+    Generalized Minimum RESiduals with M iterations per restart (GMRES_M).
+
+    Args:
+
+    A_mv     : A function `v0 = A_mv(v, *A_args, **A_kwargs)` where `v0` and
+               `v` have the same shape.
+    b        : The `b` in `A @ x = b`.
```
mganahl, line 552: reverse order b and A_args

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -517,3 +518,105 @@ def implicitly_restarted_arnoldi_method(
...
+    """
+    Computes the residual vector r and its norm, beta, which is minimized by
+    GMRES.
+    """
+    r = b - A_mv(x, *A_args)
+    beta = jnp.linalg.norm(r)
+    return r, beta
+
+
+  #@partial(jax.jit, static_argnums=(2,))
```
mganahl, line 597: why is the jit command commented?

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -433,12 +431,160 @@ def A(H,x):
...
+  def gmres(self,
+            A_mv: Callable,
+            b: Tensor,
+            A_args: Optional[List] = None,
+            A_kwargs: Optional[dict] = None,
+            x0: Optional[Tensor] = None,
+            tol: float = 1E-05,
+            atol: Optional[float] = None,
+            num_krylov_vectors: Optional[int] = None,
+            maxiter: Optional[int] = 1,
```
mganahl, line 447: remove `Optional`

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -696,3 +690,102 @@ def test_matmul():
...
+  with pytest.raises(NotImplementedError, match=diff):
+    backend.gmres(dummy_mv, b, M=M)
+
+  A_kwargs = {"bee": "honey"}
+  diff = "A_kwargs is not supported by the Jax backend."
+  with pytest.raises(NotImplementedError, match=diff):
+    backend.gmres(dummy_mv, b, A_kwargs=A_kwargs)
+
+
+#jax_qr_dtypes = [np.float32, np.float64, np.complex64, np.complex128]
```
mganahl, line 751: remove commented code

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -696,3 +690,102 @@ def test_matmul():
...
+jax_qr_dtypes = [np.float32]
```
mganahl, line 752: Also test for float64

mganahl: and complex if possible

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -696,3 +690,102 @@ def test_matmul():
...
+  x0 = jax.numpy.ones(2, dtype=dtype)
+  n_kry = 2
+
+  def A_mv(x):
+    return A @ x
+  tol = 100*jax.numpy.finfo(dtype).eps
+  x, _ = backend.gmres(A_mv, b, x0=x0, num_krylov_vectors=n_kry, tol=tol)
+  solution = jax.numpy.array([2., 1.], dtype=dtype)
+  eps = jax.numpy.linalg.norm(jax.numpy.abs(solution) - jax.numpy.abs(x))
+  print(eps)
```
mganahl, line 770: remove print

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -696,3 +690,102 @@ def test_matmul():
...
+  assert eps < tol
+
+
+@pytest.mark.parametrize("dtype", jax_qr_dtypes)
```
mganahl, line 774: test other dtypes (float64, complex64, complex128)

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -696,3 +690,102 @@ def test_matmul():
...
+  dtype = dummy.dtype
+  backend = jax_backend.JaxBackend()
+  matshape = (100, 100)
+  vecshape = (100,)
+  A = backend.randn(matshape, dtype=dtype)
+  solution = backend.randn(vecshape, dtype=dtype)
+  def A_mv(x):
+    return A @ x
+  b = A_mv(solution)
+  tol = b.size * jax.numpy.finfo(dtype).eps
```
mganahl, line 786: good to know!

tensornetwork/backends/jax/jitted_functions.py
alewis: I removed it on purpose with a vi script

tensornetwork/backends/jax/jax_backend.py
alewis: The Optional is needed so that the default can be b.size (instead of a fixed number)

alewis:
Martin's comments should be dealt with

tensornetwork/backends/jax/jitted_functions.py
mganahl: why?

tensornetwork/backends/jax/jax_backend.py
mganahl: isn't the default 1?

tensornetwork/backends/jax/jitted_functions.py
mganahl: looks like A_args is missing from the docstring

tensornetwork/backends/jax/jitted_functions.py
alewis: Trailing whitespace violates PEP8 and is in general Bad because for example it invisible modifies the behaviour of text editor commands to jump to the end of the line. Was there some reason to keep it?

mganahl: no, just curious. If you want it removed, then it would be great if you could prepare a PR that fixes this for the whole library (just run your script over all files). This way we'll have a cleaner code history

mganahl:
Hey @alewis, did you check that your modifications to `eigsh_lanczos` and `eigs` do not trigger unneccesary tracing? I checkout out your PR and on my laptop it seems it does

alewis:
I find that tracing is triggered by the first run only, as in the screenshot. However, running the cell twice does trace twice; this also appears to have been true in the older code. 

Screen Shot 2020-07-08 at 11 31 09 AM

mganahl:
```python 

import jax
import tensornetwork as tn
import numpy as np
import time
D=200
dtype=np.float32
matrix = jax.numpy.array(np.random.rand(D,D).astype(dtype))
vector = jax.numpy.array(np.random.rand(D,).astype(dtype))
@jax.jit
def matvec_jax_matrix(vec,matrix):
    return jax.numpy.tensordot(matrix, vec,([1],[0]))
jax_backend = tn.backends.jax.jax_backend.JaxBackend()
ncv=10
t1 = time.time()
eta_j, U_j = jax_backend.eigsh_lanczos(matvec_jax_matrix,[matrix],vector,num_krylov_vecs = ncv,numeig=1, 
                                       reorthogonalize=False)
print('jax eigvals:', eta_j)
t2 = time.time()
eta_j, U_j = jax_backend.eigsh_lanczos(matvec_jax_matrix,[matrix],vector,num_krylov_vecs = ncv,numeig=1, 
                                       reorthogonalize=False)

print('jax eigvals:', eta_j)
t3 = time.time()
print('jax first:', t2 - t1)
print('jax second:', t3 - t2)
```
could you run this and send the result?

alewis:
[image: Screen Shot 2020-07-08 at 11.44.19 AM.png]

mganahl:
sorry can't see it

mganahl:
looks good, I'll pull it in as soon as building's finished

PR #713, opened 2020-07-07
Backend functions

alewis:
Adds and modifies some backend functions.

- abs and sign
- A new function `pivot' that matricizes a tensor about an axis.
- `split_axis` renamed `pivot_axis` for consistency.
- The functions called foo_decompostion are now called foo
- qr and rq accept a flag `non_negative_diagonal` that enforces the matrix R have this property.
- `diag` replaced with two functions `diagonal` and `diagflat` that respectively extract diagonal entries and create diagonal matrices.
- `trace` modified to have the same syntax as `diagonal` (as it does in numpy).
- Tests for all of these


Thenerdstation:
Break this up into smaller PRs. It's too easy for things to fall through the cracks for a 1,000+ line PR.

mganahl:
Hey @alewis what's the status here?

PR #714, opened 2020-07-08
Experimental ncon

alewis:
Adds operators and more functions to Tensor, linalg.operations, and linalg.linalg


alewis:
I'll be breaking this one up too, so don't sweat it right now - I just needed the functions together to write vuMPS.

mganahl:
no problem. Should I merge it or do you still want to change stuff?

alewis:
It would be great if you merged it (I thought it would be too big), but let me get it building

mganahl:
yes it's a lot. We'll have to break it up once we decide to merge it into master. I'll look through it as good as I can

alewis:
It's not quite as big as it looks, since most of the changes are actually those to the backends in the backend_functions commit. But yes, it's not small.

PR #715, opened 2020-07-09
bug fixes, some improvements

mganahl:


tensornetwork/block_sparse/charge.py
```diff
@@ -641,7 +641,7 @@ def identity_charge() -> np.ndarray:
     def random(cls,
                dimension: int,
                minval: int = 0,
-               maxval: int = n) -> BaseCharge:
+               maxval: int = n - 1) -> BaseCharge:
       if maxval >= n:
         raise ValueError(f"maxval must be less than n={n}, got {maxval}")
       if minval < 0:
```
Thenerdstation, line 647: Can we add a test that covers this? Looks like we missed it based on your fix.

tensornetwork/block_sparse/utils.py
```diff
@@ -532,7 +524,7 @@ def _find_transposed_diagonal_sparse_blocks(
   all_degens = np.append(orig_col_degen[col_map],
                          0)[inv_row_map[orig_row_ind.charge_labels]]
   all_cumul_degens = np.cumsum(np.insert(all_degens[:-1], 0, 0)).astype(SIZE_T)
-  dense_to_sparse = np.zeros(orig_width, dtype=SIZE_T)
+  dense_to_sparse = np.empty(orig_width, dtype=SIZE_T)
```
Thenerdstation, line 527: How certain are we that this won't break anything? I assume you're doing this as a micro optimization.

tensornetwork/block_sparse/charge.py
mganahl: NP

tensornetwork/block_sparse/utils.py
mganahl: yes, that's micro optimization. If this breaks anything then it's a bug. In fact, using empty is more save to use here because it increases that the code will crash if there was bug at an earlier point (due to empty arrays potentially taking arbitrary values).

tensornetwork/block_sparse/charge.py
mganahl: fixed

PR #721, opened 2020-07-14
Pivot

alewis:
Adds the pivot backend function


tensornetwork/backends/abstract_backend.py
```diff
@@ -734,3 +734,28 @@ def matmul(self, tensor1: Tensor, tensor2: Tensor) -> Tensor:
     """
     raise NotImplementedError(
         "Backend '{}' has not implemented `matmul`.".format(self.name))
+  
+  def pivot(self, tensor: Tensor, pivot_axis: int = 1) -> Tensor:
```
Thenerdstation, line 738: Is this a standard name / do we really need this over just `reshape`?

alewis: It's a fair question. The main use case is as part of other functions that properly act on matrices. Here we see it used in SVD, QR, and RQ (although I haven't actually refactored accordingly), and as we add more decompositions we'll need to repeat it more times. So it seemed to me a potential way to compress code.

tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -787,3 +787,15 @@ def A_mv(x):
   rtol = tol*jax.numpy.linalg.norm(b)
   atol = tol
   assert err < max(rtol, atol)
+
+
```
mganahl, line 791: set seeds

tensornetwork/backends/abstract_backend.py
Thenerdstation: sgtm.

alewis:
done

PR #727, opened 2020-07-15
Jit

alewis:
Adds, in backends/decorators.py, a tn.jit function that can be used as a function decorator.

The backend is specified either as an argument to tn.jit ("backend=..."), or by specifying the index of a "backend_arg" to tn.jit which labels the argument of the decorated function which specifies the backend. This allows jit to be used without knowledge or instantiation of the backend in advance.


alewis:
Yes, and this basically wraps that. The new functionality is as such:

```
@partial(tn.jit, backend_argnum=2)
def jit_me(A, B, backend):
    return A@B

jit_me(a, b, "jax")
```

This lets you vary the backend when jitting without needing to instantiate it until you call the jitted function.

PR #731, opened 2020-07-16
Tensor sign

alewis:
Adds sign and abs wrappers for tn.Tensor


mganahl:
@alewis fix conflicts and I'll merge.
I had a very superficial look at it, my impression is that we'll need to modify this before merging into master. I'll merge it anyhow into the experimental branch for now.

alewis:
Ok, well let me know what needs to be changed. 

Once you merge this and the decomps, I'll add tensor wrappers to diag and the decompositions as well. Then there will finally be enough in the experimental_ncon branch for me to commit the vuMPS code as an example.

mganahl:
I'll have to check. 
The title of this PR is a bit misleading. Is the VUMPS stuff there on purpose?

alewis:
Oh, no, sorry. This PR is meant to concern only the tn.Tensor sign and abs functions. I must have accidentally tracked some obsolete vuMPS code. I'll see if I can reverse that.

alewis:
You comment makes a lot more sense now

PR #735, opened 2020-07-18
Diagonal, diagflat, trace, qr, rq, svd in experimental_ncon

alewis:
- Merges master into experimental_ncon
- Adds wrappers and tests for diagonal, diagflat, trace, qr, rq and svd.


alewis:
Tried to merge into the wrong checkout by mistake

PR #738, opened 2020-07-20
Import scipy.sparse.linalg in numpy_backend.py

alewis:
I don't know how this ever worked.


Thenerdstation:
Yeah that's bad.

Can you add a test that clearly breaks without this and works with this? We apparently missed it earlier.

mganahl:
lol, I was just fixing this as well

mganahl:
there's actually a few more things that need to be fixed there, I'll submit a PR

Thenerdstation:
Yeah dang we're clearly missing a LOT of tests. Unless just doing `import sp` works?

mganahl:
It's kind of strange: test are working fine, but it breaks if you run it in a jupyter notebook

mganahl:
I guess we are importing scipy.sparse.linalg and scipy.linalg somewhere else at some point

Thenerdstation:
Do the actually cover these methods? :upside_down_face:

mganahl:
yeah they do, just checked

mganahl:
looks like jupyter's import system works a bit different than the one of the python interpreter

mganahl:
this fixes it

PR #740, opened 2020-07-20
fix bug in eigs

mganahl:


mganahl:
there was actually no bug

Thenerdstation:
My favorite kind of bug!

PR #748, opened 2020-07-22
Pivot handles pivot_axis=-1 correctly, tests improved

alewis:
- Removes error trap that excluded the pivot_axis = -1 case.
- Tests of pivot cover several choices of pivot_axis.

### DIFF
--- a/tensornetwork/backends/abstract_backend.py
+++ b/tensornetwork/backends/abstract_backend.py
@@ -812,7 +812,7 @@ class AbstractBackend:
     raise NotImplementedError(
         "Backend '{}' has not implemented `sign`.".format(self.name))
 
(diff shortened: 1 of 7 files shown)


tensornetwork/backends/abstract_backend.py
```diff
@@ -827,7 +827,7 @@ def pivot(self, tensor: Tensor, pivot_axis: int = 1) -> Tensor:
       The pivoted tensor.
     """
     ndim = len(self.shape_tuple(tensor))
-    if pivot_axis < 1 or pivot_axis > ndim:
+    if pivot_axis > ndim:
```
mganahl, line 830: what about large negative values for pivot_axis?

PR #749, opened 2020-07-22
Tensor kron

alewis:
- Adds kron and pivot wrappers for tn.Tensor.
- Changes default arguments to pivot_axis in a few places to agree with those made in master during the PR process.
- A few functions added to __init__.py.


alewis:
Also merges (which is why this is so big) and gives Tensor real and imag properties in the Jax and NumPy cases. Later we will want to make these backend functions.

alewis:
Closing until pivot_fix and non matrix gmres changes are merged into master, since this is merged with these.

PR #750, opened 2020-07-27
Improved gmres

alewis:
Replaces Jax GMRES with an improved "optimal" implementation. The new implementation:

- Finds the solution vector iteratively as the Krylov subspace is built. This saves on overhead from the QR decomposition, which is updated repeatedly with Givens rotations instead of built from whole cloth after the Arnoldi process. More importantly, it allows GMRES to terminate early if convergence is reached before the full subspace is built.

-Has all Jitted control flow expressed using jax.lax.scan and jax.lax.cond, (as opposed to jax.lax.while_loop or jax.lax.for_i) which theoretically should make the full program reverse and forward mode differentiable. This is achieved by scanning over a fixed number of iterations, but using a cond statement to make any "extra" ones perform null operations.

Unfortunately it is difficult to break this PR into smaller fragments.


alewis:
The two new files in this commit were added by mistake - the correct commit is about half this size. I'll fix it when I get to my computer

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -593,28 +601,341 @@ def gmres_residual(A_mv: Callable, A_args: Sequence, b: jax.ShapedArray,
...
+    -------
+    The same entries as in gmres_carry, updated as described above.
+    """
+    _ = gmres_xs
+    gmres_variables, gmres_constants = gmres_carry
+    err = gmres_variables[4]
+    tol = gmres_constants[0]
+    gmres_carry = jax.lax.cond(err < tol,
+                               gmres_carry, null_op,
+                               gmres_carry, gmres_work)
```
Thenerdstation, line 722: So this may not work the way you expect. In XLA (especially on TPU), it'll actually evaluate BOTH methods everytime and just return the correct one based on the conditional.

Thenerdstation: Also, the API changed for cond to make it simpler. https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.cond.html?highlight=cond#jax.lax.cond

Can you change this to 

```python
jax.lax.cond(err < tol, null_op, gmres_work, gmres_carry)
```

alewis: I see. That's a significant enough problem that I'll probably need to rewrite the whole thing, actually.

PR #755, opened 2020-07-30
Fix test_layer.py to work with new tf version

bpenchas:


Thenerdstation:
To make this pass the build can you force the `requirements_travis.txt` to have `torch==...` for whatever version last worked? We can actually fix the errors in a different PR.

tensornetwork/tn_keras/test_layer.py
```diff
@@ -397,7 +397,7 @@ def test_config(make_model):
 
   # Serialize model and use config to create new layer
   model_config = model.get_config()
-  layer_config = model_config['layers'][0]['config']
+  layer_config = model_config['layers'][1]['config']
```
Thenerdstation, line 400: These seems very sketchy. Why did this change between versions? And is it likely to happen again?

bpenchas: `model.get_config()` now returns an `InputLayer` first, while in previous versions the config did not contain an `InputLayer`. If we want to avoid only selecting a specific layer by index, we could loop over layers returned and find the one with `class_name` equal to the one we are interested in. Let me know what you think.

PR #756, opened 2020-07-31
Tensor kron

alewis:
Merges master into experimental_ncon, and adds pivot and Kron functions to the latter.


alewis:
Most of these changes are due to merging master and pivot_fix into experimental_ncon, so doing that might be in order before reviewing this.

Thenerdstation:
You're merging into master and I don't think you intended that.

PR #765, opened 2020-08-04
improved block sparse

mganahl:
Improve performance of block-sparse contractions


mganahl:
I'll break this down into smaller pieces

PR #770, opened 2020-08-06
Nuke shell backend

### LINKED ISSUE #768: Nuke Shell backend
We don't use it.

alewis:
Deletes tensornetwork/backends/shell. This seems to have been modular enough that this was all that was required - good job!

Fixes #768


Thenerdstation:
You'll need to remove it from the backend factory as well.

PR #772, opened 2020-08-06
Improved krylov

alewis:
Changes the functions in tensornetwork/linalg/krylov.py so that the matvec functions can be written directly in terms of tn.Tensor without triggering spurious recompilations in Jax. Also adds more extensive testing and error checking.


alewis:
Closing to merge with master first.

PR #775, opened 2020-08-06
Merge experimental ncon

alewis:
- Merges master into experimental_ncon.
- Adds Kron and Pivot functions operating on tn.Tensor.
- Changes default arguments for decomposition functions operating on tn.Tensor.
- Modifies the functions in linalg/krylov.py to accept matvec functions which directly involve tn.Tensor.
- Also modifies these functions to cache matvecs so as to keep Jit happy.
- Also adds more extensive error trapping to these functions.
- Adds more extensive testing of krylov.py.


mganahl:
Hey @alewis we should start thinking about what and how to merge the changes in experimental_ncon into master. If the diff becomes too large it will become really hard to do this otherwise.

alewis:
The reason this diff is so large is that it merges master into the branch, actually. But yes, I think this is the last one - my barometer was having sufficient functionality to write vuMPS

alewis:
The new stuff is, fortunately, mostly (I actually think entirely) isolated from the rest of the code: tensornetwork/tensor.py tensornetwork/tests/tensor_test.py, and the content of tensornetwork/linalg.

alewis:
As for how, I would propose we begin with tensor.py (201 lines) and, then each of the files in linalg along with the corresponding tests separately (the largest is 405 lines).

mganahl:
sounds good!

PR #769, opened 2020-08-06
Blocksparse refactor

mganahl:
A pure refactoring of functions (no new implementations)

PR #776, opened 2020-08-07
improved block-sparse

mganahl:
This PR implements a few improvements:
* A custom `unique` function which is substantially faster than `np.unique`
* A custom `intersect` function which is substantially faster than `np.intersect1d`
* Improvements in `blocksparse_utils.py` and `charge.py` which considerably speed up block-sparse tensor contractions

This PR depends on #769


Thenerdstation:
Can we break this into smaller PRs? There's been a lot of 1000+ LOC PRs lately.

mganahl:
I'll break it further up

PR #777, opened 2020-08-07
Default value for 'optimize' in abstract_backend.py

alewis:
Weird this didn't upset the linter.


alewis:
Why is the delete Shell stuff being registered as a new change here? Didn't we merge that already?

alewis:
Oh ok good that's better

alewis:
what is git

PR #778, opened 2020-08-07
Fix jax eigs bug

mganahl:
This PR fixes a bug that causes `JaxBackend.eigs` to return `NaN` values for certain operators and arguments.

### DIFF
--- a/tensornetwork/backends/jax/jax_backend.py
+++ b/tensornetwork/backends/jax/jax_backend.py
@@ -26,11 +26,11 @@ Tensor = Any
 _CACHED_MATVECS = {}
 _CACHED_FUNCTIONS = {}
 _MIN_RES_THRESHS = {
-    np.float16: 1E-3,
-    np.float32: 1E-6,
-    np.float64: 1E-12,
-    np.complex128: 1E-12,
-    np.complex64: 1E-6
+    np.dtype(np.float16): 1E-3,
+    np.dtype(np.float32): 1E-6,
+    np.dtype(np.float64): 1E-12,
+    np.dtype(np.complex128): 1E-12,
+    np.dtype(np.complex64): 1E-6
 }
 
 
@@ -304,10 +304,10 @@ class JaxBackend(abstract_backend.AbstractBackend):
         (larges magnitude).
       maxiter: Maximum number of restarts. For `maxiter=0` the routine becomes
         equivalent to a simple Arnoldi method.
-      QR_thresh: Threshold parameter. Implicitly restarted arnoldi terminates
+      res_thresh: Threshold parameter. Implicitly restarted arnoldi terminates
         if the norm of the residual `fk` of the shifted arnoldi factorization 
-        falls below `res_thresh`. If `None` a default value dependent on the 
-        dtype of the operator is chosen.
+        falls below `res_thresh`. If `None` a default value depending on the 
+        `dtype` of the operator is chosen.
     Returns:
       (eigvals, eigvecs)
        eigvals: A list of `numeig` eigenvalues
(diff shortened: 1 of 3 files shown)


tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -650,6 +650,40 @@ def test_eigs_raises():
       backend.eigs(lambda x: x, which=which)
 
 
+##################################################################
+#############  This test should just not crash    ################
+##################################################################
+@pytest.mark.parametrize("dtype", [np.float64, np.complex128])
```
alewis, line 656: Why not single precision?

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -236,7 +236,8 @@ def eigs(self,
            numeig: int = 6,
            tol: float = 1E-8,
            which: Text = 'LR',
-           maxiter: int = 20) -> Tuple[Tensor, List]:
+           maxiter: int = 20,
+           QR_thresh: float = 1E-12) -> Tuple[Tensor, List]:
```
alewis, line 240: Maybe I just can't see it in this review thingy, but is the meaning of this argument documented? Also, will its not being present in abstract_backend cause issues?

Another concern (with tol as well) is the hard coded default value. In single precision, and certainly in half precision, it's hard to imagine this threshold ever being reached.

tensornetwork/backends/jax/jax_backend_test.py
mganahl: good point, thanks!

tensornetwork/backends/jax/jax_backend.py
mganahl: you're right, I'll fix this

tensornetwork/backends/jax/jax_backend.py
```diff
@@ -25,6 +25,13 @@
 
 _CACHED_MATVECS = {}
 _CACHED_FUNCTIONS = {}
+_MIN_RES_THRESHS = {
+    np.dtype(np.float16): 1E-3,
+    np.dtype(np.float32): 1E-6,
+    np.dtype(np.float64): 1E-12,
+    np.dtype(np.complex128): 1E-12,
```
alewis, line 32: Nice way of doing this

alewis: An alternative (but this is fine) would be to make it proportional to the machine epsilon of the given data type: 1000 * np.finfo(dtype).eps or something

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -390,7 +390,12 @@ def shifted_QR(Vm, Hm, fm, evals, k, p, which):
     krylov_vectors = jax.ops.index_update(krylov_vectors, jax.ops.index[0:k, :],
                                           Vk)
     krylov_vectors = jax.ops.index_update(krylov_vectors, jax.ops.index[k:], v)
-    return krylov_vectors, H, fk
+    Z = jax.numpy.linalg.norm(fk)
+    #if fk is a zero-vector then arnoldi has exactly converged.
+    #use small threshold to check this
+    converged = jax.lax.cond(Z < res_thresh, lambda x: True, lambda x: False,
+                             None)
```
Thenerdstation, line 397: Can't you just do `converged = Z < res_thresh`?

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -492,10 +496,14 @@ def implicitly_restarted_arnoldi_method(
...
       Hm = Hm.astype(dtype)
       fm = fm.astype(dtype)
-
+    converged = False
     while (it < maxiter) and (not converged):
       evals, _ = jax.numpy.linalg.eig(Hm)
-      krylov_vectors, H, fk = shifted_QR(Vm, Hm, fm, evals, numeig, p, _which)
+      krylov_vectors, H, fk, _converged = shifted_QR(Vm, Hm, fm, evals, numeig,
+                                                     p, _which, res_thresh)
+      if _converged:
```
Thenerdstation, line 504: I think you can get rid of `_converged` in favor of just `converged`

tensornetwork/backends/jax/jitted_functions.py
mganahl: lol, you're right ...

tensornetwork/backends/jax/jax_backend.py
mganahl: that's more elegant, I agree. Can we use the same prefactor in all cases? for float16 we would have res_thres ~1

mganahl:
There's still something wrong I noticed. Need some more time with this...

Thenerdstation:
Feel free to merge it when you are ready.

PR #779, opened 2020-08-07
Tensor to master

alewis:
Adds tensornetwork/tensor.py, which contains the vaunted array class of yore, along with tests of its methods. Also modifies the backend transpose functions to make perm an optional keyword argument, with all axes reversed by default. Adds tests of this default behaviour in all backends except Symmetric, because I couldn't figure out how to do it in Symmetric.

### DIFF
--- a/tensornetwork/tensor.py
+++ b/tensornetwork/tensor.py
@@ -1,0 +1,201 @@
+# Copyright 2019 The TensorNetwork Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+import copy
+import warnings
+from typing import Any, Union, Text, Optional, List, Sequence
+import numpy as np
+from tensornetwork.backends import abstract_backend
+from tensornetwork import backends, backend_contextmanager
+
+BaseBackend = abstract_backend.AbstractBackend
+
+class Tensor():
+  def __init__(self,
+               array: Any,
+               backend: Optional[Union[Text, BaseBackend]] = None) -> None:
+    if backend is None:
+      backend = backend_contextmanager.get_default_backend()
+    backend_obj = backends.backend_factory.get_backend(backend)
+    self.backend = backend_obj
+    self.array = self.backend.convert_to_tensor(array)
+    self.shape = array.shape
+    self.size = np.prod(self.shape)
+    self.ndim = len(self.shape)
+
+  @property
+  def dtype(self) -> Any: # To maintain backend independence
+    """ Returns: The dtype of the backend array.
+    """
+    return self.array.dtype
+
+  @property
+  def T(self) -> "Tensor":
+    """ The `Tensor` with reversed axes.
+    Returns:
+      The transposed `Tensor`.
+    """
+    return self.transpose()
+
+  @property
+  def H(self) -> "Tensor":
+    """ The conjugate `Tensor` with reversed axes.
+    """
+    star = self.backend.conj(self.array)
+    array_H = self.backend.transpose(star)
+    return Tensor(array_H, backend=self.backend)
+
+  def conj(self) -> "Tensor":
+    """ Returns: The complex-conjugated `Tensor`.
+    """
+    star = self.backend.conj(self.array)
+    return Tensor(star, backend=self.backend)
+
+  def conjugate(self) -> "Tensor":
+    """ Returns: The complex-conjugated `Tensor`.
+    """
+    return self.conj()
+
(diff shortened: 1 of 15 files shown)


tensornetwork/tensor.py
```diff
@@ -0,0 +1,201 @@
...
+    array = self.backend.multiply(self.array, other)
+    return Tensor(array, backend=self.backend)
+
+  __rmul__ = __mul__
+
+  def __truediv__(self, other: Union["Tensor", float]) -> "Tensor":
+    if isinstance(other, Tensor):
+      if self.backend.name != other.backend.name:
+        errstr = (f"Backends {self.backend.name} and {other.backend.name} did"
+                  f"not agree.")
```
Thenerdstation, line 163: Change to `f"Given backends are inconsistent. Found '{self.backend.name}' and '{other.backend.name}'"`

tensornetwork/tensor.py
```diff
@@ -0,0 +1,201 @@
...
+      other = other.array
+    array = self.backend.divide(self.array, other)
+    return Tensor(array, backend=self.backend)
+
+  def __sub__(self, other: Union["Tensor", float]) -> "Tensor":
+    if isinstance(other, Tensor):
+      if self.backend.name != other.backend.name:
+        errstr = (f"Backends {self.backend.name} and {other.backend.name} did"
+                  f"not agree.")
+        raise ValueError(errstr)
```
Thenerdstation, line 174: Here too

Thenerdstation: There are other places aswell. Change it everywhere.

Thenerdstation:
You'll also need to bring your testing delta to 100%.

alewis:
Sounds good

alewis:
I think this should fix the coverage issues. There's still tests/testing_utils.py, but this is actually testing code and I think is being mishandled by the bot. Is there a way to flag it as testing code?

Thenerdstation:
So I'm reading through the coverage report and I think there are still some things you can fix. I'll comment them on the file.

tensornetwork/tests/testing_utils.py
```diff
@@ -0,0 +1,108 @@
...
+  Creates a tensor of zeros, catching errors that occur when the
+  dtype is
+  not supported by the backend. Returns both the Tensor and the backend array,
+  which are both None if the dtype and backend did not match.
+  """
+  init = np.zeros(shape, dtype=dtype)
+  if backend == "pytorch" and dtype not in torch_supported_dtypes:
+    pytest.skip("dtype unsupported by PyTorch")
+    A = None
+    init = None
```
Thenerdstation, line 60: This code is never reached after the skip.

tensornetwork/tests/testing_utils.py
```diff
@@ -0,0 +1,108 @@
...
+  init = init.astype(dtype)
+
+  if dtype in np_complex:
+    init_i = np.random.randn(*shape)
+    init = init + 1.0j * init_i.astype(dtype)
+
+  if backend == "pytorch" and dtype not in torch_supported_dtypes:
+    pytest.skip("dtype unsupported by PyTorch")
+    A = None
+    init = None
```
Thenerdstation, line 43: This code is never reached.

Thenerdstation: More specifically the `A = None` and `init = None` lines are never reached because the `pytest.skip` exits before those lines can be run.

tensornetwork/tests/testing_utils.py
```diff
@@ -0,0 +1,108 @@
...
+  return (A, init)
+
+
+def np_dtype_to_backend(backend, dtype):
+  """
+  Converts a given np dtype to the equivalent in the given backend. Skips
+  the present test if the dtype is not supported in the backend.
+  """
+  if backend is None:
+    backend = backend_contextmanager.get_default_backend()
```
Thenerdstation, line 72: This line is never reached. You can reach this by having a test use the `with tn.DefaultBackend(backend=...):` context manager instead of directly setting the backend yourself.

tensornetwork/tests/testing_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    A = jnp.array(A_np)
+  elif backend_obj.name == "tensorflow":
+    A = tf.convert_to_tensor(A_np, dtype=dtype)
+  elif backend_obj.name == "pytorch":
+    if dtype not in torch_supported_dtypes:
+      pytest.skip("dtype unsupported by PyTorch")
+
+    A = torch.tensor(A_np)
+  else:
+    raise ValueError("Invalid backend ", backend)
```
Thenerdstation, line 88: This line is never reached. Not sure how I feel about testing test code, so I think you can just delete it.

tensornetwork/tests/testing_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    if dtype in [np.uint8, tf.uint8, np.uint16, tf.uint16, np.int8, tf.int8,
+                 np.int16, tf.int16, np.uint32, tf.uint32, np.uint64,
+                 tf.uint64]:
+      skip = True
+
+  if backend == "pytorch":
+    if dtype in [np.float16, torch.float16]:
+      skip = True
+  if skip:
+    pytest.skip("backend does not support multiply-add with this dtype.")
```
Thenerdstation, line 108: This function is never used so you can delete it.

PR #780, opened 2020-08-07
add unique and intersect

mganahl:
faster unique and intersect functions

### DIFF
--- a/tensornetwork/block_sparse/utils.py
+++ b/tensornetwork/block_sparse/utils.py
@@ -109,7 +109,7 @@ def _find_best_partition(dims: Union[List[int], np.ndarray]) -> int:
   """
   if len(dims) == 1:
     raise ValueError(
-        'expecting `dims` with a length of at least 2, got `len(dims ) =1`')
+        'expecting dims with a length of at least 2, got len(dims) =1')
   diffs = [
       np.abs(np.prod(dims[0:n]) - np.prod(dims[n::]))
       for n in range(1, len(dims))
@@ -284,7 +284,7 @@ def intersect_new(A: np.ndarray,
   """
   if A.dtype != B.dtype:
     raise ValueError(f"array dtypes must macht to intersect, "
-                     f"found `A.dtype = {A.dtype}`, `B.dtype = {B.dtype}`")
+                     f"found A.dtype = {A.dtype}, B.dtype = {B.dtype}")
   if axis not in (0, 1):
     raise NotImplementedError(
         "intersection can only be performed on first or second axis")
--- a/tensornetwork/block_sparse/utils.py
+++ b/tensornetwork/block_sparse/utils.py
@@ -340,18 +340,15 @@ def intersect_new(A: np.ndarray,
           collapsed_B,
           assume_unique=assume_unique,
           return_indices=return_indices)
+      C = expand(C, A.dtype, original_width, original_ndim)
+      result = C, A_locs, B_locs      
     else:
       C = np.intersect1d(
           collapsed_A,
           collapsed_B,
           assume_unique=assume_unique,
           return_indices=return_indices)
-    C = expand(C, A.dtype, original_width, original_ndim)
-    if return_indices:
-      result = C, A_locs, B_locs
-    else:
-      result = C
-
+      result = expand(C, A.dtype, original_width, original_ndim)
   return result
 
 
--- a/tensornetwork/block_sparse/utils.py
+++ b/tensornetwork/block_sparse/utils.py
@@ -13,12 +13,11 @@
 # limitations under the License.
 
 import numpy as np
-from typing import List, Union, Any, Tuple, Optional, Sequence
-Tensor = Any
-SIZE_T = np.int64  #the size-type of index-arrays
+from tensornetwork.block_sparse.sizetypes import SIZE_T
+from tensornetwork.block_sparse.caching import get_cacher
+from typing import List, Union, Any, Type
 
-
-def get_real_dtype(dtype):
+def get_real_dtype(dtype: Type[np.number]):
   if dtype == np.complex128:
     return np.float64
   if dtype == np.complex64:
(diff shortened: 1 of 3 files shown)


mganahl:
This is the first part of the improved blocks-sparse PR. To break this apart into smaller chunks I need to jump through a few hoops and have a bunch of simultaneous branches open, so let's try pull it in quickly

alewis:
Is there a way we can rope Glen into this? I have no experience with BlockSparse or knowledge of block sparse tensors more generally and can't review the logic of these commits.

mganahl:
@gevenbly could you have a look at this and let me know what you think? thanks!

gevenbly:
@mganahl unfortunately I do not have review privileges, but I took a look at the PR. The logic of the code looks good to me. It is difficult to follow the workings of the `collapse` function. As I understand it, you collapse a width-`w` 2D array of `n`-bit ints into a 1D array of `m`-bit ints, where it is required that `m ≥ w*n`. Is this correct? 

In the case that `m > w*n` I see that you pad the initial array with zeros (e.g. in the case of 3 int16's into an int64) in order to make it fit. However, I do not see where this padding is taken account for in the `expand` function. If I were to `expand(collapse(arr))` with an array `arr` of int16's with shape (N,3) would I get back an array of the same shape?

Thenerdstation:
Glen I'll give you review permissions.

Thenerdstation:
Just sent. You should be able to review this code once you accept the invite @gevenbly

tensornetwork/block_sparse/utils.py
```diff
@@ -133,6 +136,216 @@ def get_dtype(itemsize):
...
+  Reverse operation to `collapse`. 
+  Expand a 1d numpy array `array` into a 2d array
+  of dtype `original_dtype` by view-casting.
+  """
+  if array.ndim == 1:
+    return array[:, None].view(original_dtype)
+  return array
+
+
+
```
Thenerdstation, line 183: Remove 1 newline.

tensornetwork/block_sparse/utils.py
```diff
@@ -133,6 +136,216 @@ def get_dtype(itemsize):
...
+      out = np.squeeze(out)
+
+  return out
+
+
+def intersect_new(A: np.ndarray,
+                  B: np.ndarray,
+                  axis=0,
+                  assume_unique=False,
+                  return_indices=False) -> Any:
```
Thenerdstation, line 260: Types on these.

tensornetwork/block_sparse/utils.py
```diff
@@ -133,6 +136,216 @@ def get_dtype(itemsize):
...
+  Returns:
+    ndarray: sorted 1D array of common rows/cols between the input arrays
+    ndarray: the indices of the first occurrences of the common values in A.
+      Only provided if return_indices is True.
+    ndarray: the indices of the first occurrences of the common values in B.
+      Only provided if return_indices is True.
+  """
+  if A.dtype != B.dtype:
+    raise ValueError(f"array dtypes must macht to intersect, "
+                     f"found `A.dtype = {A.dtype}`, `B.dtype = {B.dtype}`")
```
Thenerdstation, line 275: You can remove backticks in error messages since they don't get formatted.

tensornetwork/block_sparse/utils.py
```diff
@@ -133,6 +136,216 @@ def get_dtype(itemsize):
...
+    C, A_locs, B_locs = np.intersect1d(
+        collapsed_A,
+        collapsed_B,
+        assume_unique=assume_unique,
+        return_indices=True)
+    C = expand(C, A.dtype)
+    if A.ndim == 1:
+      C = np.squeeze(C)
+    if return_indices:
+      result = C, A_locs, B_locs
```
Thenerdstation, line 318: There seems to be a lot of these `if return_indicies` in this function, and it affects the number of arguments returned. This seems like a recipe for issues in the future...

mganahl:
> @mganahl unfortunately I do not have review privileges, but I took a look at the PR. The logic of the code looks good to me. It is difficult to follow the workings of the `collapse` function. As I understand it, you collapse a width-`w` 2D array of `n`-bit ints into a 1D array of `m`-bit ints, where it is required that `m ≥ w*n`. Is this correct?
> 
> In the case that `m > w*n` I see that you pad the initial array with zeros (e.g. in the case of 3 int16's into an int64) in order to make it fit. However, I do not see where this padding is taken account for in the `expand` function. If I were to `expand(collapse(arr))` with an array `arr` of int16's with shape (N,3) would I get back an array of the same shape?

Thanks Glen! The docstring of `expand` is actually not quite right, I'll fix this. Right now the un-padding is done at other parts in the code, but you are right that it would make sense to move it into `expand`. I'll do that.

tensornetwork/block_sparse/utils.py
mganahl: True! The reason is to work around a `numpy` issue with causes `np.unique` to be about 5x slower for `np.int16` and `np.int8` dtypes when `return_indices=False` as opposed to `True`. `np.intersect1d has a similar issue, see https://github.com/numpy/numpy/issues/16923. I'll see if I can make the code clearer and more robust

mganahl:
@gevenbly I changed the code of `expand` to be the inverse of `collapse`.

tensornetwork/block_sparse/utils.py
mganahl: I tried to improve the code.  The issue is with`np.intersect1d`, which is called at the base of this function, returns different number of arguments based on the values of `return_indices`. Since the `intersect_new` wrapper adds an additional operation on the first returned value of this function, I need to check if `return_indices` has or has not been set (in addition to the dtype issue mentioned above).

tensornetwork/block_sparse/utils.py
```diff
@@ -133,6 +136,214 @@ def get_dtype(itemsize):
...
+    np.ndarray: The expanded array.
+  """
+  if original_ndim == 1:
+    #nothing to expand
+    return np.squeeze(array)
+  if array.ndim == 1:
+    #the array has been collapsed
+    #now we uncollapse it
+    result = array[:, None].view(original_dtype)
+    if original_width in (3, 5, 6, 7):
```
Thenerdstation, line 195: Why the magic numbers here?

PR #781, opened 2020-08-15
Tensor initialization

alewis:
Adds code to initialize tn.Tensor objects. Also renames linalg.py to node_linalg.py, and adjusts a few files accordingly.


tensornetwork/linalg/initialization.py
```diff
@@ -0,0 +1,147 @@
...
+AbstractBackend = abstract_backend.AbstractBackend
+
+
+def initialize_tensor(fname: Text,
+                      *fargs: Any,
+                      backend: Optional[Union[Text, AbstractBackend]] = None,
+                      **fkwargs: Any) -> Tensor:
+  """Return a Tensor wrapping data obtained by an initialization function
+  implemented in a backend. The Tensor will have the same shape as the
+  underlying array that function generates, with all Edges dangling.
```
mganahl, line 34: tiny nit: I guess we should not use `Edge` here, could be confusing (also later)

PR #782, opened 2020-08-19
Tensor operations

alewis:
Adds various numpy-like functions acting on Tensor in the file linalg/operations.py


tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,405 @@
...
+    all_backends_same = all_backends_same and name == backend.name
+  errstr = ""
+  if not all_backends_same:
+    errstr = "All Tensors fed to " + fname + "must have the same backend."
+    errstr += "Backends were: \n"
+    errstr += str([name + "\n" for name in backend_names])
+  return all_backends_same, errstr
+
+
+def tensordot(a: Tensor, b: Tensor, axes: Sequence[Sequence[int]]) -> Tensor:
```
mganahl, line 42: can axes be an integer (allowed by numpy)?

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,405 @@
...
+    slice_sizes: Tuple of integers denoting size of slice along each axis.
+  Returns:
+    The slice, a Tensor.
+  """
+  sliced = tensor.backend.slice(tensor.array, start_indices, slice_sizes)
+  sliced_tensor = Tensor(sliced, backend=tensor.backend)
+  return sliced_tensor
+
+
+def shape(tensor: Tensor) -> Tuple[Optional[int], ...]:
```
mganahl, line 104: why is that Optional?

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,405 @@
...
+  """Take the square root (element wise) of a given Tensor."""
+  out_array = tensor.backend.sqrt(tensor.array)
+  return Tensor(out_array, backend=tensor.backend)
+
+
+def outer(tensor1: Tensor, tensor2: Tensor) -> Tensor:
+  """Calculate the outer product of the two given Tensors."""
+  tensors = [tensor1, tensor2]
+  all_backends_same, errstr = _check_backends(tensors, "outer")
+  if not all_backends_same:
```
mganahl, line 125: is there a reason why you are not raising the error in `_check_backends`?

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,405 @@
...
+  Returns:
+    out: The trace.
+  """
+  backend = tensor.backend
+  result = backend.trace(tensor.array, offset=offset, axis1=axis1,
+                         axis2=axis2)
+  return Tensor(result, backend=backend)
+
+
+def ncon(
```
mganahl, line 270: TBH, I'd like to veto that `ncon` only accepts `Tensor`. I use `ncon` so much for all kinds of things that it would be annoying to be forced to use it on `Tensor` objects only. I would prefer to modify `ncon` in `ncon_interface.py` instead

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,405 @@
...
+  return Tensor(result, backend=backend)
+
+
+def kron(tensorA: Tensor, tensorB: Tensor, pivot_axisA: int = -1,
+         pivot_axisB: int = -1):
+  """
+  Reshape tensorA and tensorB into matrices respectively about pivot_axisA and
+  pivot_axisB, computes the Kronecker product of those matrices, and
+  reshapes to the concatenated shape of tensorA and tensorB
+  (e.g. tensorA -> (2, 2); tensorB -> (2, 2); result -> (2, 2, 2, 2)).
```
mganahl, line 394: The docstring would be more informative if the shapes of the tensors were `(2,3)` and `(4,5)`.

tensornetwork/linalg/tests/test_operations.py
```diff
@@ -0,0 +1,326 @@
+# Copyright 2019 The TensorNetwork Authors
```
mganahl, line 1: add tests for `SymmetricBackend` where applicable

tensornetwork/linalg/tests/test_operations.py
```diff
@@ -0,0 +1,326 @@
...
+
+@pytest.mark.parametrize("dtype", testing_utils.np_not_half)
+def test_trace(backend, dtype):
+  """ Checks that Tensor.trace() works.
+  """
+  shape = (2, 3, 3)
+  A, _ = testing_utils.safe_randn(shape, backend, dtype)
+  if A is not None:
+    np.testing.assert_allclose(tensornetwork.trace(A).array,
+                               A.backend.trace(A.array))
```
mganahl, line 326: I think `kron` and `pivot' are not tested

tensornetwork/linalg/operations.py
alewis: I'd suggest that tn.ncon accept Tensor and backend.ncon accept the appropriate backend object

tensornetwork/linalg/operations.py
alewis: Well, I think this would make the backtrace a bit harder to follow

tensornetwork/linalg/tests/test_operations.py
alewis: If we want the symmetric backend to be tested by default, wouldn't it make more sense to add this to the pytest config

tensornetwork/linalg/operations.py
mganahl: SG

tensornetwork/linalg/operations.py
mganahl: backends don't have `ncon` implemented. Rather, `ncon` uses backend functions in its implementation

tensornetwork/linalg/tests/test_operations.py
mganahl: I'd love to do that. The annoying thing is that tensor initialization is so different, and we need to modify basically all existing tests

tensornetwork/linalg/operations.py
alewis: It seems a bit odd to me that this is the only function which behaves this way. Is there a principled reason for it to?

tensornetwork/linalg/tests/test_operations.py
alewis: Fair enough. There is still the basic problem that I also have no idea how to initialize symmetric arrays, what their shapes mean, etc

tensornetwork/linalg/operations.py
```diff
@@ -0,0 +1,405 @@
...
+  Args:
+    tensors: The list of tensors whose backends to check.
+    fname: The name of the calling function, which will go into the errstring.
+  Returns:
+    (flag, errstr): Whether all backends agree, and an error message if not.
+  """
+  backend = tensors[0].backend
+  backend_names = [tensor.backend.name for tensor in tensors]
+  all_backends_same = True
+  for name in backend_names[1:]:
```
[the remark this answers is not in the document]
alewis: Changed

tensornetwork/linalg/tests/test_operations.py
alewis: Tests added

tensornetwork/linalg/operations.py
alewis: Changed

tensornetwork/linalg/operations.py
alewis: I dunno what I was thinking here. I changed it.

tensornetwork/linalg/operations.py
alewis: This seems to be fine, and I've changed the type signature accordingly.

tensornetwork/linalg/operations.py
mganahl: I think I'd be OK with this, but then I'd like to get rid of one of the two `ncon` functions. Otherwise this will be pretty confusing

mganahl: Will this work with `BlockSparseTensor`s (i.e. `SymmetricBackend`)?

alewis: It basically just calls ncon on the backend object, so I think it should be fine

mganahl:
Let's try to get the coverage delta to 100% in the future.

PR #792, opened 2020-08-25
fix pylint version

mganahl:


mganahl:
set pytype and pylint versions to the latest versions for which the current master builds
(pytype==2020.6.26, pylint==2.5.3)

PR #797, opened 2020-08-27
add new initialization methods for BlockSparseTensor

mganahl:
New init methods
Code refactor
depends on #789

PR #789, opened 2020-08-24
Blocksparse new encoding

mganahl:
final part of block-sparse improvement.
Depends on #790

PR #810, opened 2020-09-02
Blocksparse sign abs

mganahl:
add `sign` and `abs` to `SymmetricBackend` and `block_sparse.linalg.py`
depends on #797 anbd #789


tensornetwork/block_sparse/initialization.py
```diff
@@ -0,0 +1,163 @@
...
+  Initialize a random symmetric tensor from random normal distribution.
+  Args:
+    indices: List of `Index` objecst, one for each leg.
+    dtype: An optional numpy dtype. The dtype of the tensor
+  Returns:
+    BlockSparseTensor
+  """
+
+  return BlockSparseTensor.randn(indices, dtype)
+
```
gevenbly, line 62: Is there any reason that all of the initializations should not be moved into the `initialization.py` file? (i.e. instead of having some of them such as `randn` still defined in `BlockSparseTensor`)

gevenbly (approved):

tensornetwork/block_sparse/initialization.py
mganahl: initialization should now actually all be in `initialization.py` (except classmethod initialization).

PR #820, opened 2020-09-08
Added from topology

### LINKED ISSUE #400: Einsum support?
Should we extend our API to support einsum equations? It could potentially make connecting nodes much less verbose. However, I question whether anyone would want to use `tn.einsum` over say `np.einsum`. Perhaps we could support only doing the left side of the equation?

Thenerdstation:
Fixes #400 

I've wanted this for a while.


mganahl (approved):
Pretty cool!

Issue #814, opened 2020-09-06
repr for Tensor


mganahl:
We need to add a useful `__repr__` to `Tensor`

gaxler:
Shape and type seems like a useful information.
Should we have a string representation of an array as part of Tensor repr?

It seems to me that Tensor class is used with linalg functions, in this case it might be annoying to call Tensor.array to examine array's content.  What do you think?

mganahl:
I think the repr of the underlying array is fine.

gaxler <notifications@github.com> schrieb am Di. 8. Sep. 2020 um 22:18:

Thenerdstation:
Agreed. I think we'd be ok to to fall back on however the underlying array does it's `__repr__` + some extra info?

gaxler:
If we gonna use underlying repr we might get a lot of duplicate info in the repr or inconsistent repr depending on the backend.

e.g. TensorFlow has shape and dtype, pytorch and numpy might have dtype
> <tf.Tensor: shape=(10, 10, 10), dtype=float32, numpy=...>
> array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1], dtype=int16)

So maybe just shape + type and leave the rest for Tensor.array?

wkostuch:
Hey, does this still need to be done?  I'm interested in working on it but this would be my first open-source contribution which is a little daunting.

mganahl:
Hi! This is addressed in PR #823.  You can add @alewis' comments to #823 if you want!

alewis:
> Hey, does this still need to be done? I'm interested in working on it but this would be my first open-source contribution which is a little daunting.

I think we're good for this one (although indeed you could make the small change suggested in #823 as a first step). However, we have many things that would be a great way to dip your toes into contributing. #840 in particular would have you moving around through most of the code, making small changes, and writing tests - having done it you should be pretty much equipped to do anything. I'd be thrilled to walk you through it if you're interested.

wkostuch:
@alewis I'll go pop over to #840, thanks for the suggestion!

PR #823, opened 2020-09-10
return array repr as Tensor repr

gaxler:
As discussed in https://github.com/google/TensorNetwork/issues/814#issue-694297329

__repr__ behavior is to fall back on the underlying array __repr__

@mganahl @Thenerdstation Could you please link the issue to the PR?


gaxler:
@googlebot I signed it!

Thenerdstation (approved):

mganahl:
Hey @gaxler, once the build is fixed we can merge this!

alewis:
Sorry to barge in at the end here, but is the __repr__ going to be actually indistinguishable from that of the underlying array? I think it should also tell you the object is a tn.Tensor.

tensornetwork/tensor.py
```diff
@@ -185,3 +185,6 @@ def __matmul__(self, other: "Tensor") -> "Tensor":
       raise ValueError(errstr)
     array = self.backend.matmul(self.array, other.array)
     return Tensor(array, backend=self.backend)
+
+  def __repr__(self):
+    return self.array.__repr__()
```
alewis, line 190: Could we do something like
`return "tn.Tensor: \" + self.array.__repr__()`?

mganahl: SGTM!

PR #827, opened 2020-09-15
* Change Edge properties to nodes, axes from node1, node2, axis1, axis2

jacksonwb:
This is just a proposal for the Edge API and doesn't pass tests or refactor code yet.

### DIFF
--- a/tensornetwork/network_components.py
+++ b/tensornetwork/network_components.py
@@ -922,39 +922,38 @@ class Edge:
   """
 
   def __init__(self,
-               node1: AbstractNode,
-               axis1: int,
+               nodes: Iterable[Union[AbstractNode, None]],
+               axes: Iterable[Union[AbstractNode, None]],
                name: Optional[Text] = None,
-               node2: Optional[AbstractNode] = None,
-               axis2: Optional[int] = None) -> None:
     """Create an Edge.
 
     Args:
+      nodes: Iterable of nodes this edge connects. 
+      axes: Iterable of ints describing the axis of this edge describes for 
+        each connecting node.
       name: Name of the edge. Used primarily for debugging.
-      node1: One of the nodes edge connects.
-      axis1: The axis of node1 that represents this edge.
-      node2: The other node that this edge connects. Can be `None` if edge is
-        dangling.
-      axis2: The axis of node2 that represents this edge. Must be `None` if
-        node2 is `None`.
 
     Raises:
-      ValueError: If node2 and axis2 are not either both `None` or both
-        not be `None`.
+      ValueError: If corresponding values for nodes and axes are not of equal
+        length and are not both `None` or neither `None`.
     """
-    if (node2 is None) != (axis2 is None):
-      raise ValueError(
-          "node2 and axis2 must either be both None or both not be None")
+    if len(nodes) != len(axes):
+      raise ValueError("nodes and axes must be of the same length")
+    for n, a in zip(nodes, axes):
+      if (n is None) != (a is None):
+        raise ValueError(
+            """corresponding node and axis must either both be None or both not 
+            be None""")
     self.is_disabled = False
     if name is None:
       name = '__unnamed_edge__'
     else:
       if not isinstance(name, str):
         raise TypeError("Edge name should be str type")
-    self._nodes = [node1, node2]
-    self._axes = [axis1, axis2]
+    self.nodes = nodes
+    self.axes = axes
     self._name = name
-    self._is_dangling = node2 is None
+    self._is_dangling = self.nodes[1] is None
 
   # contraction methods now explicitly disable Edges by setting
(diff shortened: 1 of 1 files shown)


tensornetwork/network_components.py
```diff
@@ -922,39 +922,38 @@ class Edge:
...
     else:
       if not isinstance(name, str):
         raise TypeError("Edge name should be str type")
-    self._nodes = [node1, node2]
-    self._axes = [axis1, axis2]
+    self.nodes = nodes
+    self.axes = axes
     self._name = name
-    self._is_dangling = node2 is None
+    self._is_dangling = self.nodes[1] is None
```
jacksonwb, line 956: Change initializer to use iterables

tensornetwork/network_components.py
```diff
@@ -1011,9 +1010,35 @@ def axis2(self, axis2: int) -> None:
...
+  def set_axis(self, axis_pos, axis) -> None:
+    if axis_pos < 0 or axis_pos >= len(self._axes):
+      raise ValueError(
+          f"axis_pos {axis_pos} is outside the range 0 - {len(self._axes) - 1}"
+          )
+    if not (isinstance(axis, int) or axis is None):
+      raise TypeError("axis must be of type int")
+    self._axes[axis_pos] = axis
+
+
```
jacksonwb, line 1038: Change node axes to a tuple returning property and a setter that takes an iterable and checks type

tensornetwork/network_components.py
```diff
@@ -1040,6 +1065,31 @@ def update_axis(self, old_axis: int, old_node: AbstractNode, new_axis: int,
...
+    self._nodes = list(nodes)
+
+  def set_node(self, node_pos, node) -> None:
+    if node_pos < 0 or node_pos >= len(self._nodes):
+      raise ValueError(
+          f"node_pos {node_pos} is outside the range 0 - {len(self._nodes) - 1}"
+          )
+    if not (isinstance(node, AbstractNode) or node is None):
+      raise TypeError("node must be of type AbstractNode")
+    self._nodes[node_pos] = node
```
jacksonwb, line 1091: Same for nodes

tensornetwork/network_components.py
jacksonwb: A single value update can be done with the set_axis

jacksonwb:
Could add length of nodes/axes list checking if we want to fix it to either a set length 2 or whatever length it was instantiated at. Or could leave out length checking if we want a standard Edge object to flexible accept more than 2 node/axis members

@Thenerdstation Let me know if that API seems reasonable and I'll look into refactoring changes.

tensornetwork/network_components.py
Thenerdstation, line 1091: I'm not really sure how I feel about needing to know the position of the node to change it. The position in the list is really arbitrary.

What do you think about having a `{node: [axes]}` dictionary instead? That way we won't need to worry about the positions in the list.

jacksonwb: Yeah I think that makes more sense. I definitely agree that having to know the position is too clunky.

But it does seem like there is still value in having some concept of node order. Like it seems edge.nodes[0] should always be the same. And since almost all of the edge operations are build around the concept of node0, node1 it seems like being able to to maintain the concept of the first node and the second node could be helpful.

Perhaps use a dict and just rely on the ordering of python 3.6+ dicts when returning edge.nodes and edge.axes is enough?

jacksonwb: I guess I'm wondering are there any operations that would really ever directly set a single axis or node besides the `update_axis` method? Cause if not then we don't need to worry about order at all.

tensornetwork/network_components.py
```diff
@@ -922,39 +922,42 @@ class Edge:
...
     if name is None:
       name = '__unnamed_edge__'
     else:
       if not isinstance(name, str):
         raise TypeError("Edge name should be str type")
-    self._nodes = [node1, node2]
-    self._axes = [axis1, axis2]
     self._name = name
-    self._is_dangling = node2 is None
+    self._is_dangling = len(nodes) < 2
```
jacksonwb, line 960: New edge constructor

tensornetwork/network_components.py
```diff
@@ -1028,17 +1030,27 @@ def update_axis(self, old_axis: int, old_node: AbstractNode, new_axis: int,
...
+    else:
+      axes.remove(old_axis)
+    self._nodes[n] = self._nodes.get(n, []) + [a]
+
+  @property
+  def nodes(self) -> Tuple[AbstractNode]:
+    if self.is_disabled:
+      raise ValueError(
+          'Edge has been disabled, accessing nodes is no longer possible')
+    return tuple(self._nodes.keys())
```
jacksonwb, line 1049: new update method

PR #830, opened 2020-09-15
add precision argument to jitted functions and jax backend

mganahl:
allows configuring `jax.lax.Precision` in lax-operations

### DIFF
--- a/tensornetwork/backends/jax/jax_backend.py
+++ b/tensornetwork/backends/jax/jax_backend.py
@@ -21,7 +21,6 @@ from tensornetwork.backends.jax import jitted_functions
 from functools import partial
 
 Tensor = Any
-
 # pylint: disable=abstract-method
 
 _CACHED_MATVECS = {}
@@ -38,7 +37,8 @@ _MIN_RES_THRESHS = {
 class JaxBackend(abstract_backend.AbstractBackend):
   """See abstract_backend.AbstractBackend for documentation."""
 
-  def __init__(self, dtype: Optional[np.dtype] = None) -> None:
+  def __init__(self, dtype: Optional[np.dtype] = None,
+               precision: Optional[Text] = None) -> None:
     # pylint: disable=global-variable-undefined
     global libjax  # Jax module
     global jnp  # jax.numpy module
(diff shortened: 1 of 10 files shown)


tensornetwork/backends/jax/jax_backend.py
```diff
@@ -55,10 +55,17 @@ def __init__(self, dtype: Optional[np.dtype] = None) -> None:
     jsp = libjax.scipy
     self.name = "jax"
     self._dtype = np.dtype(dtype) if dtype is not None else None
+    self.jax_precision = precision if precision is not None else libjax.lax.Precision.DEFAULT #pylint: disable=line-too-long
+
+  def configure(self, #pylint: disable=arguments-differ
+                dtype: Optional[np.dtype] = None,
+                precision: Optional[Text] = None):
+    self._dtype = np.dtype(dtype) if dtype is not None else None
+    self.jax_precision = precision if precision is not None else libjax.lax.Precision.DEFAULT #pylint: disable=line-too-long
```
Thenerdstation, line 64: I just don't want the backends to be mutable. Can we remove this method? Otherwise the rest is great.

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -733,88 +797,14 @@ def gmres_krylov(A_mv: Callable, A_args: Sequence, n_kry: int,
...
-      gmres_carry: The gmres_carry from gmres_krylov.
-    Returns:
-      (bool): Whether to continue iterating.
-    """
-    gmres_constants, gmres_variables = gmres_carry
-    tol = gmres_constants[0]
-    k = gmres_variables[0]
-    err = gmres_variables[4]
-    n_kry = gmres_constants[4]
-
```
Thenerdstation, line 757: Did you mean to delete all of this?

There's a lot of code changes for what should be a simple change.

tensornetwork/backends/jax/jitted_functions_test.py
```diff
@@ -54,7 +55,9 @@ def test_gmres_on_small_known_problem(dtype):
   def A_mv(x):
     return A @ x
   tol = A.size*jax.numpy.finfo(dtype).eps
-  x, _, _, _ = gmres.gmres_m(A_mv, [], b, x0, tol, tol, n_kry, maxiter)
+  precision = jax.lax.Precision.HIGHEST
```
alewis, line 58: If this is to be hardcoded, perhaps once at the top of the file would be best?

alewis (commented):
I didn't find anything obvious to complain about

tensornetwork/backends/jax/jax_backend.py
mganahl: sure no problem.

tensornetwork/backends/jax/jitted_functions.py
mganahl: The code has been shifted into the scope of the `precision` variable. Other than that it hasn't changed, but it looks kind of messy here ...

tensornetwork/backends/jax/jitted_functions_test.py
mganahl: good point

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -4,7 +4,7 @@
 import types
 import numpy as np
 Tensor = Any
-
+JaxPrecisionType = Any
```
jacksonwb, line 7: Would it be better to bind this to the JAX Precision Object Class? i.e. `<class 'jaxlib.xla_extension.PrecisionConfig_Precision'>`

mganahl: Good point! I'll move the binding into the various generators that generate the jitted functions. We're trying to avoid importing `jax` at global scope so that users don't need to install `jax`.

mganahl:
@alewis, @jacksonwb, @Thenerdstation are we good pulling this in?

Thenerdstation (approved):

PR #831, opened 2020-09-16
Jax eigsh cleanup

mganahl:
Clean up `jax_lanczos` in jitted_functions.py
depends in #830


tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -55,83 +61,88 @@ def jax_lanczos(matvec, arguments, init, ncv, neig, landelta, reortho):
...
-
+    shape = init.shape
+    dtype = init.dtype
+    # TODO (mganahl): replace with restarted classical gram-schmidt
     def body_modified_gram_schmidt(i, vals):
       vector, krylov_vectors = vals
       v = krylov_vectors[i, :]
-      vector -= jax.numpy.vdot(v, vector) * jax.numpy.reshape(v, vector.shape)
+      vector = vector - jax.numpy.vdot(
+          v, vector, precision=precision) * v
```
alewis, line 76: Do we need the line break?

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -55,83 +61,88 @@ def jax_lanczos(matvec, arguments, init, ncv, neig, landelta, reortho):
...
+      previous_vector, krylov_vectors = jax.lax.cond(
+          reortho,
+          lambda x: jax.lax.fori_loop(1, i, body_modified_gram_schmidt, x),
+          lambda x: x, [jax.numpy.ravel(previous_vector), krylov_vectors])
+      beta = jax.numpy.linalg.norm(previous_vector)
+      normalized_vector = previous_vector / beta
+      Av = matvec(jax.lax.reshape(normalized_vector, shape), *arguments)
+      alpha = jax.numpy.vdot(normalized_vector, Av, precision=precision)
+      alphas = alphas.at[i - 1].set(alpha)
+      betas = betas.at[i].set(
```
alewis, line 92: Another line break

tensornetwork/backends/abstract_backend.py
```diff
@@ -37,6 +37,14 @@ def tensordot(self, a: Tensor, b: Tensor,
     raise NotImplementedError(
         "Backend '{}' has not implemented tensordot.".format(self.name))
 
+  def configure(self, **kwargs) -> None:  # pylint: disable=unused-argument
+    """
+    Configure the backend.
```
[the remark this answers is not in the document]
mganahl: hmm, this should actually not be here ..

tensornetwork/backends/jax/jitted_functions.py
mganahl: yapf's suggestion, can remove it

mganahl:
@alewis ping for review, thanks!

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -58,84 +62,86 @@ def jax_lanczos(matvec: Callable, arguments: List, init: jax.ShapedArray,
...
-      return [
-          res, krylov_vectors, vector_norms, diagonal_elements, matvec, args,
-          norm, threshold, i + 1, maxiteration
-      ]
+      krylov_vectors, alphas, betas, i = vals
+      previous_vector = krylov_vectors[i, :]
+
+      previous_vector, krylov_vectors = jax.lax.cond(
+          reortho,
+          lambda x: jax.lax.fori_loop(1, i, body_modified_gram_schmidt, x),
```
alewis, line 85: cond runs both branches and returns only the result dictated by the predicate, and thus using it here gives no performance benefit. Is this intended? If you want to actually stop early, I think while_loop is the way to go.

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -58,84 +62,86 @@ def jax_lanczos(matvec: Callable, arguments: List, init: jax.ShapedArray,
...
+          lambda x: x, [jax.numpy.ravel(previous_vector), krylov_vectors])
+      beta = jax.numpy.linalg.norm(previous_vector)
+      normalized_vector = previous_vector / beta
+      Av = matvec(jax.lax.reshape(normalized_vector, shape), *arguments)
+      alpha = jax.numpy.vdot(normalized_vector, Av, precision=precision)
+      alphas = alphas.at[i - 1].set(alpha)
+      betas = betas.at[i].set(beta)
+      next_vector = jax.lax.cond(
+          reortho, lambda x: Av, lambda x: jax.numpy.reshape(
```
alewis, line 94: I think this would be easier to read if the False branch were defined separately above with a def

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -58,84 +62,86 @@ def jax_lanczos(matvec: Callable, arguments: List, init: jax.ShapedArray,
...
-                                       jax.numpy.ravel(final_state))
-
-    A_tridiag = jax.numpy.diag(diags) + jax.numpy.diag(
-        norms[1:], 1) + jax.numpy.diag(jax.numpy.conj(norms[1:]), -1)
+      betas, i = vals[-2], vals[-1]
+      norm = betas[i - 1]
+      return jax.lax.cond(i <= ncv, lambda x: x[0] > x[1], lambda x: False,
+                          [norm, landelta])
+
+    numel = np.prod(shape).astype(np.int32)
```
alewis, line 110: Is this meant to be np? It's probably harmless but just checking.

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -58,84 +62,86 @@ def jax_lanczos(matvec: Callable, arguments: List, init: jax.ShapedArray,
...
+    # set betas[0] to 1.0 for initialization of loop
+    # betas[2] is set to the norm of the initial state.
+    betas = jax.numpy.zeros(ncv + 1, dtype=dtype)
+    betas = betas.at[0].set(1.0)
+    # diagonal elements of the projected linear operator
+    alphas = jax.numpy.zeros(ncv, dtype=dtype)
+    initvals = [krylov_vecs, alphas, betas, 1]
+    krylov_vecs, alphas, betas, _ = jax.lax.while_loop(cond_fun, body_lanczos,
+                                                       initvals)
+    # FIXME (mganahl): if the while_loop stopps early at iteration i, alphas
```
alewis, line 129: Another option would be to return the final iteration number to the user, who could then slice outside of Jitted code as needed.

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -58,84 +62,86 @@ def jax_lanczos(matvec: Callable, arguments: List, init: jax.ShapedArray,
...
+    # dynamic slicing with variable slice sizes, so these beta values
+    # can't be truncated. Thus, if numeig >= i - 1, jitted_lanczos returns
+    # a set of spurious eigen vectors and eigen values.
+    # If algebraically small EVs are desired, one can initialize `alphas` with
+    # large positive values, thus pushing the spurious eigenvalues further
+    # away from the desired ones (similar for algebraically large EVs)
+
+    A_tridiag = jax.numpy.diag(alphas) + jax.numpy.diag(
+        betas[2:], 1) + jax.numpy.diag(jax.numpy.conj(betas[2:]), -1)
     eigvals, U = jax.numpy.linalg.eigh(A_tridiag)
```
alewis, line 141: Might be worth flagging this for replacement with `eigh_banded` if and when Jax supports that function.

alewis (changes requested):
Looks good, either confirm the `cond` issue is as intended or change it and we're good to go imo

tensornetwork/backends/jax/jitted_functions.py
alewis: I mean it's no big deal if it's on purpose

tensornetwork/backends/jax/jitted_functions.py
mganahl: This is `cond` statements branches between doing explicit reorthogonalization of krylov vectors or not doing it, in the latter case relying on the (ill-conditioned) assumption that constructed krylov vectors are already orthogonal.

tensornetwork/backends/jax/jitted_functions.py
mganahl: yes, this needs to be `np` to run on `TPU`

tensornetwork/backends/jax/jitted_functions.py
alewis: Right, but if your goal is to save on computation in the latter case this will not achieve that. If it's merely for correctness it's fine.

tensornetwork/backends/jax/jitted_functions.py
mganahl: I added a warning. Generally yes, but I'd like to keep it for now. I finished implementing implicitly restarted lanczos in JAX, and this will become the new default anyways

tensornetwork/backends/jax/jitted_functions.py
mganahl: done

tensornetwork/backends/jax/jitted_functions.py
mganahl: done

alewis (approved):

tensornetwork/backends/jax/jitted_functions.py
mganahl: hmm, why not?

alewis: cond always executes both branches

alewis: Is this meant as an optimization? In that case it should probably be changed.

alewis: I think what you want is to use the cond as the termination condition in a while loop

mganahl: is that also true after `jit`?

alewis: Yeah I'm pretty sure. Everything gets traced inside one of those control flow thingies anyway.

mganahl: just checked, you're right. I'll fix this then

mganahl:
Hey @alewis, can you have a look at the modified cond? thanks!

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -58,84 +62,97 @@ def jax_lanczos(matvec: Callable, arguments: List, init: jax.ShapedArray,
...
-
-      return [
-          res, krylov_vectors, vector_norms, diagonal_elements, matvec, args,
-          norm, threshold, i + 1, maxiteration
-      ]
+      krylov_vectors, alphas, betas, i = vals
+      previous_vector = krylov_vectors[i, :]
+      def body_while(vals):
+        pv, kv, _ = vals
+        pv, kv = jax.lax.fori_loop(1, i, body_modified_gram_schmidt, [pv, kv])
```
alewis, line 84: I guess this is to be replaced with the blocked classical Gram-Schmidt immediately after, so looks good

alewis (approved):

PR #834, opened 2020-09-18
jittable eigs for jax

mganahl:
A thorough cleanup of ` _arnoldi_fact` and `implicitly_restarted_arnoldi_method`:
* `_arnoldi_fact`: switch to iterative classical Gram-Schmidt instead of modifed Gram Schmidt (faster)
* `implicitly_restarted_arnoldi_method`: now fully jittable
* `"LM"` sorting has been removed because it needs further testing before

The code has been considerably cleaned up, docstrings are updated, and comments added.
depends on #830, #831, #832
fixes #813

PR #832, opened 2020-09-16
iterative classical gram-schmidt in eigsh_lanczos

mganahl:
replace the modified gram-schmidt orthogonalization with an iterated classical gram schmidt (faster)
depends on #830, #831

PR #836, opened 2020-09-21
implicitly restarted lanczos method for jax backend

mganahl:
depends on #834 and #832


tensornetwork/backends/jax/jax_backend_test.py
```diff
@@ -669,6 +675,26 @@ def mv(x, H):
...
+  np.random.seed(10)
+  init = backend.randn((D,), dtype=dtype, seed=10)
+  H = backend.randn((D, D), dtype=dtype, seed=10)
+
+  def mv(x, H):
+    return jax.numpy.dot(H, x)
+
+  eta, U = backend.eigs(
+      mv, [H], init, numeig=4, num_krylov_vecs=40, maxiter=500, which=which,
+      tol=1E-5)
```
alewis, line 692: Is it ok that the tolerance is fixed independently of the dtype?

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -1,11 +1,42 @@
...
+      krylov_vectors: Matrix of krylov vectors, each row is treated as a
+        vector.
+      iterations: Number of iterations.
+    Returns:
+      jax.ShapedArray: The orthogonalized vector.
+    """
+    vec = vector
+    overlaps = 0
+    for _ in range(iterations):
+      ov = jax.numpy.dot(
```
alewis, line 31: isn't this what vdot does?

alewis (commented):
It's quite a lot of code so I may have missed something, but it looks good to me overall.

tensornetwork/backends/jax/jax_backend_test.py
mganahl: in this case they are both double precision, so its fine. For lower precision one needs to change it, but this is currently not tested.

tensornetwork/backends/jax/jitted_functions.py
mganahl: yes, thanks for catching that!

mganahl:
> It's quite a lot of code so I may have missed something, but it looks good to me overall.

Once the other PRs are in, it'll be less

tensornetwork/backends/jax/jitted_functions.py
mganahl: actually, vdot only does scalar products

tensornetwork/backends/jax/jitted_functions.py
```diff
@@ -185,19 +201,133 @@ def body_vector(i, vals):
...
+          (i >= jax.numpy.arange(Vm.shape[0]))[:, None] * Vm, precision)
+      alphas = alphas.at[i].set(overlaps[i])
+      norm = jax.numpy.linalg.norm(Av)
+      Av = jax.numpy.reshape(Av, shape)
+      # only normalize if norm is larger than threshold,
+      # otherwise return zero vector
+      Av = jax.lax.cond(norm > tol, lambda x: Av/norm, lambda x: Av * 0.0, None)
+      Vm, betas = jax.lax.cond(
+          i < num_krylov_vecs - 1,
+          lambda x: (Vm.at[i + 1, :].set(Av.ravel()), betas.at[i].set(norm)), #pylint: disable=line-too-long
```
Thenerdstation, line 309: I think you can remove this disable.

Thenerdstation (approved):

alewis (approved):

PR #837, opened 2020-09-23
Gmres default

alewis:
Changes the default num_krylov_vectors in GMRES to 20. Also refactors most of the GMRES preprocessing into abstract_backend so that it doesn't have to be repeated.


mganahl (approved):
Looks good!

PR #838, opened 2020-09-26
fixed missing edge in create_MPS

dncolomer:
The first edge created for the case where rank = 2 was not being appended to the connected_edges array


dncolomer:
@googlebot I signed it!

Thenerdstation (approved):

PR #847, opened 2020-10-02
Addition of ones_like and zeros_like

MikeMerz:
Added functions ones_like and zeros_like with test for each function, any remarks will be appreciated.

### DIFF
--- a/tensornetwork/linalg/initialization.py
+++ b/tensornetwork/linalg/initialization.py
@@ -17,7 +17,6 @@ import warnings
 from typing import Optional, Sequence, Tuple, Any, Union, Type, Callable, List
 from typing import Text
 import numpy as np
-from tensorflow.python.framework import ops
 
 from tensornetwork.backends import abstract_backend
 from tensornetwork import backend_contextmanager
@@ -124,10 +123,10 @@ def ones_like(input: Union[Any],
     return the_tensor
   else: # incase input of type np.ndarray, convert to Tensor
     try:
-      input = ops.convert_to_tensor(input)
+      input = backend.convert_to_tensor(input)
     except TypeError as e:
       error = "Input to zeros_like has invalid type causing error massage: \n" + str(e)
-      raise TypeError(error)
+      raise TypeError(error) from e
     the_tensor = initialize_tensor("ones", input.get_shape().as_list(), backend=backend, dtype=dtype)
     return the_tensor
 
@@ -149,11 +148,11 @@ def zeros_like(input: Union[Any],
     return the_tensor
   else:  # incase input of type np.ndarray, convert to Tensor
     try:
-      input = ops.convert_to_tensor(input)
+      input = backend.convert_to_tensor(input)
     except TypeError as e:
       error = "Input to zeros_like has invalid type causing error massage: \n" + str(e)
-      raise TypeError(error)
-    the_tensor = initialize_tensor("zeros", input.get_shape().as_list(), backend=backend, dtype=dtype)
+      raise TypeError(error) from e
+    the_tensor = initialize_tensor("zeros", input.shape, backend=backend, dtype=dtype)
     return the_tensor
 
 
--- a/tensornetwork/linalg/initialization.py
+++ b/tensornetwork/linalg/initialization.py
@@ -121,7 +121,7 @@ def ones_like(input: Union[Any],
   if isinstance(input, Tensor):  # incase input of type Tensor, create Tensor normally
     the_tensor = initialize_tensor("ones", input.shape, backend=input.backend, dtype=input.dtype)
     return the_tensor
-  else: # incase input of type np.ndarray, convert to Tensor
+  else:
     try:
       input = backend.convert_to_tensor(input)
     except TypeError as e:
@@ -146,7 +146,7 @@ def zeros_like(input: Union[Any],
   if isinstance(input, Tensor):  # incase input of type Tensor, create Tensor normally
     the_tensor = initialize_tensor("zeros", input.shape, backend=input.backend, dtype=input.dtype)
     return the_tensor
-  else:  # incase input of type np.ndarray, convert to Tensor
+  else:
     try:
       input = backend.convert_to_tensor(input)
     except TypeError as e:
--- a/tensornetwork/linalg/initialization.py
+++ b/tensornetwork/linalg/initialization.py
@@ -107,7 +107,7 @@ def ones(shape: Sequence[int],
   the_tensor = initialize_tensor("ones", shape, backend=backend, dtype=dtype)
   return the_tensor
 
-def ones_like(input: Union[np.ndarray,Tensor,AbstractBackend],
+def ones_like(input: Union[Any],
          dtype: Optional[Type[Any]] = None,
          backend: Optional[Union[Text, AbstractBackend]] = None) -> Tensor:
   """Return a Tensor shape full of ones the same shape as input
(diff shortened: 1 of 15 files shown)


tensornetwork/linalg/initialization.py
```diff
@@ -17,6 +17,8 @@
 from typing import Optional, Sequence, Tuple, Any, Union, Type, Callable, List
 from typing import Text
 import numpy as np
+from tensorflow.python.ops.numpy_ops.np_arrays import convert_to_tensor
```
alewis, line 20: I know we haven't completely followed this rule, but ideally one should only import modules and not functions directly

tensornetwork/linalg/initialization.py
```diff
@@ -105,6 +107,51 @@ def ones(shape: Sequence[int],
...
+         dtype: Optional[Type[np.number]] = None,
+         backend: Optional[Union[Text, AbstractBackend]] = None) -> Tensor:
+  """Return a Tensor shape full of ones the same shape as input
+  Args:
+    input : Object to recieve shape from
+     dtype (optional) : dtype of object
+     backend(optional): The backend or its name."""
+  if backend is None:
+    backend = backend_contextmanager.get_default_backend()
+  elif isinstance(backend, str):
```
alewis, line 120: This elif is unnecessary because the factory will work fine when given a backend object

tensornetwork/linalg/initialization.py
```diff
@@ -105,6 +107,51 @@ def ones(shape: Sequence[int],
...
+    backend = backend_contextmanager.backend_factory.get_backend(backend)
+  else:
+    backend = backend
+  if isinstance(input, Tensor): # encase input of type Tensor, create Tensor normally
```
alewis, line 124: typo in comment

tensornetwork/linalg/initialization.py
```diff
@@ -105,6 +107,51 @@ def ones(shape: Sequence[int],
...
+    the_tensor = initialize_tensor("ones", input.shape, backend=input.backend, dtype=input.dtype)
+    return the_tensor
+  if isinstance(input, np.ndarray): # encase input of type np.ndarray, convert to Tensor
```
alewis, line 127: I think actually a simple "else" would be better here - presently if the input is neither a an ndarray nor a Tensor, this function will silently return None. In addition we want to catch other backend types than ndarray

tensornetwork/linalg/initialization.py
```diff
@@ -105,6 +107,51 @@ def ones(shape: Sequence[int],
...
+    input = convert_to_tensor(input,dtype=dtype)
+    the_tensor = initialize_tensor("ones", input.shape, backend=backend, dtype=dtype)
+    return the_tensor
+
+
+def zeros_like(input: Union[np.ndarray,Tensor],
+         dtype: Optional[Type[np.number]] = None,
+         backend: Optional[Union[Text, AbstractBackend]] = None) -> Tensor:
+  """Return a Tensor shape full of zeros the same shape as input
```
alewis, line 136: Same comments apply as for ones

tensornetwork/linalg/initialization.py
```diff
@@ -105,6 +107,51 @@ def ones(shape: Sequence[int],
   the_tensor = initialize_tensor("ones", shape, backend=backend, dtype=dtype)
   return the_tensor
 
+def ones_like(input: Union[np.ndarray,Tensor],
+         dtype: Optional[Type[np.number]] = None,
```
alewis, line 111: I'm not totally sure, but will this work when dtype is e.g. a PyTorch dtype?

tensornetwork/linalg/tests/initialization_test.py
```diff
@@ -100,6 +100,33 @@ def test_ones(backend):
     npI = backend_obj.ones(shape, dtype=dtype)
     np.testing.assert_allclose(tnI.array, npI)
 
+def test_ones_like(backend):
+    """Tests tensornetwork.ones_like against np.ones_like"""
+    backend_obj = backends.backend_factory.get_backend(backend)
+    shape = (2, 4, 1)
+    n = np.eye(2)
```
alewis, line 107: test with the relevant array type for each backend. also vary the dtype

alewis (changes requested):
Looks good! A few small comments

tensornetwork/linalg/initialization.py
MikeMerz: I have replaced the elif and else with just an else to create the backend

tensornetwork/linalg/tests/initialization_test.py
MikeMerz: As I saw the tests run on each backend type and at line 122 the for loops over all dtypes and each run a new objTensor is created, is that you meant?

tensornetwork/linalg/initialization.py
```diff
@@ -117,21 +117,22 @@ def ones_like(input: Union[np.ndarray,Tensor],
...
-  if isinstance(input, Tensor): # encase input of type Tensor, create Tensor normally
+    backend = backend_contextmanager.backend_factory.get_backend(backend)
+  if isinstance(input, Tensor):  # incase input of type Tensor, create Tensor normally
     the_tensor = initialize_tensor("ones", input.shape, backend=input.backend, dtype=input.dtype)
     return the_tensor
-  if isinstance(input, np.ndarray): # encase input of type np.ndarray, convert to Tensor
-    input = convert_to_tensor(input,dtype=dtype)
-    the_tensor = initialize_tensor("ones", input.shape, backend=backend, dtype=dtype)
+  else: # incase input of type np.ndarray, convert to Tensor
+    try:
```
alewis, line 126:
```suggestion
    try:
      input = backend.convert_to_tensor(input)
    except TypeError as e:
      errstr = "Input to ones_like had invalid type, causing error message: \n" +str(e)
      raise TypeError(errstr)
```

alewis: There's a very confusing situation going on right now within the codebase where, for legacy reasons, the word `Tensor' is used to refer to two very different things. At the level of tn.Tensor, it refers to the tn.Tensor object. Within the backends, it refers to the array class of the backend. Thus `convert_to_tensor` does not convert to `tn.Tensor`, but rather to the array class of the relevant backend.

alewis (changes requested):

tensornetwork/linalg/initialization.py
```diff
@@ -117,21 +117,22 @@ def ones_like(input: Union[np.ndarray,Tensor],
...
+      input = ops.convert_to_tensor(input)
+    except TypeError:
+      input = ops.convert_to_tensor(input.numpy(), dtype=dtype)
```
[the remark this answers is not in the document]
MikeMerz: It does work on TensorFlow , however pytorch type raises a TypeError as it is unable to convert but it is still possible to convert going the basic numpy array, but by the comment below an excption should be raised, so i'll change it :)

tensornetwork/linalg/tests/initialization_test.py
```diff
@@ -121,10 +121,10 @@ def test_zeros_like(backend):
     n = np.eye(2)
     for dtype in dtypes[backend]["all"]:
```
mganahl, line 122: replace `dtype` loop with a pytest parametrization.

tensornetwork/linalg/initialization.py
```diff
@@ -17,7 +17,7 @@
 from typing import Optional, Sequence, Tuple, Any, Union, Type, Callable, List
 from typing import Text
 import numpy as np
-from tensorflow.python.ops.numpy_ops.np_arrays import convert_to_tensor
```
mganahl: why is there a global `tensorflow` import? The only place where `tensorflow`, `jax` and `pytorch` should be imported is inside the respective backends.

tensornetwork/linalg/initialization.py
mganahl: also raise the last TypeError with a `from`:
```python 
raise TypeError(errstr) from e

```

mganahl (changes requested):

alewis (commented):
Looks good - change that one comment and fix the build, and I think we're good to go. Thanks!

tensornetwork/linalg/tests/initialization_test.py
```diff
@@ -100,6 +100,35 @@ def test_ones(backend):
...
+        numpyCheck = backend_obj.ones(n.shape, dtype=dtype)
+        np.testing.assert_allclose(tensor.array, tensorCheck)
+        np.testing.assert_allclose(numpyT.array, numpyCheck)
+
+def test_zeros_like(backend):
+    """Tests tensornetwork.zeros_like against np.zeros_like"""
+    backend_obj = backends.backend_factory.get_backend(backend)
+    shape = (2, 4, 1)
+    n = np.eye(2)
+    @pytest.mark.parametrize("dtype,expected",(dtypes[backend]["all"]))
```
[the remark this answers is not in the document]
MikeMerz: The types differ by the backend received and I thought that each call for a test function sends only backend so in order to get the right types for the backend received I made the inner function

PR #849, opened 2020-10-06
Add power method and test to numpy backend.

wkostuch:
I'm a little unsure about the test method for power.

Due to how np.power works, if you have a negative base and raise that to a non-integer power you receive np.nan.  However, since the tests use backend.randn, it's very likely to get a negative value as a base and then try to raise that to a non-integer power, resulting in a runtime warning.

The current "solution" I have is to take the absolute value of the base tensor, ensuring the (-x)**(not-an-integer) situation never occurs.  Is there a better way to deal with the np.power quirk or does this suffice?  

Appreciate the feedback.

### DIFF
--- a/tensornetwork/backends/numpy/numpy_backend.py
+++ b/tensornetwork/backends/numpy/numpy_backend.py
@@ -759,3 +759,24 @@ class NumPyBackend(abstract_backend.AbstractBackend):
     m.write(s.encode('latin-1'))
     m.seek(0)
     return np.load(m)
+  
+  def power(self, a: Tensor, b: Union[Tensor, float]) -> Tensor:
+    """
+    Returns the exponentiation of tensor a raised to b.  
+      If b is a tensor, then the exponentiation is element-wise 
+        between the two tensors, with a as the base and b as the power.
+        Note that a and b must be broadcastable to the same shape if 
+        b is a tensor.
+      If b is a scalar, then the exponentiation is each value in a
+        raised to the power of b.
+    
+    Args:
+      a: The tensor containing the bases.
+      b: The tensor containing the powers; or a single scalar as the power.
+
+    Returns:
+      The tensor that is each element of a raised to the 
+        power of b.  Note that the shape of the returned tensor
+        is that produced by the broadcast of a and b.
+    """
+    return np.power(a, b)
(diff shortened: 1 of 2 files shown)


alewis:
So right @mganahl @Thenerdstation  I think we need to make a ruling on what to do in this case:

- A negative number raised to a fractional power should, mathematically, return a pure imaginary output.
- However, numpy instead silently returns NaN for some reason.
- Should we:
    - Do the same thing as numpy?
    - Raise an error?
    - Or give the appropriate complex-valued return?

alewis:
I think taking the absolute value of the input is not the way to go since this is actually more surprising than NaN - at least NaN will make your program crash.

kshithijiyer (approved):

mganahl:
i think we should stick to whatever the backends are doing. 
Why only add power to numpy and not the other backends as well?

alewis:
We're gonna add it to all of them, just not in the same PR

alewis:
If we stick to what the backends are doing, the behaviour of our programs
will change as the backend does, which I thought we were trying to avoid.
Is that ok do you think?

wkostuch:
In case this is helpful to have on hand: 

TensorFlow's power method throws an error when a negative float is raised to a non-integer power:
`tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute Pow as input #1(zero-based) was expected to be a int32 tensor but is a float tensor [Op:Pow]`

PyTorch's power method does the same as Numpy's and silently returns nan values:
`tensor([0.3481, 4.2450,    nan,    nan])`

mganahl:
> If we stick to what the backends are doing, the behaviour of our programs will change as the backend does, which I thought we were trying to avoid. Is that ok do you think?

Numpy returns nan because raising a negative float to a fractional power changes the dtype to complex. np.power(complex(-1), 0.1) works fine.

As of @wkostuch comment, it seems that all backends are behaving similarly in that respect, so I'd just leave it that way. Otherwise we will have to violate type preservation (which in numpy will lead to propagating dtypes), which can cause a whole bunch of issues itself.

alewis:
Ok, SGTM

alewis (approved):

mganahl:
Hey @wkostuch and @alewis, how are the other backends coming?

wkostuch:
@mganahl, I was hoping to get around to them both this weekend.  I've been swamped with school and exams the last little while so I haven't had time to do this.  Sorry 'bout that!

PR #855, opened 2020-10-14
block method added

Sanchayan-Bhowal:
I have added a function called block which converts size 1 Tensor to Python Scalar.
Please check it out and suggest changes.


tensornetwork/tensor.py
```diff
@@ -111,7 +111,11 @@ def reshape(self, shape: Sequence[int]) -> "Tensor":
     """
     reshaped = self.backend.reshape(self.array, shape)
     return Tensor(reshaped, backend=self.backend)
-
+  def block(self):
```
mganahl, line 114: Is that standard terminology?

Sanchayan-Bhowal: Sorry but I couldn't understand your question can u please explain?

mganahl: The function name `block` is not very descriptive of what the function is actually doing. `item` would be more appropriate.

Sanchayan-Bhowal: Okay I would change it but how should I change it. I am really sorry that I am disturbing you by asking silly questions. Once I get to know the procedure I would change it. @mganahl Thank you for your feedback

tensornetwork/tensor.py
```diff
@@ -111,7 +111,11 @@ def reshape(self, shape: Sequence[int]) -> "Tensor":
     """
     reshaped = self.backend.reshape(self.array, shape)
     return Tensor(reshaped, backend=self.backend)
```
alewis, line 113: Put one space between methods

tensornetwork/tensor.py
```diff
@@ -111,7 +111,11 @@ def reshape(self, shape: Sequence[int]) -> "Tensor":
...
-
+  def item(self):
+    """Return a Python scalar formed after converting a size-1 Tensor
```
alewis, line 115: "Copies the single element of this `Tensor` and returns it. If this `Tensor` has more than one element,
raises `ValueError`."

tensornetwork/tensor.py
```diff
@@ -111,7 +111,11 @@ def reshape(self, shape: Sequence[int]) -> "Tensor":
...
+        if size of Tensor is not 1
+        then raises ValueError"""
+    return self.array.item()
```
alewis, line 118: I don't think all the backend arrays have this method, so we will have to add an item method to the backends as well. See e.g. `reshape` for how that looks within `Tensor`.

alewis (commented):
A few comments. We also need a test.

alewis:
Fix the build and I'll take a look here

Sanchayan-Bhowal:
> Fix the build and I'll take a look here

@alewis Sorry, I couldn't understand. Which build are you talking about? I fixed the issues that you described. Can you please elaborate.

alewis:
Hi Sanchayan-Bhowal,

Sorry for the long delay! We use a continuous integration tool called Travis that makes sure all code passes certain prerequisites before pushing. This PR is currently failing as indicated by the red x. You can see the problem by scrolling down to where it says "some checks were not successful" and clicking 'Details'.

PR #856, opened 2020-10-16
Support of .imag and .real for backends

MikeMerz:


alewis:
Fix the build and I'll look this over. Thanks!

PR #865, opened 2020-10-30
add item() to all backends

mganahl:


tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -383,3 +385,15 @@ def sign(self, tensor: Tensor) -> Tensor:
       tensor: The input tensor.
     """
     return tf.math.sign(tensor)
+
+  def item(self, tensor):
+    numel = 0
+    if len(tensor.shape) > 0:
+      numel = reduce(mul, tensor.shape)
```
Thenerdstation, line 392: We generally prefer using `np.prod`.

tensornetwork/block_sparse/blocksparsetensor.py
```diff
@@ -430,10 +431,15 @@ def __repr__(self):
     return output
 
   def item(self):
-    if self.ndim == 0:
-      if len(self.data) == 1:
-        return self.data[0]
-    raise ValueError("can only convert an array of size 1 to a Python scalar")
+    if len(self.shape) > 0:
+      if reduce(mul, self.shape) != 1:
```
Thenerdstation, line 435: `np.prod` here too

Thenerdstation (approved):

tensornetwork/backends/tensorflow/tensorflow_backend.py
mganahl: np.prod has a significantly larger overhead for small sequences, this is why I'm using reduce here

mganahl: it's not important though

Thenerdstation: If it's actually faster than we can keep it.

Thenerdstation:
Feel free to merge this martin

tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -17,6 +17,8 @@
 from tensornetwork.backends import abstract_backend
 from tensornetwork.backends.tensorflow import decompositions
 from tensornetwork.backends.tensorflow import tensordot2
+from functools import reduce
```
alewis, line 20: Only import modules

alewis (changes requested):

alewis:
As you're probably getting by now I think we should not import classes or functions directly but only modules (which you can give a short name if you want). Otherwise looks good.

mganahl:
taking executive action

PR #867, opened 2020-10-30
DMRG optimization different backends (including symmetric) in examples

mganahl:
depends on #866, #865


examples/dmrg/symmetric_dmrg.py
```diff
@@ -0,0 +1,179 @@
...
+Example of backend independent DMRG calculation for the
+spin 1/2 Heisenberg Model.
+"""
+
+from typing import Type, Text
+import tensornetwork as tn
+import numpy as np
+from tensornetwork import (FiniteMPO, FiniteMPS, U1Charge, Index,
+                           BlockSparseTensor, FiniteXXZ, FiniteDMRG)
+import jax
```
alewis, line 11: Presumably we don't want to import this unless the Jax backend is desired? Otherwise this code will fail if Jax isn't installed.

examples/dmrg/symmetric_dmrg.py
```diff
@@ -0,0 +1,179 @@
+"""
+Example of backend independent DMRG calculation for the
+spin 1/2 Heisenberg Model.
+"""
+
+from typing import Type, Text
+import tensornetwork as tn
+import numpy as np
+from tensornetwork import (FiniteMPO, FiniteMPS, U1Charge, Index,
```
alewis, line 9: I think we should only import modules except from typing

examples/dmrg/symmetric_dmrg.py
```diff
@@ -0,0 +1,179 @@
...
+  Args:
+    N: Number of spins.
+    D: The bond dimension.
+    dtype: The data type of the MPS.
+
+  Returns:
+    `FiniteMPS`: A spin 1/2 mps for the corresponding backend.
+  """
+  if backend == 'symmetric':
+    return blocksparse_halffilled_spin_mps(N=N, D=D, B=5, dtype=dtype)
```
alewis, line 89: Is B=5 fixed by the problem?

tensornetwork/backends/abstract_backend.py
```diff
@@ -1010,4 +1010,15 @@ def power(self, a: Tensor, b: Union[Tensor, float]) -> Tensor:
     """
     raise NotImplementedError(
         f"Backend {self.name} has not implemented power.")
-        
\ No newline at end of file
+
+  def item(self, tensor) -> Union[float, int, complex]:
+    """
+    Return the item of a 1-element tensor.
```
alewis, line 1016: This docstring is a bit opaque. What's the "item" of a tensor?

tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -17,6 +17,8 @@
 from tensornetwork.backends import abstract_backend
 from tensornetwork.backends.tensorflow import decompositions
 from tensornetwork.backends.tensorflow import tensordot2
+from functools import reduce
```
alewis, line 20: Only import modules

alewis (changes requested):

examples/dmrg/symmetric_dmrg.py
mganahl: I think that's fine in the example folder. That's not really part of the package, and we're already importing this for example in `simple_mera.py`

examples/dmrg/symmetric_dmrg.py
mganahl: no, it's a hyper parameter, but not important for the actual simulation

alewis (approved):

mganahl:
Hmm, let me fix that

Adam Lewis <notifications@github.com> schrieb am Mi. 2. Dez. 2020 um 16:33:

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,36 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
```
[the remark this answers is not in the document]
mganahl: that shouldn't be here. I guess it was added by accident.

Issue #863, opened 2020-10-29
Giving ncon a pythonic makeover


repst:
The function `ncon` is powerful and beloved by veteran tensor networkers. Yet despite its prowess, it hasn't shed its matlabian exterior which prevents it from feeling like a natural python function. First, edge indices that start at 1 don't mesh well with python nor with TN in general where tensor axes that start at index 0. Second, python has a nice alternative to passing in two lists of associated things.

Here are a few ideas on how to adapt `ncon` to python.

Given tensors `left`, `mid` and `right`, an example of the current syntax is:
`ncon([left, mid, right],  [[1, -1], [1, -2, 2], [2, -3]])`

If instead of negative numbers, complex numbers are used to represent dangling edges, 0-based indexing is possible. In addition, replacing the two related lists with a dictionary brings edge indices next to their associated tensors, removes a nesting of lists and reduces the number of function arguments by one:
`ncon({left:[0, 0j], mid:[0, 1j, 1], right: [1, 2j]})`

Alternatively, one could use strings instead of lists with `d` representing dangling edges:
`ncon({left:'0, 0d', mid:'0, 1d, 1', right:'1, 2d'})`

I'd be curious to hear thoughts from the various stakeholders, especially the black belt `ncon` users.

mganahl:
Hey @repst thanks for the proposal. Sounds very interesting to me! Let me think a bit about this. Also roping in @gevenbly on this one!

Thenerdstation:
This seams very much in the spirit of the "functional node" design we discussed months prior.

The idea was very similar to what you are describing, only instead of wrapping everything in a single ncon function, you would build the contraction like

```
left = tn.Tensor(...)
mid = tn.Tensor(...)
right = tn.Tensor(...)
resultbuilder = left['0,0d'] @ mid['0, 1d, 1'] @ right['1, 2d']
result = resultbuilder['0d,1d,2d'].to_tensor() # tn.Tensor
```

alewis:
oh no

alewis:
wait didn't Martin and Glen implement the string thing as "fancy ncon"? what happened to that?

anyway so basically the concern is how to handle contractions involving, say, 10 or more tensors with 15 legs each or something. in addition one needs a means to specify the contraction order, ideally without having to rewrite the whole thing

[...]

PR #869, opened 2020-10-30
Added NconBuilder

Thenerdstation:
Allows you to build ncon functions not in a single function call. 

Example:

```python
a = tn.Tensor(...)
b = tn.Tensor(...)
builder = a(-1, 1) @ b(1, -2)
# tn.finalize converts a tn.NconBuilder to a tn.Tensor
print(tn.finalize(builder)) # tn.Tensor
```
Discussion in #863


Thenerdstation:
The cool part about building it this way is that any upgrade to ncon upgrades this API automatically.

Thenerdstation:
I also wanted it to be `builder.finalize()`, but I was hitting cyclic import errors with `ncon`. I still think this isn't terrible.

alewis:
I like it! I also prefer the int arguments to strings, so that's good. 

`builder.finalize()` would indeed be better, but yeah this is ok. We can always add that later if we figure it out, and just leave tn.finalize in for legacy code.

tensornetwork/ncon_interface.py
```diff
@@ -661,3 +661,9 @@ def ncon(
   if all(are_tensors):
     return tn_tensor.Tensor(res_tensor, backend=backend_obj)
   return res_tensor
+
+def finalize(ncon_builder: tn_tensor.NconBuilder) -> tn_tensor.Tensor:
+  return ncon(
+      ncon_builder.tensors, 
+      ncon_builder.axes,
+      backend=ncon_builder.tensors[0].backend)
```
alewis, line 669: github wants a newline at the end of the file here

tensornetwork/tensor.py
```diff
@@ -185,3 +185,18 @@ def __matmul__(self, other: "Tensor") -> "Tensor":
...
     array = self.backend.matmul(self.array, other.array)
     return Tensor(array, backend=self.backend)
+
+  def __call__(self, *args):
+    return NconBuilder([self], [list(args)]) 
+
+
+class NconBuilder():
+  def __init__(self, tensors, axes):
+    self.tensors = tensors[:] # Forces a copy.
```
alewis, line 195: wait, why do we want a copy?

alewis (approved):
Nice, I like this a lot. I'm also amazed it's so short. Two small comments.

alewis:
I suppose later we could have finalize take a dict argument specifying a permutation of the natural contraction order, and/or specifications of an optimization algorithm

alewis:
Do you think `ContractionBuilder` or even just `Contraction` might be a more evocative name to people who don't already know what `ncon` is?

alewis:
If so (and maybe regardless) I would recommend also changing `finalize` to `contract`

alewis:
Ooh, we could also later modify this so that the contraction pattern could be specified or retrieved independently of the concrete tensors:

```
a = tn.Tensor(...)
b = tn.Tensor(...)
builder = a(-1, 1) @ b(1, -2)
# tn.finalize converts a tn.NconBuilder to a tn.Tensor
print(tn.finalize(builder)) # tn.Tensor
contraction = builder.contraction() #tn.Contraction
print(contraction.pattern()) # (-1, 1); (1, -2)

c = a + 2
d = tn.abs(b)
print(tn.finalize(contraction(c, d))) #tn.Tensor
```

tensornetwork/tensor.py
Thenerdstation: Doesn't matter too much tbh. Just guarantees the function to be pure.

tensornetwork/ncon_interface.py
Thenerdstation: Fixed.

Thenerdstation:
> I suppose later we could have finalize take a dict argument specifying a permutation of the natural contraction order, and/or specifications of an optimization algorithm

Yes these all sound like good things to add! The contraction/optimization algorithm is definitely something we'd want.

> Ooh, we could also later modify this so that the contraction pattern could be specified or retrieved independently of the concrete tensors

Oh yeah that's sick.

Thenerdstation:
> contract

Problem is we already have a `tn.contract` :/

alewis:
Haha fair enough. What's it do?

Thenerdstation:
Contacts a single edge, literally the most useless method in the library lol.

alewis:
Is it too widely-used to change? I like it better for this, but not so much
better that breaking stuff would be worth it.

tensornetwork/tensor.py
alewis: I see. I suppose that is important, but it could have significant overhead. Let's leave it though yes, I think there's no other way.

Thenerdstation:
I'll merge this as-is and open an issue for renaming.

PR #872, opened 2020-11-04
Contraction path node

mganahl:
Add `contract_path` and `path_solver` for use with `Node`


mganahl:
Hey @Thenerdstation pls have a look at this. This is something that we needed for a long time. Tests will be added after your had a chance to look at this. This should be pulled in ASAP because I'd make it part of a presentation

tensornetwork/contractors/opt_einsum_paths/path_contractors.py
```diff
@@ -290,3 +290,82 @@ def custom(nodes: Iterable[AbstractNode],
...
+  elif algorithm == "branch":
+    alg = functools.partial(
+      opt_einsum.paths.branch, memory_limit=memory_limit, nbranch=nbranch)
+  elif algorithm == "greedy":
+    alg = functools.partial(opt_einsum.paths.greedy, memory_limit=memory_limit)
+  elif algorithm == "auto":
+    n = len(list(nodes))  #pytype thing
+    _nodes = nodes
+    if n <= 1:
+      raise ValueError(f"network is trivial with {n} nodes. ")
```
Thenerdstation, line 325: Prefer to just return a trivial empty path.

Thenerdstation (approved):
Looks good. Only concern is passing in the algorithm as a string feels dirty, but I don't expect us to add a ton more so this should be fine for now.

Thenerdstation:
Also yes please add tests

mganahl:
> Looks good. Only concern is passing in the algorithm as a string feels dirty, but I don't expect us to add a ton more so this should be fine for now.
It's not beautiful, but the advantage is that the user doesn't have to pull in a module for that.

Thenerdstation:
> It's not beautiful, but the advantage is that the user doesn't have to pull in a module for that.

Agreed. Simplicity is a virtue.

Thenerdstation:
Test + passing build and I'll merge this in.

If you also want, I can also push version 0.4.3 once this is in

mganahl:
> If you also want, I can also push version 0.4.3 once this is in

That would be great!

Issue #840, opened 2020-09-28
Add power to each backend and as a Tensor method in linalg


alewis:
We should support the numpy function np.power.

alewis:
@wkostuch how's this sound? I could help you through it if you want a good way to get started!

wkostuch:
@alewis, this looks good, thanks!  I'd love to have a go at it.  

I'm excited to (hopefully) start contributing, and if something comes up I'll be sure to contact you.

alewis:
Great! The first step would be to add the function to the abstract backend.

To do this, first fork TensorNetwork, track the google repo as a remote, and start a new branch for your modification (lemme know if you need help with that part).

Within this new branch, go to `tensornetwork/backends/abstract_backend.py`, where various "empty" backend functions are implemented. These implementations just raise `NotImplementedError` when called. So the first step would be to add an equivalent for `power`.

Next you check `backend_tests.py`, which contains a bunch of tests that make sure `NotImplementedError` is indeed called, and add your own for the relevant function. Make sure your tests work: `cd tensornetwork`, `pytest ./backends/backend_tests.py`.

That would be enough for your first PR! Commit, push, and navigate back to this repo, where there should be an option to merge your recent push.

wkostuch:
So a pull request with just the "empty" function (and corresponding test) that raises an error and then go from there after that?  

I wrote the code for that, but then when trying to run the tests found out that one of the dependencies doesn't play nice with Windows, so I'm in the midst of setting up a Linux environment to get around that.  Once that's done I'll run the tests and keep on working.  Thanks again for all the guidance, I really appreciate it!

alewis:
Yes that's right, just the empty function at first. In general we try to keep pull requests to 200 lines or shorter, excluding test code, although sometimes this is not possible. That's because all the pull requests are to be peer reviewed, and it's hard for the reviewer to catch things when they are longer than this.

Regarding Windows, if you are using Windows 10 there is a so called "Windows Subsystem for Linux" that essentially allows  you to run Ubuntu inside Windows as an app. I used to develop this way - it works pretty well and while some setup is required, it's a lot easier than setting up a partition and so on, especially on a laptop with a GPU. There's instructions to get it working here: https://docs.microsoft.com/en-us/windows/wsl/install-win10

wkostuch:
Makes sense!  I got WSL running, so the empty function and test are now waiting for a review.

alewis:
[...]

PR #884, opened 2020-12-05
Added power function to Jax backend and test.

LuiGiovanni:
This is in relation to #840
Specifically adding to the Jax backend files, where I implement a power function and added the test function for said power function.


alewis (approved):
LGTM

PR #889, opened 2020-12-18
Added power function to backend to new backends

LuiGiovanni:
This PR implements the power function specifically to symmetric and PyTorch backends with their respective test.

The changes were implemented in the files:
```
tensornetwork/backends/pytorch/*
tensornetwork/backends/symmetric/*
```
This is in regards to the issue #840.

### DIFF
--- a/tensornetwork/backends/pytorch/pytorch_backend.py
+++ b/tensornetwork/backends/pytorch/pytorch_backend.py
@@ -475,3 +475,17 @@ class PyTorchBackend(abstract_backend.AbstractBackend):
   
   def item(self, tensor):
     return tensor.item()
+
+  def power(self, a: Tensor, b: Tensor) -> Tensor:
+    """
+    Returns the power of tensor a to the value of b.
+    In the case b is a tensor, then the power is by element
+      with a as the base and b as the exponent.
+    In the case b is a scalar, then the power of each value in a
+      is raised to the exponent of b.
+
+    Args:
+      a: The tensor that contains the base.
+      b: The tensor that contains the exponent or a single scalar.
+    """
+    return a ** b
--- a/tensornetwork/backends/symmetric/symmetric_backend.py
+++ b/tensornetwork/backends/symmetric/symmetric_backend.py
@@ -689,3 +689,17 @@ class SymmetricBackend(abstract_backend.AbstractBackend):
     if (tensor1.ndim != 2) or (tensor2.ndim != 2):
       raise ValueError("inputs to `matmul` have to be matrices")
     return tensor1 @ tensor2
+
+  def power(self, a: Tensor, b: Tensor) -> Tensor:
+    """
+    Returns the power of tensor a to the value of b.
+    In the case b is a tensor, then the power is by element
+      with a as the base and b as the exponent.
+    In the case b is a scalar, then the power of each value in a
+      is raised to the exponent of b.
+
+    Args:
+      a: The tensor that contains the base.
+      b: The tensor that contains the exponent or a single scalar.
+    """
+    return a ** b
--- a/tensornetwork/backends/symmetric/symmetric_backend_test.py
+++ b/tensornetwork/backends/symmetric/symmetric_backend_test.py
@@ -610,6 +610,19 @@ def test_addition_raises(R, dtype, num_charges):
 
 
 @pytest.mark.parametrize("dtype", np_dtypes)
+@pytest.mark.parametrize("num_charges", [1, 2])
+def test_power(dtype, num_charges):
+  np.random.seed(10)
+  R = 4
+  backend = symmetric_backend.SymmetricBackend()
+  a = get_tensor(R, num_charges, dtype)
+  b = BlockSparseTensor.random(a.sparse_shape)
+  expected = np.power(a.data, b.data)
+  actual = backend.power(a.data, b.data)
+  np.testing.assert_allclose(expected, actual)
+
+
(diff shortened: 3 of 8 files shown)


LuiGiovanni:
@mganahl @alewis Hello! here is the new PR where you may find the new changes and the tensordot issue. I await your feedback and thank you for the oportunity

LuiGiovanni:
Hello @alewis @mganahl! Running the test from pylint seems to cause and issue which I cannot resolve, it states that Python does not exist in module tensorflow, I assume the issue is init.py thought updating it with a small change such as adding a `.` does not seem to solve the issue, also my pylint returns a 10/10 both locally and here in GitHub's CI, the difference being that I do not get this issue locally. 

In my search I found that this issue is not some enclosed issue, I found some people raising this issue as well in other places here are some examples:

- https://github.com/Unity-Technologies/ml-agents/issues/1411
- https://stackoverflow.com/questions/40334643/no-name-in-module-error-from-pylint
- https://pycodequ.al/docs/pylint-messages/e0611-no-name-in-module.html

I will continue my search for a solution, my main problem being that we can't resolve the tensordot issue without resolving this first, I would love some help or feedback. Thank you so much in advance for all the help and support you have given me!

tensornetwork/backends/pytorch/pytorch_backend.py
```diff
@@ -475,3 +475,17 @@ def sign(self, tensor: Tensor) -> Tensor:
   
   def item(self, tensor):
     return tensor.item()
+
+  def power(self, a: Tensor, b: Tensor) -> Tensor:
+    """
+    Returns the power of tensor a to the value of b.
```
mganahl, line 481: the wording of this docstring seems somewhat imprecise. Can you fix this?

tensornetwork/backends/symmetric/symmetric_backend.py
```diff
@@ -689,3 +689,17 @@ def matmul(self, tensor1: Tensor, tensor2: Tensor):
     if (tensor1.ndim != 2) or (tensor2.ndim != 2):
       raise ValueError("inputs to `matmul` have to be matrices")
     return tensor1 @ tensor2
+
+  def power(self, a: Tensor, b: Tensor) -> Tensor:
+    """
+    Returns the power of tensor a to the value of b.
```
mganahl, line 695: The `__pow__`  method of `BlockSparseTensor` does currently not support broadcasting of any type, hence `b` has to be a scalar. Pls  modify the docstring accordingly.

tensornetwork/backends/symmetric/symmetric_backend_test.py
```diff
@@ -609,6 +609,19 @@ def test_addition_raises(R, dtype, num_charges):
...
 
 
+@pytest.mark.parametrize("dtype", np_dtypes)
+@pytest.mark.parametrize("num_charges", [1, 2])
+def test_power(dtype, num_charges):
+  np.random.seed(10)
+  R = 4
+  backend = symmetric_backend.SymmetricBackend()
+  a = get_tensor(R, num_charges, dtype)
+  b = BlockSparseTensor.random(a.sparse_shape)
```
mganahl, line 619: Pls modify the test so that `b` is a numpy-scalar

mganahl (changes requested):

tensornetwork/backends/symmetric/symmetric_backend.py
LuiGiovanni: I'm a bit confused with this request since this test works fine for me with no failures could you elaborate on what the problem is please?

tensornetwork/backends/pytorch/pytorch_backend.py
LuiGiovanni: >  """
>     Returns the exponentiation of tensor a raised to b.
>     In the case b is a tensor, then the power is by element
>       with a as the base and b as the exponent.

>     Args:
>       a: The tensor that contains the base.
>       b: The tensor that contains a single scalar.
>     """

This is the change I made to it, once we resolve the other request I will push the changes

tensornetwork/backends/symmetric/symmetric_backend.py
mganahl: The reason the tests passes is due to a bug in the test (I had actually missed that in the earlier reviewa, see below).
in numpy, `a**b` below does different things depending on how `a` and `b` look like. Generally, numpy applies broadcasting (https://numpy.org/doc/stable/user/basics.broadcasting.html) rules when computing this expression. Typically, this approach is relatively widely used by other libraries as well (e.g. tensorflow or pytorch). Supporting broadcasting for`BlockSparseTensor` is however more complicated because this class has some non-trivial internal structure that needs to be respected  by such operations. Currently we just don't support it. The only case supported by `BlockSparseTensor` is when `b` is a scalar-type object.

mganahl: pls add some test ensuring that `a` and `b` are `BlockSparseTensor`s

tensornetwork/backends/symmetric/symmetric_backend_test.py
mganahl: `backend.power` should only take `BlockSparseTensor` objects, pls fix.

mganahl (changes requested):
thanks again for the PR! Left some comments to be fixed. Thanks!

Issue #852, opened 2020-10-12
Add cholesky decomposition


alewis:
We should support the Cholesky decomposition per https://numpy.org/doc/stable/reference/generated/numpy.linalg.cholesky.html#numpy.linalg.cholesky, adding a pivot argument in the same way as qr and svd.

Dbhasin1:
Hello, can this issue be assigned to me?

mganahl:
Sure thing!

Dbhasin1:
Hello, 
what exactly do you mean by the last line, "adding a pivot argument in the same way as qr and svd."
Thanks

Dbhasin1:
Could you please tell me exactly where I need to add this function?

alewis:
Hi @Dbhasin1, 

The function and tests first need to be added to the various backends, which live in tensornetwork/backends/blah. The first is the AbstractBackend, a base class that just throws NotImplementedError. The others all wrap functions from other numerical packages, the idea being to provide a means to quickly switch between these. 

Inside the backend files you will see many other functions, such as qr. Our QR differs from e.g. NumPy's by the presence of a pivot argument that matricizes the input. Cholesky should have a similar mechanism.

Once added to the backends, the function should also be added to the tn.Tensor interface, which lives in tensornetwork/tensor.py with functions defined in tensornetwork/linalg/blah

Dbhasin1:
I don't see the svd or qr functions, however, in the tn.Tensor interface you mentioned. Is there some reason for that?

LuiGiovanni:
Hello there @alewis I'm interested in working on this issue has a PR been done? If not I would like to work on this.

LuiGiovanni:
Hello, please ignore this PR I made a mistake, the changes implemented for the Cholesky decomposition will be from a different PR.

LuiGiovanni:
Hello, I am sorry for the mess but #886 is the correct PR I was working on disregard #885 thank you.

LuiGiovanni:
@alewis @mganahl I have added the NotImplementedError function and it's respective test and made a PR for you to review.

alewis:
Looks good! Just a small fix.

LuiGiovanni:
@alewis Alright and what would that fix be? Also I will start implementing the Cholesky decomposition, quick question, why does Jax not have it's own decompositions file?

mganahl:
it uses the numpy decompositions.py file

LuiGiovanni:
[...]

PR #890, opened 2020-12-23
Implemented Cholesky decomposition to numpy and tensorflow backends

LuiGiovanni:
Added the decomposition function following the idea of having a pivot on the tensor, this is in reference to issue #852. which we then reshape into a matrix for the Cholesky function used in NumPy linear algebra. which you may find [here](https://github.com/numpy/numpy/blob/v1.19.0/numpy/linalg/linalg.py#L673-L765) for reference.

I Will start working on adding the function to the missing backend which is symmetric since Jax uses NumPy's decomposition file. I await any feedback you may have and that you for the help!

### DIFF
--- a/tensornetwork/backends/numpy/decompositions.py
+++ b/tensornetwork/backends/numpy/decompositions.py
@@ -122,3 +122,22 @@ def rq(
   r = np.reshape(r, list(left_dims) + [center_dim])
   q = np.reshape(q, [center_dim] + list(right_dims))
   return r, q
+
+
+def cholesky(
+    np, # TODO: Typing
+    tensor: Tensor,
+    pivot_axis: int
+) -> Tuple[Tensor, Tensor]:
+  """
+  Computes the Cholesky decomposition of a tensor
+
+  See tensornetwork.backends.tensorflow.decompositions for details.
+  """
+  left_dims = np.shape(tensor)[:pivot_axis]
+  right_dims = np.shape(tensor)[pivot_axis:]
+  tensor = np.reshape(tensor,
+                      [numpy.prod(left_dims),
+                       numpy.prod(right_dims)])
+  L = np.linalg.cholesky(tensor)
+  return L
--- a/tensornetwork/backends/numpy/decompositions_test.py
+++ b/tensornetwork/backends/numpy/decompositions_test.py
@@ -52,6 +52,12 @@ class DecompositionsTest(tf.test.TestCase):
       q, r = decompositions.qr(np, random_matrix, 1, non_negative_diagonal)
       self.assertAllClose(q.dot(r), random_matrix)
 
+  def test_cholesky(self):
+    random_matrix = np.random.rand(10, 10)
+    for non_negative_diagonal in [True, False]:
+      L = decompositions.cholesky(np, random_matrix, 1)
+      self.assertAllClose(np.cholesky(random_matrix), L)
+
   def test_max_singular_values(self):
     random_matrix = np.random.rand(10, 10)
     unitary1, _, unitary2 = np.linalg.svd(random_matrix)
--- a/tensornetwork/backends/abstract_backend.py
+++ b/tensornetwork/backends/abstract_backend.py
@@ -1029,4 +1029,4 @@ class AbstractBackend:
                              non_negative_diagonal: bool = False) -> \
                              Tuple[Tensor, Tensor]:
     raise NotImplementedError(
-        f"Backend {self.name} han not implemented cholesky_decomposition.")
+        f"Backend {self.name} has not implemented cholesky_decomposition.")
--- a/tensornetwork/backends/abstract_backend.py
+++ b/tensornetwork/backends/abstract_backend.py
@@ -1023,6 +1023,9 @@ class AbstractBackend:
     """
     raise NotImplementedError("Backend {self.name} has not implemented item")
 
-  def chsky(self, tensor: Tensor, pivot_axis: int = -1, non_negative_diagonal: bool = False) -> Tuple[Tensor, Tensor]:
+  def chsky(self, tensor: Tensor, 
+            pivot_axis: int = -1, 
+            non_negative_diagonal: bool = False) -> 
+            Tuple[Tensor, Tensor]:
     """Computes the Cholskey decomposition of a tensor."""
     raise NotImplementedError("Backend '{}' has not implemented chsky.".format(self.name))
--- a/tensornetwork/backends/abstract_backend.py
+++ b/tensornetwork/backends/abstract_backend.py
@@ -1029,4 +1029,4 @@ class AbstractBackend:
                non_negative_diagonal: bool = False) -> \
                Tuple[Tensor, Tensor]:
     raise NotImplementedError(
-        f"Backend {self.name} has not implemented cholesky_decomposition.")
+        f"Backend {self.name} has not implemented cholesky.")
--- a/tensornetwork/backends/abstract_backend.py
+++ b/tensornetwork/backends/abstract_backend.py
@@ -1022,3 +1022,7 @@ class AbstractBackend:
       The value in tensor.
     """
     raise NotImplementedError("Backend {self.name} has not implemented item")
+
+  def chsky(self, tensor: Tensor, pivot_axis: int = -1, non_negative_diagonal: bool = False) -> Tuple[Tensor, Tensor]:
+    """Computes the Cholskey decomposition of a tensor."""
+    raise NotImplementedError("Backend '{}' has not implemented chsky.".format(self.name))
--- a/tensornetwork/backends/abstract_backend.py
+++ b/tensornetwork/backends/abstract_backend.py
@@ -1023,9 +1023,10 @@ class AbstractBackend:
     """
     raise NotImplementedError("Backend {self.name} has not implemented item")
 
-  def chsky(self, tensor: Tensor, 
-            pivot_axis: int = -1, 
-            non_negative_diagonal: bool = False) -> 
-            Tuple[Tensor, Tensor]:
-    """Computes the Cholskey decomposition of a tensor."""
-    raise NotImplementedError("Backend '{}' has not implemented chsky.".format(self.name))
+  def cholesky_decomposition(self, 
+                             tensor: Tensor,
+                             pivot_axis: int = -1,
+                             non_negative_diagonal: bool = False) -> \
+                             Tuple[Tensor, Tensor]:
+    raise NotImplementedError(
+        f"Backend {self.name} han not implemented cholesky_decomposition.")
--- a/tensornetwork/backends/abstract_backend.py
+++ b/tensornetwork/backends/abstract_backend.py
@@ -1023,10 +1023,10 @@ class AbstractBackend:
     """
     raise NotImplementedError("Backend {self.name} has not implemented item")
 
-  def cholesky_decomposition(self, 
-                             tensor: Tensor,
-                             pivot_axis: int = -1,
-                             non_negative_diagonal: bool = False) -> \
-                             Tuple[Tensor, Tensor]:
+  def cholesky(self, 
+               tensor: Tensor,
+               pivot_axis: int = -1,
+               non_negative_diagonal: bool = False) -> \
+               Tuple[Tensor, Tensor]:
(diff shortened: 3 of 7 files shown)


LuiGiovanni:
@alewis @mganahl Hello! I implemented the Cholesky decomposition to various backends following the idea of a pivot value for the tensors, with their respective tests where I defined a matrix that is assured to be positive-definite and also a hermitian matrix:

> [1, 0]
> [0, 1]

I assumed we won't be validating if the matrices are hermitian or positive-definite since there could be some performance cost in doing so (at least for positive-definite checking) the addition of a hermitian check would be nice to have.

A Pylint error has been occurring in the test runs, but it is somewhere in the code where I haven't made changes to. the file in question is `tensornetwork/tn_keras/conv2d_mpo.py` which seems to only be affected in versions 3.6 and 3.7, this issue is also occurring in my other PR #889 maybe freezing the version could fix this but this is a lazy solution and would not solve it for all cases, so I will wait for your feedback to see what can be done here. Here is an image of the error for reference.

image

tensornetwork/backends/numpy/decompositions.py
```diff
@@ -122,3 +122,28 @@ def rq(
   r = np.reshape(r, list(left_dims) + [center_dim])
   q = np.reshape(q, [center_dim] + list(right_dims))
   return r, q
+
+
+def cholesky(
+    tf: Any,
```
mganahl, line 128: Rename `tf` to `np` (the user passes the `numpy` module here)

tensornetwork/backends/numpy/decompositions_test.py
```diff
@@ -52,6 +52,13 @@ def test_qr(self):
       q, r = decompositions.qr(np, random_matrix, 1, non_negative_diagonal)
       self.assertAllClose(q.dot(r), random_matrix)
 
+  def test_cholesky(self):
+    #Assured positive-definite hermitian matrixs
+    random_matrix = [[1.0, 0], [0, 1.0]]
```
mganahl, line 57: Using the identity matrix  as "random" matrix is not an ideal test case. You can generate a positive definite matrix e.g. by
```python 
A = np.random.rand(D,D)
A  = A @ A.T.conj()
```

mganahl: btw. also test this for all supported dtypes

tensornetwork/backends/numpy/decompositions.py
```diff
@@ -122,3 +122,28 @@ def rq(
...
+
+  See tensornetwork.backends.tensorflow.decompositions for details.
+  """
+  left_dims = tf.shape(tensor)[:pivot_axis]
+  right_dims = tf.shape(tensor)[pivot_axis:]
+  tensor = tf.reshape(tensor,
+                      [tf.reduce_prod(left_dims),
+                       tf.reduce_prod(right_dims)])
+  L = tf.linalg.cholesky(tensor)
+  if non_negative_diagonal:
```
mganahl, line 144: This is unnecessary. The cholesky decomposition should already return a lower-triangular matrix with a real, positive diagonal.

tensornetwork/backends/numpy/decompositions.py
```diff
@@ -122,3 +122,28 @@ def rq(
   r = np.reshape(r, list(left_dims) + [center_dim])
   q = np.reshape(q, [center_dim] + list(right_dims))
   return r, q
+
+
+def cholesky(
+    tf: Any,
+    tensor: Tensor,
+    pivot_axis: int,
+    non_negative_diagonal: bool
```
mganahl, line 131: remove `non_negative` argument

tensornetwork/backends/pytorch/decompositions.py
```diff
@@ -121,6 +121,29 @@ def svd(
   return u, s, vh, s_rest
 
 
+def cholesky(
+    torch: Any,
+    tensor: Tensor,
+    pivot_axis: int,
+    non_negative_diagonal: bool = False
```
mganahl, line 128: remove `non_negative_diagonal` (see above)

tensornetwork/backends/pytorch/decompositions.py
```diff
@@ -121,6 +121,29 @@ def svd(
...
+  Computes the Cholesky decomposition of a tensor
+
+  See tensornetwork.backends.tensorflow.decompositions for details.
+  """
+  left_dims = np.shape(tensor)[:pivot_axis]
+  right_dims = np.shape(tensor)[pivot_axis:]
+  
+  tensor = torch.reshape(tensor, (np.prod(left_dims), np.prod(right_dims)))
+  L = torch.cholesky(tensor)
+  if non_negative_diagonal:
```
mganahl, line 140: remove (see above)

tensornetwork/backends/pytorch/decompositions_test.py
```diff
@@ -42,6 +42,13 @@ def test_expected_shapes_rq():
     assert r.shape == (2, 3, 6)
     assert q.shape == (6, 4, 5)
 
+def test_cholesky():
+  #Assured positive-definite hermitian matrix
+  random_matrix = np.array([[1.0, 0], [0, 1.0]])
```
mganahl, line 47: use better test matrix (see above)

tensornetwork/backends/pytorch/decompositions_test.py
```diff
@@ -42,6 +42,13 @@ def test_expected_shapes_rq():
...
+  random_matrix = torch.from_numpy(random_matrix)
+  for non_negative_diagonal in [True, False]:
```
mganahl, line 49: see above

tensornetwork/backends/tensorflow/decompositions.py
```diff
@@ -226,3 +226,28 @@ def rq(
   r = tf.reshape(r, tf.concat([left_dims, [center_dim]], axis=-1))
   q = tf.reshape(q, tf.concat([[center_dim], right_dims], axis=-1))
   return r, q
+
+
+def cholesky(
+    tf: Any,
+    tensor: Tensor,
+    pivot_axis: int,
+    non_negative_diagonal: bool
```
mganahl, line 235: see above

tensornetwork/backends/tensorflow/decompositions.py
```diff
@@ -226,3 +226,28 @@ def rq(
...
+  Returns the Cholesky decomposition of a tensor which we treat as a 
+  square matrix
+  """
+  left_dims = tf.shape(tensor)[:pivot_axis]
+  right_dims = tf.shape(tensor)[pivot_axis:]
+  tensor = tf.reshape(tensor,
+                      [tf.reduce_prod(left_dims),
+                       tf.reduce_prod(right_dims)])
+  L = tf.linalg.cholesky(tensor)
+  if non_negative_diagonal:
```
mganahl, line 248: see above

tensornetwork/backends/tensorflow/decompositions_test.py
```diff
@@ -54,6 +54,13 @@ def test_qr(self):
       q, r = decompositions.qr(tf, random_matrix, 1, non_negative_diagonal)
       self.assertAllClose(tf.tensordot(q, r, ([1], [0])), random_matrix)
 
+  def test_cholesky(self):
+    #Assured positive-definite hermitian matrix
+    random_matrix = [[1.0, 0], [0, 1.0]]
```
mganahl, line 59: modify test, see above

mganahl (changes requested):

mganahl:
> A Pylint error has been occurring in the test runs, but it is somewhere in the code where I haven't made changes to. the file in question is `tensornetwork/tn_keras/conv2d_mpo.py` which seems to only be affected in versions 3.6 and 3.7, this issue is also occurring in my other PR #889 maybe freezing the version could fix this but this is a lazy solution and would not solve it for all cases, so I will wait for your feedback to see what can be done here. Here is an image of the error for reference.
> 
> image

Thanks for pointing this out. Likely this is due to a new tensorflow release. I'll fix it.

tensornetwork/backends/numpy/decompositions.py
alewis: ya

tensornetwork/backends/numpy/decompositions.py
alewis: (it only makes sense for QR)

tensornetwork/backends/numpy/decompositions_test.py
alewis: do `np.random.rand(D, D).astype(dtype)` to get your fav type. if it's imaginary you might need to initialize the real and imaginary parts separately.

put `np.random_seed(10)` (or any other fixed integer) before the `rand` so that the matrix is fixed.

tensornetwork/backends/tensorflow/decompositions.py
```diff
@@ -226,3 +226,28 @@ def rq(
...
+    phases = tf.math.sign(tf.linalg.diag_part(L))
+    L = phases[:, None] * L
+  center_dim = tf.shape(L)[1]
+  L = tf.reshape(L, tf.concat([left_dims, [center_dim]], axis=-1))
+  return L
```
alewis, line 253: add a newline here (at the end of the file)

tensornetwork/backends/numpy/decompositions_test.py
LuiGiovanni: There is no `dtype` defined in any of the tests, should I define one myself?

LuiGiovanni:
Hi there @mganahl @alewis In the end I implemented the test following your requested changes and tried these changes out, I'd like your opinion on them. The biggest difference being: 

image

`torch.from_numpy(random_ matrix)` since it wasn't reading it like a tensor. Let me know what you think. Thank you!

tensornetwork/backends/numpy/decompositions_test.py
```diff
@@ -52,6 +54,13 @@ def test_qr(self):
       q, r = decompositions.qr(np, random_matrix, 1, non_negative_diagonal)
       self.assertAllClose(q.dot(r), random_matrix)
 
+  def test_cholesky(self):
+    #Assured positive-definite hermitian matrixs
+    random_matrix = np.random.rand(10, 10)
+    random_matrix = random_matrix @ random_matrix.T.conj()
+    L = decompositions.cholesky(tf, random_matrix, 1)
```
mganahl, line 61: you are passing the tensorflow module here, but it should be numpy

tensornetwork/backends/numpy/decompositions.py
```diff
@@ -122,3 +122,22 @@ def rq(
...
+) -> Tuple[Tensor, Tensor]:
+  """
+  Computes the Cholesky decomposition of a tensor
+
+  See tensornetwork.backends.tensorflow.decompositions for details.
+  """
+  left_dims = np.shape(tensor)[:pivot_axis]
+  right_dims = np.shape(tensor)[pivot_axis:]
+  tensor = np.reshape(tensor,
+                      [np.reduce_prod(left_dims),
```
mganahl, line 140: `np` is the `numpy` module. `numpy` does not have `reduce_prod`, that's a `tensorflow` routine. This only passes the test because in `decompositions_test.py` you are erroneously passing the `tensorflow` module instead of the `numpy` module to `numpy.decompositions.cholesky`. pls fix this

tensornetwork/backends/pytorch/decompositions.py
```diff
@@ -121,6 +121,23 @@ def svd(
...
+  """
+  Computes the Cholesky decomposition of a tensor
+
+  See tensornetwork.backends.tensorflow.decompositions for details.
+  """
+  left_dims = list(tensor.shape)[:pivot_axis]
+  right_dims = list(tensor.shape)[pivot_axis:]
+
+  tensor = torch.reshape(tensor, (np.prod(left_dims), np.prod(right_dims)))
+  L = np.linalg.cholesky(tensor)
```
mganahl, line 138: you have to call the `torch` version of `choleksy` here, not `numpy`.

mganahl (changes requested):

mganahl:
Hi @LuiGiovanni, once the comments are fixed we can merge this

PR #898, opened 2021-01-12
SU(2) Exact Diagonalization

qsukhi:
Added a function to compute the low energy eigenstates of a local, SU(2)-symmetric Hamiltonian, in a way that exploits the SU(2) symmetry. Also added some generic SU(2) functionality e.g. functions to compute Clebsch-Gordan coefficients, Fusion Rules, Fusing two indices etc.


qsukhi:
@googlebot I signed it!

mganahl:
thanks @qsukhi I'll take a look!

mganahl:
Hi @qsukhi, I'm working through the code. It will take me a bit of time since it is a lot of code. In the meantime, can you change indenting to two spaces instead of four? I wonder why the linter didn't catch this.
Also, I noticed that there is a binary file in this PR, is this intentional?

qsukhi:
Thanks, @mganahl. 
-I've reduced the indent to 2 spaces. 
-The binary file is a table of factors that come from the Clebsch-Gordan part of SU(2)-symmetric tensor contractions. It is not necessary, as the function Su2ExactDiagonalization() can build the factor tables if the user specifies so (as illustrated in Su2ExactDiagonalization/main.py).

PR #912, opened 2021-03-24
fix apply_twosite_gate

mganahl:
Changes `base_mps.apply_twosite_gate` to use QR decomp is no truncation is required.


jacksonwb:
Are there any changes here? File diffs are empty?

mganahl:
hmm, strange

mganahl:
OK, there was another change that you or adam had reviewed already that was depending on this change. So this got merged already ... . Closing the PR

PR #914, opened 2021-03-25
Fix trace for PyTorch backend

merajhashemi:
Currently, the trace operator for PyTorch only supports up to 25-dimensional inputs due to using `einsum`. This PR fixes this issue.


merajhashemi:
@googlebot I signed it!

mganahl:
@merajhashemi thanks for the PR. Looks good from my side. The build error is some flaky in the jax-backend. Let me first fix this test and we'll merge this in then.

Issue #844, opened 2020-09-29
Support pinv


alewis:
We need `np.linalg.pinv`, in both the backends and on `Tensor`.

virajone:
@alewis Hey I'm new to the project and would love to contribute.  Given this seems like a good first issue, can you please provide the scope for this one ?

alewis:
Sure, happy to have you. The goal would be to add the numpy function `pinv` https://numpy.org/doc/stable/reference/generated/numpy.linalg.pinv.html to the library via each of the various backends. The first step is to go to `tensornetwork/backends/abstract_backend.py` and to add it there, using the other existing functions as guides, and then to the tests in the same folder. Why don't you give that a shot and we'll sync up?

virajone:
@alewis Thanks for the details. Seems fairly straightforward. I will get back to you once I'm done with the first one. Can you assign me the issue in the meantime ?

alewis:
Done!

virajone:
@alewis , Added a PR #850 for adding PINV to abstract_backend. Your feedback would be really appreciated! I shall also start working on adding it to the other backends and eventually the tensor, per your approval.

virajone:
@alewis I have made the requested changes. In case you will be approving  the #850 , shall I proceed to adding PINV to other backends ?

raunakbhupal:
Hi. I'm new over here.
If this issue is still open then can I add PINV to other backends?

alewis:
That’d be great!

raunakbhupal:
Since I'm new and this is my first time
Could you please direct me to which other backends shall I make those changes?

GouravWadhwa:
Hi @alewis, I am new to Open-source and would love to contribute.
Please let me know which other backends do we need to add the PINV function

alewis:
I believe it has only been added to the abstract backend (the base class). It needs to be added to all the other ones. Maybe you guys can divide the work?

krbhanushali:
@alewis I added a PR https://github.com/google/TensorNetwork/pull/935. I ran the pytests but am not sure how to run the workflow checks. Please suggest any improvements which can be made in this pr.

GSAUC3:
@alewis hi, is this issue still open? if, yes, may I work on this issue, since this issue was opened nearly 3 years ago?

PR #918, opened 2021-04-02
Pinv numpy backend

raunakbhupal:
This is done with respect to issue: #844
`pinv` function is added to `tensornetwork/backends/tensorflow/numpy_backend.py`
The `pinv` function has been referred from https://numpy.org/doc/stable/reference/generated/numpy.linalg.pinv.html
Please do have a look and let me know if should make any other change.
Thank you.

### DIFF
--- a/tensornetwork/backends/tensorflow/tensorflow_backend.py
+++ b/tensornetwork/backends/tensorflow/tensorflow_backend.py
@@ -444,5 +444,4 @@ class TensorFlowBackend(abstract_backend.AbstractBackend):
     Returns:
      tensor: The pseudo inverse of tensor.
          """
-    raise NotImplementedError(
-        "Backend '{}' has not implemented pinv".format(self.name))
+    return np.linalg.pinv(tensor)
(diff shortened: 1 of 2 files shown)


tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -430,3 +430,18 @@ def eps(self, dtype: Type[np.number]) -> float:
...
+    
+    Args:
+     tensor: A tensor.
+     rcond: Cutoff for small singular values.
+     hermitian(optional): If True, matrix provided is assumed to be Hermitian (symmetric if real-valued). Defaults to False.
+
+    Returns:
+     tensor: The pseudo inverse of tensor.
+         """
+    return np.linalg.pinv(tensor)
```
mganahl, line 447: you should use the tensorflow version here

mganahl (changes requested):
Please also add `pinv` to all other backends, or raise a `NotImplementedError`,
and add tests for each function. Thanks!

tensornetwork/backends/tensorflow/tensorflow_backend.py
raunakbhupal: Hi.
In a [StackOverflow article](https://stackoverflow.com/questions/60025950/tensorflow-pseudo-inverse-doesnt-work-for-complex-matrices/60128892#60128892) someone has mentioned that the Tensorflow version doesn't work for complex matrices and that's why I went with the Numpy Version.
So shall I still use the Tensorflow version only?

mganahl: yes, please use tensorflow. The point here is to use backend-native operations.

PR #923, opened 2021-05-20
Export typing information

JanLuca:
This commit adds the flag that the package contains typing
information. Therefore other packages can use the typing data to check their
code by mypy.

Ref.: https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages

Signed-off-by: Jan Luca Naumann <j.naumann@fu-berlin.de>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,5 +43,6 @@ setup(
     description=description,
     long_description=long_description,
     long_description_content_type="text/markdown",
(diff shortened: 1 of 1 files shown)


JanLuca:
@googlebot I signed it!

mganahl:
Thanks for the PR! I'll fix the build and merge this.

Issue #853, opened 2020-10-12
Add Cholesky to symmetric backend


mganahl:

Shashwatpandey4:
Hi, i would love to work on this issue

PR #932, opened 2021-08-26
Symm cholesky

Shashwatpandey4:
this pull request is being made in response to issue #853


mganahl:
thanks for the PR, and sorry for the late reply! This issue is actually quite complicated to implement, and requires intricate knowledge of the code and also some physics background, so it is not the best place to start with. I'm closing this for now,
but if you'd like to give it a try anyways then pls feel free to reopen.

Issue #354, opened 2019-10-29
Other decomposition methods (CP, Tucker)


jypark0:
It would be great to have other general tensor decomposition methods besides SVD and QR, such as CP/tensor rank or Tucker. Thank you!

DefUs3r:
Hi @Thenerdstation can I work on this enhancement? Please let me know where I can start from.

PR #934, opened 2021-08-30
Update decompositions.py

Ayushparikh-code:
this PR solves #354


mganahl:
Thanks for the PR! I think #354 requires some careful thinking on our side about how we want to do this. I think the question posed there is if the tensornetwork repo should offer more general decompositions of the form of tensor-trains/matrix product states for all backends. We are already offering (limited) support for matrix product states and tensor tensor-train decompositions of neural network layers. Full fledged support though should likely be implemented natively though, not through the use of third-party libraries.

PR #935, opened 2021-08-31
Adds pinv to tensorflow backend (https://github.com/google/TensorNetwork/issues/844)

krbhanushali:
Added pinv to tensorflow_backend.py and corresponding tests to tensorflow_backend_tests,py

### DIFF
--- a/tensornetwork/backends/tensorflow/tensorflow_backend.py
+++ b/tensornetwork/backends/tensorflow/tensorflow_backend.py
@@ -430,3 +430,26 @@ class TensorFlowBackend(abstract_backend.AbstractBackend):
       float: Machine epsilon.
     """
     return tf.experimental.numpy.finfo(dtype).eps
+
+  def pinv(self, tensor: Tensor, rcond: float = 1E-15, hermitian: bool = False) -> Tensor:
+    """
+    Compute the Moore-Penrose/Pseudo inverse of a tensor.
+
+    Args:
+      tensor: A tensor.
+      rcond: Cutoff for small singular values
+      hermitian: If True, matrix provided in function argument is assumed to be
+                  Hermitian (symmetric if real-valued)
+                  This argument is not supported in the TensorFlow
+                  backend and an error will be raised if it is
+                  specified.
+
+    Returns:
+      tensor: The pseudo inverse of `tensor`
+    """
+    if hermitian != False:
+      errstr = (f"hermitian = {hermitian} must be False (the default)"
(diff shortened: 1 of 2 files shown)


tensornetwork/backends/tensorflow/tensorflow_backend_test.py
```diff
@@ -629,3 +629,15 @@ def test_power(dtype):
...
+  backend = tensorflow_backend.TensorFlowBackend()
+  np.random.seed(10)
+  tensor = np.random.rand(2, 3, 4)
+  rcond = np.random.rand(1)[0] * 1E-14
+  expected = tf.linalg.pinv(tensor, rcond)
+  actual = backend.pinv(tensor, rcond)
+  np.testing.assert_allclose(expected, actual)
+  expected = tf.linalg.pinv(tensor)
+  actual = backend.pinv(tensor)
+  np.testing.assert_allclose(expected, actual)
```
mganahl, line 643: add a test that `actual @ tensor` and `tensor @ actual` equal the identity (within the desired error)

tensornetwork/backends/tensorflow/tensorflow_backend.py
```diff
@@ -430,3 +430,26 @@ def eps(self, dtype: Type[np.number]) -> float:
       float: Machine epsilon.
     """
     return tf.experimental.numpy.finfo(dtype).eps
+
+  def pinv(self, tensor: Tensor, rcond: float = 1E-15, hermitian: bool = False) -> Tensor:
```
mganahl, line 434: I think you can remove the `hermitian` argument

mganahl (changes requested):
Thanks for the PR, looks good! Could you also add `pinv` to the `AbstractBackend` class, as well as `numpy` and `jax`  backends?

tensornetwork/backends/tensorflow/tensorflow_backend_test.py
krbhanushali: Would testing for all the criteria in https://en.wikipedia.org/wiki/Moore%E2%80%93Penrose_inverse be a better test? (As for an example given in tf's documentation, the product wasn't identity)

tensornetwork/backends/tensorflow/tensorflow_backend.py
krbhanushali: I had included it as it was an argument for numpy. I'll remove it in the next commit. 
Thanks for the review!

PR #938, opened 2021-09-22
Fix JAX convert_to_tensor in anticipation of upcoming change

jacksonwb:
[JAX] Replace uses of `isinstance(x, jax.numpy.ndarray)` with
`isinstance(x, (numpy.ndarray, jax.numpy.ndarray))` where
they were leading to test failures.

An upcoming change to JAX will make `isinstance(x, jax.numpy.ndarray)`
return true if and only if `x` is an instance of a JAX array type.

Previously `isinstance(x, jax.numpy.ndarray)` also returned true
for classic NumPy's `numpy.ndarray` objects as well. After the
upcoming change, it will return false for `numpy.ndarray` objects.
This change updates users of JAX who were depending on the
current behavior of the `isinstance` check to instead explicitly
check for `numpy.ndarray` instances as well.

These changes should have no effect on using `jax.numpy.ndarray` as
a type annotation. That does little and never has, although it is
possible that may change in the future. This change is strictly
about what `jax.numpy.ndarray` means to runtime `isinstance` checks.

### DIFF
--- a/tensornetwork/backends/jax/jax_backend.py
+++ b/tensornetwork/backends/jax/jax_backend.py
@@ -121,8 +121,8 @@ class JaxBackend(abstract_backend.AbstractBackend):
   def convert_to_tensor(self, tensor: Tensor) -> Tensor:
     if (not isinstance(tensor, (np.ndarray, jnp.ndarray))
         and not jnp.isscalar(tensor)):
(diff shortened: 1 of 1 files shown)


jacksonwb:
@mganahl  The lint errors are not from new changes. Have these been breaking every new PR?

mganahl:
Hi @jacksonwb closing this because I'm appending these changes to #939

PR #939, opened 2021-09-23
Fix build

mganahl:


jacksonwb:
Lol that TN layer accuracy test just breaks on like the last iteration of 50 😅. We should probably just remove that as a flaky test. That code is pretty fringe anyway... Thoughts?

mganahl:
yes, that test already has had problems in the past. Lemme take a look.

jacksonwb:
👌

PR #941, opened 2021-09-24
temporarily switch to tf.tensordot instead of tensordot2

### LINKED ISSUE #940: Buggy interplay between ``tf.vectorized_map`` and ``tn.contractors.*``
When a function ``f`` has, say ``tn.contractors.optimal()`` inside, the results of the function is of course right, but the results for ``tf.vectorized_map(f, batched_input)`` might be wrong. 
See a full demo script at  https://gist.github.com/refraction-ray/3a24e1ca09d9c2850d62b40cbf57946c. 
The results are always correct for individual input (jitted or not) and for ``jax.vmap`` interface but become wrong for ``tf.vectorized_map`` interface. I suspect the problem may be related to the interplay of ``tf.vectorized_map`` and the path solver provided by opteinsum, though I cannot see why they are entangled.

The output for the above demo script in tensorflow case is
```bash
$ python vectorized_map_bug.py tensorflow

3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34)
[GCC 7.3.0]
tensornetwork version: 0.4.5
tensorflow version: 2.5.0
tf.Tensor((-0.4161468+0.90929735j), shape=(), dtype=complex64)
tf.Tensor((-0.65364355-0.7568025j), shape=(), dtype=complex64)
jitted
tf.Tensor((-0.4161468+0.90929735j), shape=(), dtype=complex64)
tf.Tensor((-0.65364355-0.7568025j), shape=(), dtype=complex64)
mapfn
tf.Tensor([-0.4161468 +0.90929735j -0.65364355-0.7568025j ], shape=(2,), dtype=complex64)
vectorized map
tf.Tensor([0.99999994+0.j 0.99999994+0.j], shape=(2,), dtype=complex64)
```

mganahl:
`tf.matmul` with `transpose_a` and ` transpose_b` arguments has a bug for complex dtypes if wrapped into `tf.vectorized_map` (see https://github.com/tensorflow/tensorflow/issues/52148). Our modified version of `tf.tensordot`, `tensordot2`, uses a `tf.matmul` with these arguments to improve efficiency. This unfortunately causes the afore mentioned bug to hit us in certain use cases.
Hence we temporarily revert to using `tf.tensordot` in the `TensorflowBackend` directly.
fixes #940

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.6', '3.7', '3.8']
+        python-version: ['3.7', '3.8']
 
     steps:
     - uses: actions/checkout@v2
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,3 +1,4 @@
 [style]
 # TensorNetwork uses the yapf style
 based_on_style = yapf
+indent_width = 2
--- a/examples/dmrg/symmetric_dmrg.py
+++ b/examples/dmrg/symmetric_dmrg.py
@@ -148,9 +148,9 @@ if __name__ == '__main__':
   #        are ramped up during the simulation close to the boundaries.
   #
   #        The symmetric backend is for small bond dimensions typicall slower
-  #        Than other backends, due to inherent book-keeping overhead.
+  #        than other backends, due to inherent book-keeping overhead.
   #        In comparison with numpy, the two backends typically are of the same
-  #        speed for a bond dimension of D ~ 100. For value of D >~ 400, their
+  #        speed for a bond dimension of D ~ 100. For value of D >~ 400, the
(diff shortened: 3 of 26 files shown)


refraction-ray:
BTW, there is another call for ``tensordot2`` on L111 of the same file, and it also induce similar issues with ``tf.vectorized_map``

mganahl:
thanks, yes your're right! Fixed it

jacksonwb:
Should we deprecate the tensorflow2 module?

mganahl:
the `tensordot2` module applies some tricks which make it more efficient than `tf.tensordot`. From looking at the latest implementation of `tf.tensordot`, it seems that these improvements are still valid. I would say we leave it in for now.

jacksonwb (approved):

jacksonwb:
Sounds good. Should we at least open an issue to track the tf bug and re-use tensordot2 when it is resolved?

mganahl:
good point, yes we should!

PR #951, opened 2021-10-29
Added init file.

AGaliciaMartinez:
I added the init file to the quantum module. To import the module it suffices to do `tensornetwork.quantum` instead of `tensornetwork.quantum.quantum`. Also, the only imported functions are those included in `__all__`  so that autocompletion works better.

### DIFF
--- a/tensornetwork/quantum/__init__.py
+++ b/tensornetwork/quantum/__init__.py
@@ -1,0 +1,1 @@
+from .quantum import *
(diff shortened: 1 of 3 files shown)


AGaliciaMartinez:
@googlebot I signed it!

AGaliciaMartinez:
It seems the CI is failing due to pylint. It complains that `from_tensor` method, when overridden by its subclasses, has a different number of parameters than the parent class. I would prefer not to change those methods in this PR as I believe to be out of its scope. Any suggestions on how to proceed?

